PR #163, opened 2018-11-19
BUG Fix error when calling `unique` on CUDA tensor

janden:
To get the number of classes in `plot_classif.py`, we need to
transfer the Tensor to CPU memory before calling `unique`, since
it is not implemented for CUDA Tensors.


janden:
Ok. Do we want to merge this? Like I said, I don't reproduce the bug on my end, but it seems to work for @eugenium.

eickenberg:
Yes, let's merge this - it can only make it work for more pytorch versions, not fewer

PR #164, opened 2018-11-19
MAINT Replace some environment variables and config file names

janden:
Before, we had `SCATTERING_*` variables. Now we have `KYMATIO_*` variables. This PR also changes the names of some files (config files and caching/datasets directories) to say `kymatio` instead of `scattering`.

### DIFF
--- a/kymatio/caching.py
+++ b/kymatio/caching.py
@@ -9,9 +9,9 @@ def find_cache_base_dir(cache_base_dir=None):
     ---------
     cache_base_dir: string, optional
         Defaults to None. If None, then the cache directory is searched in the
-        environement variable 'SCATTERING_CACHE'. If the latter does not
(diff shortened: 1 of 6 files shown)


lostanlen:
This looks good, thanks for doing this.
As time goes by, I realize how thankful we should be to @gexarcha for coining the name "Kymatio". Grepping for "scattering"  produces a whole lot of false positives, contrary to grepping for "kymatio"

janden:
No problem.

Well the issue right now is that we are indeed grepping for "scattering" right now, which is why all this renaming stuff is messy.

I don't want to merge this until we get an ok from the backend people. @eickenberg @edouardoyallon?

eickenberg:
This looks great to me, thanks for updating that!

PR #152, opened 2018-11-18
WIP working on docs

eickenberg:
push to this branch to collaboratively update the docs

PR #166, opened 2018-11-19
Update timing userguide.rst

eugenium:
Edouard can you confimr the next 2 last entry was supposed to be a V100?


janden:
@eugenium could you apply this to #152 instead? Already made some changes to `userguide.rst` there.

eugenium:
ok!

PR #172, opened 2018-11-20
[MAINT] write contributing guidelines

lostanlen:
adapted from scikit-learn and librosa
the two main dependencies are NumPy, SciPy, and PyTorch

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,123 @@
+
+Contributing code
+=================
+
+How to contribute
+-----------------
+
+The preferred way to contribute to Kymatio is to fork the 
+[main repository](http://github.com/kymatio/kymatio/) on
+GitHub:
+
+1. Fork the [project repository](http://github.com/kymatio/kymatio):
+   click on the 'Fork' button near the top of the page. This creates
+   a copy of the code under your account on the GitHub server.
+
+2. Clone this copy to your local disk:
+
+          $ git clone git@github.com:YourLogin/kymatio.git
+          $ cd kymatio
+
+3. Remove any previously installed versions of Kymatio:
+            
+          $ pip uninstall kymatio
+
+and install your local copy with testing dependencies:
+
+          $ pip install -e .[tests]
+
+4. Create a branch to hold your changes:
+
+          $ git checkout -b my-feature
+
+   and start making changes. Never work in the ``master`` branch!
+
+5. Work on this copy on your computer using Git to do the version
+   control. When you're done editing, do:
+
+          $ git add modified_files
+          $ git commit
+
+   to record your changes in Git, then push them to GitHub with:
+
+          $ git push -u origin my-feature
+
+Finally, go to the web page of the your fork of the librosa repo,
+and click 'Pull request' to send your changes to the maintainers for
+review. This will send an email to the committers.
+
+(If any of the above seems like magic to you, then look up the 
+[Git documentation](http://git-scm.com/documentation) on the web.)
+
+It is recommended to check that your contribution complies with the
+following rules before submitting a pull request:
+
+-  All public methods should have informative docstrings with sample
+   usage presented.
+
+You can also check for common programming errors with the following
+tools:
+
+-  Code with good test coverage (at least 80%), check with:
+
+          $ pytest
+
+-  No pyflakes warnings, check with:
+
+           $ pip install pyflakes
+           $ pyflakes path/to/module.py
+
+-  No PEP8 warnings, check with:
+
+           $ pip install pep8
+           $ pep8 path/to/module.py
+
+-  AutoPEP8 can help you fix some of the easy redundant errors:
+
+           $ pip install autopep8
+           $ autopep8 path/to/pep8.py
+
+Filing bugs
+-----------
+We use Github issues to track all bugs and feature requests; feel free to
+open an issue if you have found a bug or wish to see a feature implemented.
+
+It is recommended to check that your issue complies with the
+following rules before submitting:
+
+-  Verify that your issue is not being currently addressed by other
+   [issues](https://github.com/kymatio/kymatio/issues?q=)
+   or [pull requests](https://github.com/kymatio/kymatio/pulls?q=).
+
+-  Please ensure all code snippets and error messages are formatted in
+   appropriate code blocks.
+   See [Creating and highlighting code blocks](https://help.github.com/articles/creating-and-highlighting-code-blocks).
+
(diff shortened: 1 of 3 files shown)


lostanlen:
@janden fixed in  21fab1c. I removed references to pyflakes and PEP8
I also added specific guidelines for troubleshooting GPU-related bugs in 326eb7a

janden:
Cool. Will squash and clean up.

eickenberg:
> so it's not reasonable to expect new additions to pass.

Actually ....... :) 

I have had a `set cc=80` in my `vim` all throughout this development and have tried to make things conform to PEP8 (noting that I have let a bunch of things slide in big PRs because I valued merging of functionality over my own nitpicks).

I think it's something to strive for, so I wouldn't mind putting it :)

CONTRIBUTING.md
```diff
@@ -0,0 +1,108 @@
...
+          $ git clone git@github.com:YourLogin/kymatio.git
+          $ cd kymatio
+
+3. Remove any previously installed versions of Kymatio:
+
+          $ pip uninstall kymatio
+
+and install your local copy with testing dependencies:
+
+          $ pip install -e .[tests]
```
eickenberg: shouldn't we recommend `python setup.py develop` here?

Or does this do this?

CONTRIBUTING.md
```diff
@@ -0,0 +1,108 @@
...
+
+5. Work on this copy on your computer using Git to do the version
+   control. When you're done editing, do:
+
+          $ git add modified_files
+          $ git commit
+
+   to record your changes in Git, then push them to GitHub with:
+
+          $ git push -u origin my-feature
```
eickenberg: lazy question: what does `-u` do?

CONTRIBUTING.md
```diff
@@ -0,0 +1,108 @@
...
+
+Finally, go to the web page of the your fork of the librosa repo,
```
eickenberg: `librosa` -> `kymatio`

CONTRIBUTING.md
```diff
@@ -0,0 +1,108 @@
...
+-  Please ensure all code snippets and error messages are formatted in
+   appropriate code blocks.
+   See [Creating and highlighting code blocks](https://help.github.com/articles/creating-and-highlighting-code-blocks).
+
+-  Please include your operating system type and version number, as well
+   as your Python, NumPy, SciPy, PyTorch, and Kymatio versions. This information
+   can be found by running the following code snippet:
+
+  ```python
+  import platform; print(platform.platform())
```
eickenberg: whoa ... didn't know that existed

CONTRIBUTING.md
```diff
@@ -0,0 +1,108 @@
...
+  import sys; print("Python", sys.version)
+  import scipy; print("NumPy", scipy.__version__)
+  import scipy; print("SciPy", scipy.__version__)
+  import torch; print("PyTorch", torch.__version__)
+  import kymatio; print("Kymatio", kymatio.__version__)
```
eickenberg: this will work like this after the version.py PR?

CONTRIBUTING.md
```diff
@@ -0,0 +1,108 @@
...
+  ```
+
+- If your issue is related to GPU acceleration, please copy-paste the output
+  from the [environment collection script](https://raw.githubusercontent.com/pytorch/pytorch/master/torch/utils/collect_env.py)
```
eickenberg: Huh cool! What's that? :)

CONTRIBUTING.md
lostanlen: I'm not sure so I'll parrot what I found
It's the same as `--set-upstream`
It's used to set origin as the upstream remote, and sets up tracking.
http://mislav.net/2010/07/git-tips/

CONTRIBUTING.md
lostanlen: only if we put
`from .version import version as __version__`
in `__init__.py`

I like the idea of having a `version.py` separate from `__init__.py`
That way the log of the file `version.py` matches exactly the release cycle of the library, and nothing more

CONTRIBUTING.md
lostanlen: it's a really complex but very cool script written by PyTorch people, which gives very detailed information on the environment, including versions of GCC compiler, GPU models and configuration, NVIDIA drivers, etc. it's really cool
The only drawback is that it takes a while to run (10-20 seconds at least)
in  the long term we could fork this script and make one for our own purposes of troubleshooting

CONTRIBUTING.md
lostanlen: so I see two ways of encoding the version number

(a) the librosa way: there is a very short `version.py` module which just contains the definition of a constant string `version = '0.1.0-dev'` for instance. And this string is made accessible to the module by adding `from .version import version as __version__` in `__init__.py` so that `librosa.__version__` indeed contains `librosa.version.version`

(b) the sklearn way: defining directly `__version__ = '0.1.0-dev'` in `__init__.py`. Again `sklearn.__version__` contains the version number (as a string). But in this case there is no such thing as `sklearn.version.version`.

lostanlen:
I've ran pyflakes and shown the output on the Slack channel. It's not that bad. There are 2-3 small bugs in parts of the code that would be extremely hard to reach anyway. But there are a lot of unused imports. So getting a  💯 from pyflakes is definitely within reach

CONTRIBUTING.md
lostanlen: > Note that it is highly recommended to use pip install . (install) and pip install -e . (developer install) to install packages, as invoking setup.py directly will do the wrong things for many dependencies like pulling prereleases and incompatible packages versions and make the package hard to uninstall with pip.

https://stackoverflow.com/a/19048754

CONTRIBUTING.md
lostanlen: aouch

lostanlen: fixed in 471bc0c

janden:
@lostanlen did you not reset your HEAD after the force-push? Looks like it re-introduced your old (squashed) commits.

CONTRIBUTING.md
lostanlen: OK we'll go for (a) then. This was also the impression I had.
It also makes `setup.py` and `docs/conf.py` simpler

lostanlen: done in 9caa3f9

lostanlen:
I am going to start over with this.

PR #173, opened 2018-11-20
ENH Include CUDA check in 1D `compute_speed.py`

janden:
If we don't have a GPU, don't try to use a CUDA implementation.


janden:
Unless there are objections, will merge this in a little bit.

eickenberg:
LGTM

PR #181, opened 2018-11-21
MAINT Fix `scattering` in code

janden:
Need to have `scattering`, not `scat`.

### DIFF
--- a/kymatio/scattering1d/scattering1d.py
+++ b/kymatio/scattering1d/scattering1d.py
@@ -415,16 +415,16 @@ class Scattering1D(object):
                 raise ValueError(
                     'Options average=False and vectorize=True are ' +
                     'mutually incompatible. Please set vectorize to False.')
-            size_scat = self.precompute_size_scattering(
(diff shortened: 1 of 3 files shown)


kymatio/scattering1d/scattering1d.py
```diff
@@ -619,7 +619,7 @@ def scattering(x, psi1, psi2, phi, J, pad_left=0, pad_right=0,
         Whether to compute the 2nd order or not. Defaults to False.
     average_U1 : boolean, optional
         whether to average the first order vector. Defaults to True
-    size_scat : dictionary or int, optional
+    size_scattering : dictionary or int, optional
```
lostanlen: it would be good to take this opportunity to document how to use it as a `dict` input

janden: Not in this PR, I think. At any rate, this function is an internal one (not to be used publicly), so its documentation is not top priority. We may even integrate it into the `Scattering1D` class.

PR #183, opened 2018-11-22
MAINT define __version__

### LINKED ISSUE #182: [MAINT] AttributeError: module 'kymatio' has no attribute '__version__'
This came up in #172

PEP 396 has introduced the convention of exposing the constant `__version__` to the top-level module
At the moment this is in `kymatio.version.version` (so that `version.py` can be loaded in `setup.py` without importing the whole `kymatio` module)
https://www.python.org/dev/peps/pep-0396/

Will submit a PR later

lostanlen:
Closes #182
from .version import version as __version__ in __init__.py allows kymatio.__version__ to be equal to kymatio.version.version
This is compliant with PEP 396
I'm taking the opportunity of this PR to reorder 1D, 2D, and 3D imports

### DIFF
--- a/kymatio/__init__.py
+++ b/kymatio/__init__.py
@@ -1,13 +1,3 @@
-__all__ = [
-            'Scattering3D',
-            'Scattering2D',
-            'Scattering1D'
-            ]
-
-from .scattering2d.scattering2d import Scattering2D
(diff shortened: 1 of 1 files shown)


eickenberg:
Thanks!

PR #184, opened 2018-11-23
FIX Removing a warning

### LINKED ISSUE #180: Warning in `mnist.py`
Specifically, https://github.com/kymatio/kymatio/blob/289bc26551e92456ef7a48fbe83d48e157f7632c/examples/2d/mnist.py#L50 generates a warning saying that `size_average` will be deprecated and says to use `reduction='sum'` instead. Is this ok for us to do?

edouardoyallon:
Close #180 

Depreciation due to pytorch.


lostanlen:
I just read `legacy_get_string`
https://pytorch.org/docs/stable/_modules/torch/nn/functional.html
and your replacement looks good to me.
Can we merge this?

PR #188, opened 2018-11-23
FIX __all__ becomes correct both for 1D and 2D

### LINKED ISSUE #176: [BUG] __all__ in scattering2d.py
another bug I found while eyeballing at pyflakes

https://github.com/kymatio/kymatio/blob/master/kymatio/scattering2d/scattering2d.py#L5

`Scattering` should read as `Scattering2D`

this affects the behavior of `from kymatio.scattering2d import *`

EDIT: the reason why it does work as expected is that it gets overwritten by `__init__.py` in the module. Still, it would be good to clarify the situation

tagging @eugenium

edouardoyallon:
Close #176

### DIFF
--- a/kymatio/scattering1d/scattering1d.py
+++ b/kymatio/scattering1d/scattering1d.py
@@ -7,6 +7,8 @@ import math
 
 import warnings
 
+__all__ = ['Scattering1D']
+
 from .backend import pad, unpad, real, subsample_fourier, modulus_complex, fft1d_c2c, ifft1d_c2c
 from .utils import compute_padding, compute_border_indices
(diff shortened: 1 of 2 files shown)


lostanlen:
Thanks. Do

`from kymatio.scattering1d.scattering1d import *`
`from kymatio.scattering2d.scattering2d import *`
`from kymatio.scattering3d.scattering3d import *`

all work properly now?

lostanlen:
So what is the purpose of the list
`kymatio.scattering2d.scattering2d.__all__`
if the file
`kymatio.scattering2d.__init__.py`
contains
`from .scattering2d import Scattering2D`?

PR #189, opened 2018-11-24
FIX removing ignore-pattern files from sphinx' conf.py

### LINKED ISSUE #150: [DOC] solving Sphinx issues
W.r.t.:
Issues: #97  #110 #135 #105 #106 #108 #111 
PR: #144

We must find a way to remove the list of "ignore-patterns" here:

https://github.com/eickenberg/kymatio/blob/master/doc/source/conf.py#L61 

At the moment we do this, and that the sphinx works, then we will be good with the gallery!

edouardoyallon:
Close #150

### DIFF
--- a/examples/3d/scattering3d_qm7.py
+++ b/examples/3d/scattering3d_qm7.py
@@ -1,5 +1,5 @@
 """
-3D scattering quantum chemestry regression
+3D scattering quantum chemistry regression
 ==========================================
 This uses the 3D scattering on a standard dataset.
 """
--- a/examples/3d/scattering3d_qm7.py
+++ b/examples/3d/scattering3d_qm7.py
@@ -1,7 +1,7 @@
 """
 3D scattering quantum chemestry regression
 ==========================================
-This uses the 3D scattering on a standard dataset.s
+This uses the 3D scattering on a standard dataset.
 """
 
 import numpy as np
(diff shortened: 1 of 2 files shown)


eickenberg:
Sure about including the 3D example?

It takes like 3 or 8h to run.

I asked for there to be caching for sphinx-gallery purposes, but this is not done yet.

I guess if we keep it as `qm7_...` and not `plot_...` then it won't be run, so maybe it's OK to include. But i'd like to be sure it doesn't run

edouardoyallon:
> It takes like 3 or 8h to run.
> 
> I asked for there to be caching for sphinx-gallery purposes, but this is not done yet.

leads to:

`copying downloadable files... [100%] gallery_3d/scattering3d_qm7.ipynb `

The current build of the doc on this branch takes about 1-2 minutes.

Thus I believe it is a good idea to merge.

examples/3d/scattering3d_qm7.py
```diff
@@ -1,3 +1,9 @@
+"""
+3D scattering quantum chemestry regression
```
lostanlen: chemestry -> chemistry

edouardoyallon: This is not the purpose of this PR.

lostanlen: I see it in the diff. So if it's an addition to the codebase, it should be made correct in this PR. If it's already on the master branch, you are right that we should fix the spelling in a separate PR, but I don't think it is the case.

PR #192, opened 2018-11-24
FIX caching was not usable + missing library to import

### LINKED ISSUE #175: [BUG] cache management of 2-D filter banks
A bug found by running `pyflakes` on the Kymatio source.
Originating in code by @edouardoyallon  and @gabrielhuang

It happens in a `try` statement which is itself part of an `except` statement, so the bug is really hard to trigger, but it's there.

https://github.com/kymatio/kymatio/blob/56a846f059ba5e185a7378e4a057f11bea25552a/kymatio/scattering2d/filter_bank.py#L119
`filer_file` should read as `filter_file`

I think that, long term, this caching mechanic would be more elegantly (and more stably) implemented with `joblib.Memory`, and that the calls to `print` could be replaced by `logging`.

One example of joblib-based cache manager which comes very close to what we need here is available in `librosa`, cf. https://github.com/librosa/librosa/blob/master/librosa/cache.py

edouardoyallon:
Close #175

### DIFF
--- a/kymatio/scattering2d/filter_bank.py
+++ b/kymatio/scattering2d/filter_bank.py
@@ -7,6 +7,7 @@ __all__ = ['filter_bank']
 
 import torch
 import numpy as np
+import os
 from .utils import fft2
 from ..caching import get_cache_dir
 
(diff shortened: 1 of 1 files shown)


eickenberg:
This PR needs a test

lostanlen:
Honestly I'm not very enthusiastic at the idea of having a single kwarg produce big side effects like calling `torch.save`. This may cause race conditions when deployed on distributed computing architectures, has stark differences in speed for reasons that are difficult to debug (because they are not within scope), etc.
As written in #175 I would prefer, if anything, to have this rely on a `CacheManager` from `joblib`. 
But for the time being we can also get rid of it altogether.

PR #195, opened 2018-11-25
[API] set J as first positional arg + introduce shape kwarg

### LINKED ISSUE #194: Inconsistency in the position of J across Scattering*d constructors
A follow-up on a discussion with @janden  and @eickenberg  in #158 
At the moment, the order of positional arguments in the scattering constructors are:
Scattering1d: `T, J, Q`
Scattering2d: `M, N, J`
Scattering 3d: `M, N, O, J, L, sigma_0`

The only parameter that is truly common to each of these classes is `J`. It is also the only one in which we really can't provide a good default value. Lastly, we could make the case that the parameters which are relative to input shape (`T`, `M`, `N`, `O`) could be inferred directly from `J` by default. This would pave the way for allowing variable-sized inputs at runtime.

That's clearly a lot of work, and I don't want to schedule all that at once. We can work towards this progressively. However, one necessary change to the API is the order of positional parameters. I would recommend

Scattering1d: `J, T, Q`
Scattering2d: `J, M, N`
Scattering 3d: `J, M, N, O, L, sigma_0`

That way, in the future we will be able to progressively include simpler constructors in a backwards compatible way, and/or with appropriate deprecation warnings.
This is a big API decision but it is not a lot of work on the part of the code. There are only a few places in the docs and examples in which these constructors are called with positional parameters.

### LINKED ISSUE #202: need for a cross-module `shape` kwarg to constructors
An issue that appeared in discussion of #194 
It is already on its way to be solved because of PR #195 (which incorporated github.com/lostanlen/kymatio/pull/2 by @eickenberg)
But I'm still opening it so that we can add it to the alpha milestone and be reminded of it in our API changelog

lostanlen:
Fixes #194
This change is API-breaking and irreversible insofar as it cannot be deprecated.
It is a step in the direction of making the input shape optional.
It also improves consistency across modules.
[EDIT: fixes #202 too]

### DIFF
--- a/doc/source/userguide.rst
+++ b/doc/source/userguide.rst
@@ -119,7 +119,7 @@ When initialized, a scattering transform object is stored on the CPU::
     import torch
     from kymatio import Scattering2D
 
-    scattering = Scattering2D(32, 32, 2)
+    scattering = Scattering2D(2, 32, 32)
 
 We use this to compute scattering transforms of signals in CPU memory::
 
--- a/kymatio/scattering1d/scattering1d.py
+++ b/kymatio/scattering1d/scattering1d.py
@@ -120,15 +120,15 @@ class Scattering1D(object):
     ::
 
         # Set the parameters of the scattering transform.
-        T = 2**13
         J = 6
+        T = 2**13
         Q = 8
 
         # Generate a sample signal.
         x = torch.randn(1, 1, T)
 
         # Define a Scattering1D object.
-        S = Scattering1D(T, J, Q)
+        S = Scattering1D(J, T, Q)
 
         # Calculate the scattering transform.
         Sx = S.forward(x)
@@ -142,11 +142,11 @@ class Scattering1D(object):
 
     Parameters
     ----------
-    T : int
-        The length of the input signals.
     J : int
         The maximum log-scale of the scattering transform. In other words,
         the maximum scale is given by `2**J`.
+    T : int
+        The length of the input signals.
     Q : int >= 1
         The number of first-order wavelets per octave (second-order wavelets
         are fixed to one wavelet per octave).
@@ -202,11 +202,11 @@ class Scattering1D(object):
 
     Attributes
     ----------
-    T : int
-        The length of the input signals.
     J : int
         The maximum log-scale of the scattering transform. In other words,
         the maximum scale is given by `2**J`.
+    T : int
+        The length of the input signals.
     Q : int
         The number of first-order wavelets per octave (second-order wavelets
         are fixed to one wavelet per octave).
@@ -245,14 +245,14 @@ class Scattering1D(object):
         or collected into a dictionary. For more details, see the
         documentation for `forward()`.
     """
-    def __init__(self, T, J, Q, normalize='l1', criterion_amplitude=1e-3,
+    def __init__(self, J, T, Q, normalize='l1', criterion_amplitude=1e-3,
                  r_psi=math.sqrt(0.5), sigma0=0.1, alpha=5.,
                  P_max=5, eps=1e-7, max_order=2, average=True,
                  oversampling=0, vectorize=True):
         super(Scattering1D, self).__init__()
         # Store the parameters
-        self.T = T
         self.J = J
+        self.T = T
         self.Q = Q
         self.r_psi = r_psi
         self.sigma0 = sigma0
--- a/kymatio/scattering1d/scattering1d.py
+++ b/kymatio/scattering1d/scattering1d.py
@@ -4,6 +4,7 @@
 import torch
 import numpy as np
 import math
+import numbers
 
 
 __all__ = ['Scattering1D']
@@ -205,7 +206,7 @@ class Scattering1D(object):
     J : int
         The maximum log-scale of the scattering transform. In other words,
         the maximum scale is given by `2**J`.
-    T : int
+    shape : int
         The length of the input signals.
     Q : int
         The number of first-order wavelets per octave (second-order wavelets
@@ -245,14 +246,14 @@ class Scattering1D(object):
         or collected into a dictionary. For more details, see the
         documentation for `forward()`.
     """
-    def __init__(self, J, T, Q, normalize='l1', criterion_amplitude=1e-3,
+    def __init__(self, J, shape, Q, normalize='l1', criterion_amplitude=1e-3,
                  r_psi=math.sqrt(0.5), sigma0=0.1, alpha=5.,
                  P_max=5, eps=1e-7, max_order=2, average=True,
                  oversampling=0, vectorize=True):
         super(Scattering1D, self).__init__()
         # Store the parameters
         self.J = J
-        self.T = T
+        self.shape = shape
         self.Q = Q
         self.r_psi = r_psi
         self.sigma0 = sigma0
@@ -277,6 +278,18 @@ class Scattering1D(object):
         automatically during object creation and no subsequent calls are
         therefore needed.
         """
+
+        # check the shape
+        if isinstance(self.shape, numbers.Integral):
+            self.T = self.shape
+        elif isinstance(self.shape, tuple):
+            self.T = self.shape[0]
+            if len(self.shape) > 1:
+                raise ValueError("If shape is specified as a tuple, it must "
+                                 "have exactly one element")
+        else:
+            raise ValueError("shape must be an integer or a 1-tuple")
+
         # Compute the minimum support to pad (ideally)
         min_to_pad = compute_minimum_support_to_pad(
             self.T, self.J, self.Q, r_psi=self.r_psi, sigma0=self.sigma0,
--- a/doc/source/quickstart.rst
+++ b/doc/source/quickstart.rst
@@ -22,7 +22,7 @@ consisting of Gaussian white noise::
 
     import torch
     from kymatio import Scattering2D
-    scattering = Scattering2D(32, 32, 2)
+    scattering = Scattering2D(2, 32, 32)
     x = torch.randn(1, 1, 32, 32)
     Sx = scattering(x)
     print(Sx.size())
--- a/examples/1d/compute_speed.py
+++ b/examples/1d/compute_speed.py
@@ -76,7 +76,7 @@ if backend.NAME == 'skcuda' and torch.cuda.is_available():
 # Create the `Scattering1D` object using the given parameters and generate
 # some compatible test data with the specified batch size.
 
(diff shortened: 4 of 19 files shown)


kymatio/scattering1d/scattering1d.py
```diff
@@ -16,7 +16,7 @@
 from .filter_bank import calibrate_scattering_filters
 
 
-def compute_minimum_support_to_pad(T, J, Q, criterion_amplitude=1e-3,
+def compute_minimum_support_to_pad(J, T, Q, criterion_amplitude=1e-3,
```
eickenberg: not sure if this makes sense semantically. padding is about `T` first and foremost

lostanlen: you're right. I was on the verge about this one. will fix

lostanlen: fixed in 805a9f1

lostanlen:
@eickenberg yes I just successfully built the docs. They appear as expected.

eickenberg:
It would also be great to have a `shape` argument (which then gets checked by `M, N = shape` or `M, N, O = shape`). That would bring us another step closer to uniform arguments.

What does 1D typically call the size of its signal? Length? size? shape?

lostanlen:
@eickenberg absolutely. But this can be a separate PR.

Also the name of this keyword argument is crucial, as you point out very well.
In my Julia implementation (github.com/lostanlen/WaveletScattering.jl) I use `log2_size`. Type is `integer` or vector of integer. In `scattering.m` I use `size` but the problem is that my code isn't meant to work on non-power-of-two values, so it throws a `RuntimeError` rather than a `TypeError`. 
In Kymatio the size of the signal is called `T` which is a terrible name because

1. it doesn't explicitly state that only powers of two are allows
2. In Joakim's papers, T is 2**J, whereas the length of the signal is `N`. So it's prone to confusion.

let's move this discussion to a separate issue

eickenberg:
I'm fine with this, but I do think that an API change to have `shape` (or, unfortunately, `size`) should go along with this change. 

If we set J first without adding `shape/size` then this change will not make prior `pyscatwave` code absolutely fail, but make for weird shapes and annoyance with those who have been using `pyscatwave` (i.e. people interested in scattering). 

I'd like the instantiation of the object to throw an error and that is achievable by having the shape variable or by having the shape inputs be keyword only. But the former change seems better to me at the moment

eickenberg:
@janden @edouardoyallon

lostanlen:
that makes sense. while we're still not in alpha, we can still claim a situation of `hic sunt dracones` regarding backwards compatibility; so although I can see why both changes should happen at the same release, they don't necessarily have to belong to the same exact PR.
But I'm completely fine with working on that urgently.

I guess that the main question we should ask ourselves is: do we want to guarantee that non-powers-of-two filterbank sizes should be supported? At the moment the 1d does not support non-powers-of-two but it does not catch it appropriately. I *think* that 2-d supports non-powers-of-two because it talks about ImageNet in the README, but it is not tested and there are no examples beyong `M=32`. I would vote for making expressing this shape in `log2` domain.

eickenberg:
> I would vote for making expressing this shape in log2 domain.

As much as that is beautiful, I think that will kill all practical applications for deep learning on 2D images. That would mean that they would either have to be 128 or 256 which are both not used. Instead, for imagenet there are lots of examples that use  betwee 224 and 231 or 112. This is often after careful consideration of fitting network and a minimal batch size into a GPU. Having powers of 2 etched in here (especially without having some sort of overlap-add procedure or equivalent that is as fast as the current 2D backend that can bring a smaller-size filter to analyze a bigger image) will make a lot of that work impossible.

> I think that 2-d supports non-powers-of-two because it talks about ImageNet in the README, but it is not tested and there are no examples beyong M=32

Indeed it should be tested. There is evidence of it working very well in @edouardoyallon 's and @eugenium 's papers.

lostanlen:
> Well I assume you need an even number to be able to fourier-subsample. 
True. `size=96` passes without problem. So I guess that the same is true of `224` because it's equal to `32*7`.
But that API seems like really hard to maintain.
I don't want to spend my life explaining to beginners why 
`Scattering2D(M=224, N=224,J=5, L=4)` passes at runtime and why
`Scattering2D(M=224, N=224,J=6, L=4)` fails at runtime.
Or else we need to check at construction time that `2**J` divides both `M` and `N`.

lostanlen:
I should add that `scattering2d` defines `M` and `N` as input size *before padding*.
So:
`S = Scattering2D(M=224, N=224, J=4, L=4)`
leads to
`S.M_padded` being `256`, which is a power of two.
Therefore, constraining the `shape` to be a power of two, i.e. `log2_size` to be integer, is compatible with Edouard's ImageNet experiments without modification on the actual processing pipeline.

lostanlen:
One thing that I *can* put in this PR without too much effort is to enforce that `M` and `N` be required keyword arguments rather than positional arguments. This will break the three-positional-argument form inherited from `pyscatwave`, so there is no risk of people getting false but non-failing results while updating from `pyscatwave`. 
It's only a matter of adding a `*` in the prototype, writing nice error messages for missing `M` and `N` values, and updating the docs. The code itself needs not be affected.

eickenberg:
I made a PR to your PR here: https://github.com/lostanlen/kymatio/pull/2

so meta

eugenium:
> I should add that `scattering2d` defines `M` and `N` as input size _before padding_.
> So:
> `S = Scattering2D(M=224, N=224, J=4, L=4)`
> leads to
> `S.M_padded` being `256`, which is a power of two.
> Therefore, constraining the `shape` to be a power of two, i.e. `log2_size` to be integer, is compatible with Edouard's ImageNet experiments without modification on the actual processing pipeline.

If you use J=3, which is sometimes (more commonly) used in our imagenet experiments it would be 240

lostanlen:
@eugenium what's your opinion on that, long-term?

eugenium:
We could warn the user about the padding behavior at some point. I don't see an easy solution really, letting the user control the input size trumps all these concerns.

lostanlen:
Right, me neither. I think that what we have is more than good enough for an alpha version.
The user supplies a `J` and a `shape`, and can retrieve the scattering coefficients from any signal of size `shape`.  It sounds like a reasonable spec.

janden:
> Also the name of this keyword argument is crucial, as you point out very well.
> In my Julia implementation (https://github.com/lostanlen/WaveletScattering.jl) I use `log2_size`. Type is `integer` or vector of integer. In `scattering.m` I use `size` but the problem is that my code isn't meant to work on non-power-of-two values, so it throws a `RuntimeError` rather than a `TypeError`.
> In Kymatio the size of the signal is called `T` which is a terrible name because ...

Agreed that `T` is bad, but I don't think we should restrict ourselves to powers-of-two, for the reasons described by others. If we do need powers-of-two for computational reasons, this can be fixed in the padding (although this has an effect on the calculation, so maybe not a good idea).

While we can enforce powers-of-two for now (throwing some sort of `Error` if the value is not a power of two), we leave ourselves the ability to implement arbitrary shapes later on. We may not even have to force `M*2**J` shapes if we're clever about the padding (need to look into this more carefully).

As a sidenote, the whole padding thing needs to be standardized across dimensions and thought about. For example, right now, the 1D code complains for small signal shapes because it is unhappy about the padding. Ultimately, we may do better by including the details of the padding (and potentially giving the user some control over these) when describing the scattering computation in the docs.

> I don't want to spend my life explaining to beginners why
> `Scattering2D(M=12, N=12, J=2, L=4)` passes at runtime and why
> `Scattering2D(M=18, N=18, J=2, L=4)` fails at runtime.
> Or else we need to check at construction time that `2**J` divides both `M` and `N`. But that will also raise some eyebrows.

Do you think so? What do all the other libraries that allow for subsampling do? They must do some sort of check like this.

> One thing that I _can_ put in this PR without too much effort is to enforce that `M` and `N` be required keyword arguments rather than positional arguments. This will break the three-positional-argument form inherited from `pyscatwave`, so there is no risk of people getting false but non-failing results while updating from `pyscatwave`.

This seems to be quite strong. Brings back memories of those terrifying "blacklists" from ScatNet days.. Are we supposed to be PyScatWave compatible? We certainly don't claim to be. If anything, we should probably have a migration guide somewhere in the docs.

lostanlen:
> This seems to be quite strong. Brings back memories of those terrifying "blacklists" from ScatNet days.. Are we supposed to be PyScatWave compatible? We certainly don't claim to be. If anything, we should probably have a migration guide somewhere in the docs.

I've changed my mind between yesterday and today. @eickenberg thinks that all of this should be supplied by a simple tuple kwarg like `shape` rather than multiple kwargs `M`, `N` ,and it would be much better this way. See his PR on lostanlen/kymatio
https://github.com/lostanlen/kymatio/pull/2

janden:
We'll most likely want to rebase this. First, it contains a merge commit, which we'd like to avoid. Second, I don't think all of @eickenberg's commits pass the tests (since each seems to refactor a different set of files separately), so we don't want them in the repo (confirm?).

lostanlen:
Yes, let's rebase

janden:
Ok. Give me a moment.

janden:
Hold on. Don't touch it. Will need to force-push.

janden:
Ok. Looks like we're good.

Are we ready to merge? I'm asking because there's only one "Approve" so far. Will leave this here.

lostanlen:
Hello @edouardoyallon  @eugenium @eickenberg @AndreuxMath 
our big API overhaul of the `Scattering*D` constructors is ready for merging. Basically, we're going from
* `Scattering1D(T,J,Q)` to `Scattering(J,T=T,Q=Q)`
* `Scattering2D(M,N,J,L)` to `Scattering2D(J,shape=(M,N),L=L)`
* `Scattering3D(M,N,O,J,L)` to `Scattering3D(J,shape=(M,N,O),L=L)`

please review this change and approve it in the "code review" section of this PR on the GitHub browser interface.

doc/source/userguide.rst
```diff
@@ -119,7 +119,7 @@ When initialized, a scattering transform object is stored on the CPU::
     import torch
     from kymatio import Scattering2D
 
-    scattering = Scattering2D(32, 32, 2)
+    scattering = Scattering2D(2, 32, 32)
```
lostanlen: `scattering = Scattering2D(2, shape=(32,32))`

eickenberg:
Approved :)

lostanlen:
@eickenberg click the button :)

PR #200, opened 2018-11-25
MAINT set version to 0.1.0-dev [ci skip]

lostanlen:
Semver 2.0.0 compliant
https://semver.org/spec/v2.0.0.html

PEP 0440 compliant
https://www.python.org/dev/peps/pep-0440/

### DIFF
--- a/kymatio/version.py
+++ b/kymatio/version.py
@@ -1,2 +1,2 @@
-short_version = '0.1.1-alpha'
-version = '0.1.1'
+short_version = '0.1'
+version = '0.1.0-dev'


eickenberg:
So, test.pypi has already locked the versions 0.1.0 and 0.1.1 and we can't use them anymore on test.pypi (so 0.1.1 will not be possible to upload anymore).

Luckily we *can* use them on pypi, because I haven't uploaded to there yet.

lostanlen:
phew!  that is good to hear.
NB: the master branch should never be on PyPI. To make our alpha release, we need to
1. tag master as `0.1.0-branching`
2. create a `0.1.*` branch
3. change `version.version` to `0.1.0-alpha`
3. tag the `0.1.*` as `0.1.0-alpha`
4. release the `0.1.*` branch on PyPI

eickenberg:
Wait, so you actually put a `*` in the name of a branch?

I was actually given a 4-step way of doing this that seems a bit simpler

eickenberg:
I will merge this change now and keep note of where to find these instructions :)

lostanlen:
Yes. scikit-learn uses the letter `X` but I would prefer an asterisk `*`. It looks more like a wildcard and less like a constant literal.
Steps 1 and 7 are optional but would make debugging so much easier
The purpose of step 1 is to be able to quickly trace back to last commit on master before a given release. It's also useful for making diffs on master.

The purpose of step 7 is that, assuming that `0.1.0-beta` is released, if a user complains about a bug in `0.1.0-beta`, we cannot know for sure whether this bug is really in the release or whether it's from a development build of `-beta` that the user has made from source on the `0.1.*` branch. Whereas if we have `version==0.1.0-beta-dev` on the `0.1.*` until beta release, then we ask the user to 
`print(kymatio.__version__)`
and if the answer is not `0.1.0-beta` but `0.1.0-beta-dev`, we don't even look at the bug and answer: "please upgrade to beta".
on the other hand, if the answer is `0.1.0-beta`, then the bug is indeed in the PyPI `0.1.0-beta` release and it should be attended before releasing `0.1.0-rc0`.

janden:
I'm not sure I understand the rationale for not publishing `master` onto PyPI. Having our `master` not track the stable releases of the software seems odd. I would assume that any development would happen in some dev branch (`dev`, `next`, or something) that is then merged into `master` as needed.

As for `0.1.*`, that sounds terrifying. Can we do `0.1.X`?

lostanlen:
> I'm not sure I understand the rationale for not publishing master onto PyPI.

The rationale is that, once `0.1.0` will be out, we will have two kinds of user feedback:
* bug reports
* feature requests
Both kinds of user feedback will lead to pull requests. Those of the former kind will be merged to both `0.1.X` and `master`. Those of the second kind will be merged only to `master`.
If we publish `master` into PyPI, we can only do major-minor releases (`0.1.0`, `0.2.0`, etc.), not bugfix releases (`0.1.1`, `0.1.2`, etc.) because these bugfix release contain only a subset of the commits that are posterior to the minor release.


> As for 0.1.*, that sounds terrifying. Can we do 0.1.X?

Yes we can. Are you worried of passing the asterisk wildcard to git command line?
I liked the idea of an asterisk but you're right that it would be more prudent to use an X.

janden:
> If we publish `master` into PyPI, we can only do major-minor releases (`0.1.0`, `0.2.0`, etc.), not bugfix releases (`0.1.1`, `0.1.2`, etc.) because these bugfix release contain only a subset of the commits that are posterior to the minor release.

I guess what I was thinking was the opposite (dual?) approach. Bugfixes are committed into `master` and `devel` while feature requests are only in `devel`, then when a new major-minor is ready, we merge `devel` into `master` (or rebase, really). Perhaps there are some issues with this.

> Yes we can. Are you worried of passing the asterisk wildcard to git command line?

Not worried, really. I mean, I can escape my asterisks when necessary. I'd just rather it would not be necessary...

> I liked the idea of an asterisk but you're right that it would be more prudent to use an X.

Ok good.

PR #209, opened 2018-11-28
STY pycodestyle and pyflakes scattering1d.py

lostanlen:
also alphabetize imports, put Scattering1D class at the top of the file, fix maths in docstrings, Unicode "Andén"

### DIFF
--- a/kymatio/scattering1d/scattering1d.py
+++ b/kymatio/scattering1d/scattering1d.py
@@ -1,80 +1,19 @@
-# Authors: Mathieu Andreux, Joakim Anden, Edouard Oyallon
-# Scientific Ancestry: Mathieu Andreux, Vincent Lostanlen, Joakim Anden
+# Authors: Mathieu Andreux, Joakim Andén, Edouard Oyallon
+# Scientific Ancestry: Joakim Andén, Mathieu Andreux, Vincent Lostanlen
 
-import torch
-import numpy as np
 import math
 import numbers
+import numpy as np
+import torch
+
+from .backend import (fft1d_c2c, ifft1d_c2c, modulus_complex, pad, real,
+    subsample_fourier, unpad)
+from .filter_bank import (calibrate_scattering_filters,
+    scattering_filter_factory)
+from .utils import cast_phi, cast_psi, compute_border_indices, compute_padding
 
 
 __all__ = ['Scattering1D']
-
-from .backend import pad, unpad, real, subsample_fourier, modulus_complex, fft1d_c2c, ifft1d_c2c
-from .utils import compute_padding, compute_border_indices
-from .utils import cast_psi, cast_phi
-
-from .filter_bank import scattering_filter_factory
-from .filter_bank import calibrate_scattering_filters
-
-
-def compute_minimum_support_to_pad(T, J, Q, criterion_amplitude=1e-3,
-                                   normalize='l1', r_psi=math.sqrt(0.5),
-                                   sigma0=1e-1, alpha=5., P_max=5, eps=1e-7):
-    """
-    Computes the support to pad given the input size and the parameters of the
-    scattering transform.
-
-    Parameters
-    ----------
-    T : int
-        temporal size of the input signal
-    J : int
-        scale of the scattering
-    Q : int
-        number of wavelets per octave
-    normalize : string, optional
-        normalization type for the wavelets.
-        Only 'l2' or 'l1' normalizations are supported.
-        Defaults to 'l1'
-    criterion_amplitude: float >0 and <1, optional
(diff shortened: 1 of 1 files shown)


lostanlen:
This is a big PR but the changes have no effect on the code. I wanted to get this done before going further with the 1-D API. Can I merge?

kymatio/scattering1d/scattering1d.py
```diff
@@ -1,80 +1,19 @@
...
+# Scientific Ancestry: Joakim Andén, Mathieu Andreux, Vincent Lostanlen
 
-import torch
-import numpy as np
 import math
 import numbers
+import numpy as np
+import torch
 
+from .backend import fft1d_c2c, ifft1d_c2c, modulus_complex, pad, real,\
```
eickenberg: Is this to respect 80 columns? Can we have parentheses instead of backslashes?

kymatio/scattering1d/scattering1d.py
```diff
@@ -90,9 +29,12 @@ class Scattering1D(object):
 
         :math:`S_J^{(0)} x(t) = x \\star \\phi_J(t)`,
 
-        :math:`S_J^{(1)} x(t, \\lambda) = |x \\star \\psi_\\lambda^{(1)}| \\star \\phi_J`, and
+        :math:`S_J^{(1)} x(t, \\lambda) =
```
eickenberg: We should really really be using $-math in these docstrings.

I had pointed to the `texext` sphinx extension but never added it. Hope to get round to this.

Don't fix it here.

kymatio/scattering1d/scattering1d.py
```diff
@@ -359,8 +301,8 @@ def meta(self):
         meta : dictionary
             See the documentation for `compute_meta_scattering()`.
         """
-        return Scattering1D.compute_meta_scattering(self.J, self.Q,
```
eickenberg: remind me: what is the rule behind this modification?

kymatio/scattering1d/scattering1d.py
```diff
@@ -692,7 +701,7 @@ def scattering(x, psi1, psi2, phi, J, pad_left=0, pad_right=0,
             S[:, cc, :] = S1_J.squeeze(dim=1)
             cc += 1
         else:
-            S[n1,] = S1_J
+            S[n1, ] = S1_J
```
eickenberg: uuuuh ... ? That looks like R code

`S[n1]`

kymatio/scattering1d/scattering1d.py
lostanlen: the rule is that, on a multiline statement within parentheses, all lines following the opening parenthesis should be at the same indentation level. So either:
1. put the `m` of  `max_order` at the same level as the `s` of `self.J`
or
2. return to next line right after the opening parenthesis so that `self.J, self.Q, max_order=self.max_order` are a single line. in this case the indentation level can be anything

I opted for 2. because 1. would lead to 80+ characters on the second line

kymatio/scattering1d/scattering1d.py
lostanlen: oh, ok. i was confused seeing this

kymatio/scattering1d/scattering1d.py
janden: Ok. I put this `` :math:`...` `` stuff all over the 1D docstrings (and probably other places too). Will open an issue to address this.

kymatio/scattering1d/scattering1d.py
janden: It's because when `vectorize=True`, the output is a dictionary indexed by tuples of filter indices. Consequently, the zeroth order is in `S{()]`, the first order with filter `n1` is in `S[(n1,)]`, and the second order with filters `n1` and `n2` is in `S[(n1,n2)]`. Not sure if we're going to keep this capability, but that's what's going on.

I'm assuming correct style would call for `S[(n1,)]`?

kymatio/scattering1d/scattering1d.py
eickenberg: :pray:

kymatio/scattering1d/scattering1d.py
eickenberg: OOOH, my bad! Sorry about  that.

Yeah, so as much as I like to point out that tuples are defined because of commas, not parentheses (i.e. `n1,` is a tuple just fine), I think it's better to have the parentheses here for readability. So yeah `S[(n1,)]` as you say

lostanlen: addressed in 89e81fe

PR #208, opened 2018-11-27
implement skcuda backend for 3D

louity:
13% speedup:

==> Testing Float32 with skcuda backend, on GPU, forward
Elapsed time: 26.752710 [s / 10 evals], avg: 2.68 (s/batch)

==> Testing Float32 with torch backend, on GPU, forward
Elapsed time: 29.869728 [s / 10 evals], avg: 2.99 (s/batch)

PR #210, opened 2018-11-28
REF Scattering3d now returns tensor, not tuple

### LINKED ISSUE #158: scattering3d returns a tuple instead of tensor
Each implementation has very different expectations of its inputs and outputs:
- 1D: Expects inputs of size `(B, 1, T)`, outputs `(B, P, T/2**J)`, where `P` is the number of scattering coefficients. Note that if the second dimension of the input is not `1`, it errors.
- 2D: Expects inputs of size `(B, C, M, N)`, outputs `(B, C, P, M/2**J, N/2**J)`. Note that the number of dimensions changes here between input and output (this may be desirable but is not consistent with 1D). Also, from what I understand, the second dimension here is essentially treated as a batch dimension.
- 3D: Expects inputs of size `(B, M, N, O)`, outputs a _tuple_ with elements of size `(B, ?, ?, ?)`. Not sure what's going on here.

We should think about harmonizing these to some extent.

lostanlen:
fixes #158


lostanlen:
Thanks @janden. Let's rebase and merge when you're ready. Hopefully this should not interfere with #208

janden:
> Thanks @janden. Let's rebase and merge when you're ready.

Ok. Do we have sign-off on this one from the others? I seem to remember @eickenberg being on board, at least.

> Hopefully this should not interfere with #208

It will, but we'll get through it. :grin:

examples/3d/scattering3d_qm7.py
```diff
@@ -208,61 +206,56 @@ def compute_qm7_solid_harmonic_scattering_coefficients(
...
 
     order_0 = torch.cat(order_0, dim=0)
-    order_1 = torch.cat(order_1, dim=0)
-    order_2 = torch.cat(order_2, dim=0)
+    orders_1_and_2 = torch.cat(orders_1_and_2, dim=0)
 
-    return order_0, order_1, order_2
+    return order_0, orders_1_and_2
 
 M, N, O, J, L = 192, 128, 96, 2, 3
```
eickenberg: I see the `shape` is used within the function above for creating the scattering object. So this example runs, right? (I'm fine keeping MNO here until going over this to create a narrative)

eickenberg:
Yes, Looks good to me!

eickenberg:
The commit is more of a `REF`actoring than an ` ENH`ancement

eickenberg:
Alright, merging this

janden:
> The commit is more of a `REF`actoring than an ` ENH`ancement

Is it? Since it breaks the API, it doesn't feel like a refactoring, exactly... NumPy devs tag such commits with `API`, come to think of it. May be a good idea.

eickenberg:
Ok good point. Let's call such commits API henceforth

Issue #196, opened 2018-11-25
Inconsistency in disabling second order


lostanlen:
The order of the scattering transform is integer in 1d (`max_order=2`), boolean in 2d (`order2=True`), decided at forward runtime in 3d (`order_2=True`).
I would propose to unify this and make it integer, so that third-order scattering could, in principle, be implemented in the future.
I don't love the name `max_order`. The `max` prefix makes it sound like it's signal-adaptive, whereas it's a static choice.
I suggest `n_orders=2`.
If we want to have different outputs at runtime, we can always have kwarg `orders=[1,2]` in `forward`, but that's another story.

lostanlen:
Either we do it now, either we postpone it to `0.2.0` along with a deprecation, and remove the deprecation in `0.3.0`. Thoughts?

lostanlen:
> Not sure about n_orders. Note that current max_order=2 corresponds to n_orders=3 since we have orders zero, one, and two.
Oh my, that's a good point.
Come to think of it, I like `max_order` more than `n_orders`. Because the `orders` can be changed at runtime between `0` and `max_order-1`. Shall I make a PR?

janden:
> Come to think of it, I like `max_order` more than `n_orders`.

Ok great!

> Because the `orders` can be changed at runtime between `0` and `max_order-1`.

Am I missing something here or do you mean `orders` can contain values between `0` and `max_order`?

> Shall I make a PR?

Yes please!

lostanlen:
> Am I missing something here or do you mean orders can contain values between 0 and max_order?

No, you're right. Anywhere between `0` and `max_order`. And it defaults to `range(0, 1+max_order)`. 
I got bitten by EWD 831 again

eugenium:
wait your telling me I can do Scattering2d(max_order=0)?  lol

lostanlen:
@eugenium I mean, it's a pretty contrived way to compute a Gaussian blur, but who's to say that there will never be a use case for it? :p

janden:
> wait your telling me I can do Scattering2d(max_order=0)? lol

Why not? I think it works in `Scattering1D` already. Go ahead, try it for yourself!

lostanlen:
@eickenberg the kwarg `orders` can be added later.
here we do not have one because we only officially support `orders=range(0, 1+max_order)`
but this kwarg is useful for:
* skipping the zeroth order (in audio for example)
* skipping the second order without having to construct a whole new operator

[...]

PR #212, opened 2018-11-28
REF bool order2 -> int max_order in Scattering2D

lostanlen:
a contribution to solving #196 in the case of 2D
more consistency between 1D and 2D
enabling the development of third-order scattering in the future

### DIFF
--- a/kymatio/scattering1d/scattering1d.py
+++ b/kymatio/scattering1d/scattering1d.py
@@ -177,7 +177,7 @@ class Scattering1D(object):
     eps : float, optional
         Required precision for the periodization. Defaults to `1e-7`.
     max_order : int, optional
-        The maximum order of scattering coefficients to compute. Must be one
+        The maximum order of scattering coefficients to compute. Must be either
         `1` or `2`. Defaults to `2`.
     average : boolean, optional
         Determines whether the output is averaged in time or not. The averaged
--- a/kymatio/scattering1d/scattering1d.py
+++ b/kymatio/scattering1d/scattering1d.py
@@ -459,8 +459,8 @@ class Scattering1D(object):
             The number of first-order wavelets per octave (second-order wavelets
             are fixed to one wavelet per octave).
         max_order : int, optional
-            The maximum order of scattering coefficients to compute. Must be one
-            `1` or `2`. Defaults to `2`.
+            The maximum order of scattering coefficients to compute.
+            Must be either equal to `1` or `2`. Defaults to `2`.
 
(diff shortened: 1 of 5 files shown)


kymatio/scattering1d/scattering1d.py
```diff
@@ -178,8 +178,7 @@ class Scattering1D(object):
         Required precision for the periodization. Defaults to `1e-7`.
     max_order : int, optional
         The maximum order of scattering coefficients to compute. Must be one
-        `1` or `2`. This parameter may be modified after object creation.
-        Defaults to `2`.
+        `1` or `2`. Defaults to `2`.
```
janden: This should be in a separate PR (or at least a separate commit).

kymatio/scattering2d/scattering2d.py
```diff
@@ -48,6 +48,9 @@ class Scattering2D(object):
         spatial support (M, N) of the input
     L : int, optional
         number of angles used for the wavelet transform
+    max_order : int, optional
+        The maximum order of scattering coefficients to compute. Must be one
```
janden: "Must be one" -> "Must be one of"

kymatio/scattering2d/scattering2d.py
```diff
@@ -61,6 +64,9 @@ class Scattering2D(object):
...
```
janden: Missing "of" (see above).

kymatio/scattering1d/scattering1d.py
lostanlen: ok. I'll revert this and make a separate PR.
I'd like to think of `max_order` as an immutable and then maybe in the future we could have `orders` as mutable, which defaults to `range(1+max_order)` and needs to be a subset thereof. this is why i removed this sentence: i did not want to offer the guarantee that `max_order` will remain mutable

janden: That makes sense. It should be tied into the filter generation, after all.

Not here, but we may want to rename this to `_max_order` so people won't mess with it, but I'm not sure how these things are normally done. There's no encapsulation in Python, right?

lostanlen: that's right. nothing is private in Python, but naming an object attribute with a leading underscore is widespread convention for scaring away the user to change these values after construction. the same can be said of methods
(and for C numba bindings, the convention is *two* leading underscores! :D )

janden: Ok.

PR #215, opened 2018-11-29
ENH Make 1D input shapes `(B, T)`, not `(B, 1, T)`

### LINKED ISSUE #213: 1D needs a (B,T) input rather than (B, 1, T)
originally pointed by @janden in #158 

> 1D: Expects inputs of size (B, 1, T), outputs (B, P, T/2**J), where P is the number of scattering coefficients. Note that if the second dimension of the input is not 1, it errors.

That way, the `ndims` of output should be 1 + `ndims` of input, across modules.

janden:
Previously, inputs to the scattering transform had to have a
channel axis that was always of dimension one. This is no longer
the case. An input of shape `(B, T)` is converted to a scattering
output of shape `(B, C, T1)`, where `C` is the number of scattering
coefficients and `T1` is the subsampled length of the signal.

Fixes #213.

### DIFF
--- a/examples/1d/compute_speed.py
+++ b/examples/1d/compute_speed.py
@@ -78,7 +78,7 @@ if backend.NAME == 'skcuda' and torch.cuda.is_available():
 
 scattering = Scattering1D(J, T, Q)
 
-x = torch.randn(batch_size, 1, T, dtype=torch.float32)
+x = torch.randn(batch_size, T, dtype=torch.float32)
 
 ###############################################################################
 # Run the benchmark
--- a/examples/1d/plot_classif.py
+++ b/examples/1d/plot_classif.py
@@ -105,7 +105,7 @@ path_dataset = info_data['path_dataset']
 ###############################################################################
 # Set up Tensors to hold the audio signals (`x_all`), the labels (`y_all`), and
 # whether the signal is in the train or test set (`subset`).
-x_all = torch.zeros(len(files), 1, T, dtype=torch.float32)
+x_all = torch.zeros(len(files), T, dtype=torch.float32)
 y_all = torch.zeros(len(files), dtype=torch.int64)
 subset = torch.zeros(len(files), dtype=torch.int64)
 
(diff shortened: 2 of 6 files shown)


lostanlen:
LGTM

lostanlen:
this change disallows `(B, C, T)`. but it is something we can add further down the line in a backwards-compatible way.
the consistency we're promoting here is that the output of `Scattering*D` has one more dimension than the input.
in 1D: `2 -> 3` (was `3 -> 3` before this change)
in 2D: `4 -> 5`
in 3D: `4 -> 5`

generalizing this to `n -> (n+1)` in all dimensions is left as future work

eickenberg:
Alright, then this is good to go

janden:
> But can this code handle `(B, C, T)`, now, too, or are we postponing that?

Yes I realize in retrospect it would have been easy to go all the way and allow `(..., T)` to `(..., C, T1)`, that is, to allow the "batch" dimensions to have any shape. Will add an issue and to remind us to do this later.

lostanlen:
@janden FWIW  i find the fact that 2-D * requires* a color channel dimension equally concerning, if not more, than the fact that 1-D and 3-D disallows channel dimensions. there are many applications of 2-D scattering that do not involve a color channel. But yes, let's open this issue and mark it `0.2.0`

eickenberg:
It is good that these ideas are crystallizing.

The other way of doing this project would have been to write a unified
thing from scratch instead of collecting together existing code bases, but
I do think the approach we took has led to a lot of valuable thoughts about
structure and API and gives us direction for moving forward.
Had we started with the unified thing we would have probably gotten it
wrong + incurred performance drops due to insistence on elegance rather
than practical aspects.

But yeah, to be addressed moving forward

janden:
> It is good that these ideas are crystallizing. The other way of doing this project would have been to write a unified thing from scratch instead of collecting together existing code bases, but I do think the approach we took has led to a lot of valuable thoughts about structure and API and gives us direction for moving forward. Had we started with the unified thing we would have probably gotten it wrong + incurred performance drops due to insistence on elegance rather than practical aspects. But yeah, to be addressed moving forward

There are merits to both. For what it's worth, ScatNet was built following the second philosophy, with all the pains that entailed. I think this has been working out relatively well.

PR #217, opened 2018-11-30
FIX Make `Scattering3D` work for `J != 2`

janden:


kymatio/scattering3d/tests/test_scattering3d.py
```diff
@@ -78,6 +78,19 @@ def test_against_standard_computations():
             x, order_2=True, method='integral',
             integral_powers=integral_powers)
 
+        # WARNING: These are hard-coded values for the setting J = 2.
+        n_order_1 = 3
+        n_order_2 = 3
+
+        # Extract orders and make order axis the slowest in accordance with
+        # the stored reference scattering transform.
+        order_1 = orders_1_and_2[:,:,0:n_order_1,:]
```
eickenberg: Can you walk me through this procedure? 

`orders_1_and_2` comes out of the scattering transform and is of shape `(batch,  ?,  stacked order + scale, ?)` (might be good to have output shapes in the docstring because I can't infer them from looking at the code)

The orders get separated into `order1` and `order2`, which both get reshaped into `batch, -1`. and the concatenated along axis 1.

I am having troubles following how this is different from just doing `orders_1_and_2.reshape(batch, -1)`, since there was no permuting of axes in between.

What am I missing?

janden: Doing `orders_1_and_2.reshape(batch, -1)` will give you a memory layout of batch, integral order, order, scale, then ell. Now the stored coefficients are ordered with a layout of batch, order, integral order, scale, ell. If the number of scales is independent of order (which it is if `J=2`), we can reshape to a Tensor of size `(batch_size, n_integral_orders, n_orders, n_scales, n_ells)`, permute axes 1 and 2, then do the `reshape(batch_size, -1)`, as you say. What the code does above is do this in a safer way by manually separating out the different orders from the mixed scale + order axis.

janden: Yeah. Like I said, it's a mess, but it'll do for now.

Issue #218, opened 2018-11-30
Make inputs dimension-agnostic (1D/2D)


janden:
In other words, make `Scattering*D` operators act only on last dimension without forcing a specific convention on previous dimensions. The output would have the same shape, but this the time/spatial dimension subsampled and another (scattering) dimension added right before it. So we'd have

- in 1D, `batch_shape + (T,)` input gives `batch_shape + (C, T/2**J)` output, and
- in 2D, `batch_shape + (N, M)` input gives `batch_shape + (C, N/2**J, M/2**J)` output,

where `batch_shape` is some (possibly empty) tuple specifying the batch dimensions that are not to be processed and `C` is the number of scattering coefficients.

The 3D case is more complicated since it follows a different output schema compared to the 1D and 2D, which will be addressed in a separate issue. That being said, it would be good if it could mimic the above behavior.

lostanlen:
Thanks for raising this Joakim. For future reference, this originated in PR #215 which closed issue #213, which itself was derived off #158.
Quoting Semver 7:
> Minor version Y (x.Y.z | x > 0) MUST be incremented if new, backwards compatible functionality is introduced to the public API. 

Therefore this change should be marked as `0.2.0`, not `0.1.0-beta`.

janden:
Even if it's backwards-compatible, huh? Didn't know that.

Ok so unless we can push this into `0.1.0-alpha`, `0.2.0` it is. It honestly shouldn't be that hard.

lostanlen:
https://semver.org is an insightful read.
during `0.y.z` phase:
* backwards-incompatible API changes push to `0.(y+1).z`
* backwards-compatible API changes push to `0.(y+1).z`
* bugfix changes push to `0.y.(z+1)`

during `x.y.z` phase where `x>0`:
* backwards-incompatible API changes push to `(x+1).y.z`
* backwards-compatible API changes push to `x.(y+1).z`
* bugfix changes push to `x.y.(z+1)`

janden:
Ok makes sense. Good thing we don't have `x>0` then. Sounds like a pain....

PR #221, opened 2018-11-30
ENH Make transforms batch shape agnostic

janden:
This addresses #218, making the transforms accept any shape input, as long as the last `d` dimensions line up with the shape for which the transform is defined, where `d` is the dimension of the transform.

### DIFF
--- a/kymatio/scattering1d/scattering1d.py
+++ b/kymatio/scattering1d/scattering1d.py
@@ -355,11 +355,17 @@ class Scattering1D(object):
             is `False`, it is a dictionary indexed by tuples of filter indices.
         """
         # basic checking, should be improved
-        if len(x.shape) != 2:
+        if x.dim() < 1:
             raise ValueError(
-                'Input tensor x should have 2 axis, got {}'.format(
-                    len(x.shape)))
-        x = x.unsqueeze(1)
+                'Input tensor x should have at least one axis, got {}'.format(
+                    x.dim()))
+
+        batch_shape = x.shape[:-1]
+        signal_shape = x.shape[-1:]
+
+        batch_length = int(np.prod(batch_shape))
+
+        x = x.view((batch_length, 1) + signal_shape)
 
         # get the arguments before calling the scattering
         # treat the arguments
@@ -379,6 +385,15 @@ class Scattering1D(object):
                        oversampling=self.oversampling,
                        vectorize=self.vectorize,
                        size_scattering=size_scattering)
+
+        if self.vectorize:
+            scattering_shape = S.shape[-2:]
+            S = S.view(batch_shape + scattering_shape)
+        else:
+            for k, v in S.items():
+                scattering_shape = v.shape[-2:]
+                S[k] = v.view(batch_shape + scattering_shape)
+
         return S
 
     def __call__(self, x):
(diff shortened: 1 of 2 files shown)


janden:
I'm going to leave this here for comments before I proceed with 2D. As discussed in #218, 3D will have to wait for a different PR.

kymatio/scattering1d/scattering1d.py
```diff
@@ -355,11 +355,17 @@ def forward(self, x):
             is `False`, it is a dictionary indexed by tuples of filter indices.
         """
         # basic checking, should be improved
-        if len(x.shape) != 2:
```
eickenberg: I prefer `len(x.shape)` because `.dim` is very pytorch-bound (the numpy equivalent is `.ndim`)

kymatio/scattering1d/scattering1d.py
```diff
@@ -355,11 +355,17 @@ def forward(self, x):
...
-        x = x.unsqueeze(1)
+                'Input tensor x should have at least one axis, got {}'.format(
+                    x.dim()))
+
+        batch_shape = x.shape[:-1]
+        signal_shape = x.shape[-1:]
+
+        batch_length = int(np.prod(batch_shape))
+
+        x = x.view((batch_length, 1) + signal_shape)
```
eickenberg: `reshape` should probably be fine and is more framework-agnostic.

Also I'd remove `batch_length` in favor of `-1` if it isn't used anywhere else

kymatio/scattering1d/scattering1d.py
```diff
@@ -379,6 +385,15 @@ def forward(self, x):
                        oversampling=self.oversampling,
                        vectorize=self.vectorize,
                        size_scattering=size_scattering)
+
+        if self.vectorize:
+            scattering_shape = S.shape[-2:]
+            S = S.view(batch_shape + scattering_shape)
```
eickenberg: reshape creates a view if the array is contiguous and copies if it is not.

`.view()` plus assigning shape in numpy will error if there was non-contiguity and a copy is necessary. In pytorch I wouldn't even be too sure they have already dealt with all the edge cases

kymatio/scattering1d/scattering1d.py
```diff
@@ -379,6 +385,15 @@ def forward(self, x):
...
+        else:
+            for k, v in S.items():
+                scattering_shape = v.shape[-2:]
+                S[k] = v.view(batch_shape + scattering_shape)
```
eickenberg: reshape

kymatio/scattering1d/tests/test_scattering1d.py
```diff
@@ -277,3 +277,46 @@ def test_scattering_shape_input():
         # should invoke the else branch
     assert "1-tuple" in ve.value.args[0]
     assert "integer" in ve.value.args[0]
+
+def test_batch_shape_agnostic():
+    J, Q = 3, 8
+    length = 1024
+    shape = (length,)
+
+    length_ds = length/2**J
```
eickenberg: I know PEP8 allows this, but lots of linters don't get it and ask for spaces between operators

janden:
Will proceed with the same approach for 2D if there are no other comments on this.

lostanlen:
@janden let's merge this! 👏

janden:
Oh no! I still had to do the 2D. Will make a new PR.

edouardoyallon:
> Oh no! I still had to do the 2D. Will make a new PR.

I think it's better to split the PRs anyway..!

janden:
Ok. I will proceed then.

edouardoyallon:
thanks! 😄

PR #226, opened 2018-12-11
ENH Make transforms batch shape agnostic (2D)

### LINKED ISSUE #218: Make inputs dimension-agnostic (1D/2D)

janden:
See #221. Fixes #218.

### DIFF
--- a/kymatio/scattering2d/scattering2d.py
+++ b/kymatio/scattering2d/scattering2d.py
@@ -146,6 +146,10 @@ class Scattering2D(object):
         if not torch.is_tensor(input):
             raise(TypeError('The input should be a torch.cuda.FloatTensor, a torch.FloatTensor or a torch.DoubleTensor'))
 
+        if len(input.shape) < 2:
+            raise (RuntimeError('Input tensor must have at least two '
+                'dimensions'))
+
         if (not input.is_contiguous()):
             raise (RuntimeError('Tensor must be contiguous!'))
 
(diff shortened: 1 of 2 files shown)


janden:
If I don't hear anything by tomorrow, I will merge this.

lostanlen:
This looks good to me. I suppose that the `reshape`s do not produce copies?

janden:
> This looks good to me. I suppose that the `reshape`s do not produce copies?

They do, if necessary. See https://github.com/kymatio/kymatio/pull/221/files/73942eb4adc6e4d42615865ac0c59119f7569804#r238076672. @eickenberg says we should prefer it to `view()` for compatibility purposes.

lostanlen:
Ok. Let's merge this. If we can make it faster / less memory-intensive in the future, we can always refine further without breaking the API.

janden:
> Ok. Let's merge this. 

Ok good. Want to get input from some 2D people as well before going ahead. @edouardoyallon?

> If we can make it faster / less memory-intensive in the future, we can always refine further without breaking the API.

You mean removing the `reshape()`s to avoid copies? I don't think that's going to be possible. There are other places in the code that requires the arrays to be contiguous (e.g., the `skcuda` backend), I think, so if no copy is made here, a copy will be made there.

eickenberg:
This looks great to me. Thanks!

PR #228, opened 2018-12-11
API Create CPU/CUDA interface for 3D

### LINKED ISSUE #220: CUDA<->CPU interface for 3D
Continuation of the discussion from #122. It would be good to have a way to set the `Scattering3D` object to CPU or CUDA modes, just like the 1D and 3D codes with `cpu()` and `cuda()` functions that mutate the object. Note that calling `cuda()` doesn't have to make the object move its filters to GPU (like the 1D and 2D code currently does), but simply gives it the option to. This may be better in 3D due to the memory constraints.

Right now, all this could do is throw and error if the input to `forward()` is not compatible (i.e., sending a CPU tensor to a `Scattering3D` object that is in CUDA mode, or vice versa), but it unifies the interfaces and allows for more complex behavior in the future.

janden:
This will make `Scattering3D` behave more like the 1D/2D code and have `cpu()` and `cuda()` methods that make it accept CPU or GPU tensors depending on the state. Right now, this only changes a flag, but it might be useful later to optimize the GPU calculations.

Fixes #220.

### DIFF
--- a/kymatio/scattering3d/scattering3d.py
+++ b/kymatio/scattering3d/scattering3d.py
@@ -40,6 +40,8 @@ class Scattering3D(object):
         self.L = L
         self.sigma_0 = sigma_0
 
+        self.is_cuda = False
+
         self.build()
 
     def build(self):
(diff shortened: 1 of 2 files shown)


janden:
Needs docs. Will add tomorrow.

janden:
Docs are there now. Ready to merge.

edouardoyallon:
Looks good to me!

janden:
Ok great. @eickenberg? You're more familiar with this code.

eickenberg:
Yup, forgot to comment on this -- looks very good to me. I really like the test ! :)

PR #234, opened 2018-12-14
API Default values and remove kwargs (3D)

### LINKED ISSUE #230: Remove kwargs to `forward` in `Scattering3D`
For consistency with 1D and 2D, we should have the keyword arguments be set as attributes instead of arguments to `forward`. Like in 1D, we can have these set in the constructor, if supplied.

### LINKED ISSUE #231: Default arguments for 3D constructor
Currently, neither `L` or `sigma0` have default arguments, which is different from 1D and 2D. For `sigma0`, it makes sense to have a default value of `1`. For `L`, a popular value seems to be `3`. Do we want to fix these as defaults?

janden:
Fixes #231.

[EDIT by @lostanlen: also fixes #230]

### DIFF
--- a/examples/3d/compute_speed.py
+++ b/examples/3d/compute_speed.py
@@ -101,14 +101,17 @@ for device in devices:
     else:
         x = x.cpu()
 
-    scattering.forward(x, method='integral', integral_powers=integral_powers)
+    scattering.method = 'integral'
+    scattering.integral_powers = integral_powers
+
+    scattering.forward(x)
 
     if device == 'gpu':
         torch.cuda.synchronize()
 
     t_start = time.time()
     for _ in range(times):
-        scattering.forward(x, method='integral', integral_powers=integral_powers)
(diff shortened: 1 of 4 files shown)


janden:
Also fixes #230.

edouardoyallon:
should be merged ASAP

eickenberg:
Rebased this, waiting for travis to approve

janden:
> @janden during interactive rebase it looked like some commits removed this line https://github.com/kymatio/kymatio/pull/234/files#diff-c913e2608d62fd77f2f15827eedd4b85R264 (it happens several times throughout other tests too). The tests pass but I am wondering whether you wanted to keep this or not

Those line definitely need to be in there.

The issue is that Travis only tests the torch backend with CPU. To make sure everything is ok, you need to test locally with GPU for both torch and skcuda backends. Removing that line will make the skcuda backend crash since it doesn't support CPU tensors.

PR #235, opened 2018-12-14
API Simplify 1D constructor

### LINKED ISSUE #227: Simplify 1D constructor
Right now, there are several arguments to the `Scattering1D` class that are not used: `criterion_amplitude`, `r_psi`, `sigma0`, `alpha`, `P_max`, `eps`. Another argument, `normalize`, is used in one of the tests, but set to its default value (`'l1'`). While it is good to have flexibility, there is currently no evidence that the defaults give unsatisfactory results. Furthermore, having these options make it harder to follow what's going on and increases the burden for us in maintaining backwards compatibility.

This last point is important because there is likely some reduction and refactoring possible here. For example, `criterion_amplitude` and `eps` both refer to a precision in computing the filter bank, while `P_max` refers to the number of periods to use in computing a filter. These could probably be computed from a single tolerance value. Further, `alpha` is the "aliasing tolerance after subsampling" and may be calculated from the `oversampling` parameter (or vice versa). The other parameters, `r_psi`, `sigma0`, and `normalize`, are more independent, but the question remains how to specify them in a way that's simple and compatible with the other transforms.

Until we have decided how to structure these interfaces, I think it's a good idea to fix these parameters to their default values and bring them back into the interface when it's time.

janden:
Remove some of the fine-tuning parameters for the filters.

Fixes #227.

### DIFF
--- a/kymatio/scattering1d/scattering1d.py
+++ b/kymatio/scattering1d/scattering1d.py
@@ -90,31 +90,6 @@ class Scattering1D(object):
     Q : int >= 1
         The number of first-order wavelets per octave (second-order wavelets
         are fixed to one wavelet per octave).
-    normalize : string, optional
-        Normalization of the wavelets: `'l1'`, corresponding to
-        :math:`\\ell^1`-normalization, or `'l2'`, corresponding to
-        :math:`\\ell^2`-normalization. Defaults to `'l1'`.
-    criterion_amplitude : float, optional
-        Controls the padding size (the larger this value, the smaller the
-        padding size). Measures the amount of the Gaussian mass (in the
-        :math:`\\ell^1` norm) which can be ignored after padding. Defaults
-        to `1e-3`.
-    r_psi : float, optional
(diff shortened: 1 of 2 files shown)


edouardoyallon:
can be merged imho

kymatio/scattering1d/scattering1d.py
```diff
@@ -184,22 +159,26 @@ class Scattering1D(object):
...
         self.Q = Q
-        self.r_psi = r_psi
-        self.sigma0 = sigma0
-        self.alpha = alpha
-        self.P_max = P_max
-        self.eps = eps
-        self.criterion_amplitude = criterion_amplitude
-        self.normalize = normalize
+
+        # Set these default values for now. In the future, we'll want some
```
eickenberg: I'd prefer to see these in `build`

janden: Ok. Thanks.

PR #238, opened 2018-12-14
REF Homogenize 3d output

### LINKED ISSUE #222: 3D output format
Currently, `Scattering3D` takes an input of shape `(B, M, N, O)`, where `B` is the batch size and `M, N, O` is the shape of the input signal. Depending on `method`, the output takes different formats:

- If `method='standard'`, the output is of shape `(B, 2*M*N*O, total_scales, L+1)`, where `total_scales` is the number of scales at all orders, that is `(J+1)+J*(J+1)/2` if `order_2=True`. The factor `2` in the second axis comes from the fact that these are still stored as "complex" values (even though they are real), so every other element along that axis is zero. These zeroes should probably be eliminated.
- If `method='integral'`, the output is of shape `(B, len(integral_powers), total_scales, L+1)`, where `integral_powers` is the argument provided to `forward()`.
- If `method='local'`, the output is of shape `(B, points.shape[1], total_scales, L+1)`, where `points` is the argument provided to `forward()`.

The first question is whether we should provide all these outputs through the same interface. To a certain extent, `'integral'` and `'local'` can be computed from `'standard'`, but this may not be computationally optimal.

The second question is how to structure these outputs to be more consistent with 1D and 2D. I suggest that we adopt a similar output shape `(B, C, M, N, O)`, where the second axis comingles orders, scales, ells, integral powers, and points, as the case may be. Alternatively, for `'integral`' and `'local'`, the third axis could index `integral_powers` or `points` instead of a singleton spatial dimension.

eickenberg:
This doesn't work yet. I have refactored the `standard` output so that it doesn't ravel, but my brain is in a state where I can't get the test data reshaped correctly to make the test pass.

This is the last thing we need to fix before being able to release

[EDIT by @lostanlen: closes #222]

### DIFF
--- a/kymatio/scattering3d/scattering3d.py
+++ b/kymatio/scattering3d/scattering3d.py
@@ -361,7 +361,7 @@ class Scattering3D(object):
                 [torch.stack(s_order_1, dim=2),
                 torch.stack(s_order_2, dim=2)], 1)
         else:
-            return torch.stack(s_order_1, dim=1)
+            return torch.stack(s_order_1, dim=2)
 
 
     __call__ = forward
--- a/kymatio/scattering3d/scattering3d.py
+++ b/kymatio/scattering3d/scattering3d.py
@@ -118,8 +118,7 @@ class Scattering3D(object):
 
         """
         convolved_input = self._low_pass_filter(input_array, self.J)
-        return subsample(convolved_input, self.J).view(
-            input_array.size(0), -1, 1)
+        return subsample(convolved_input, self.J)
 
     def _compute_local_scattering_coefs(self, input_array, points, j):
         """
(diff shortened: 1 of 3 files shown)


janden:
I can take a look at this tonight, but a quick fix is to recalculate the test data and save it do the `test_data_3d.pt` file. It's up to you.

janden:
(we're talking ETA 6-8 hours)

janden:
So the tests are ok now. I also corrected (?) a small bug in order 1 since it was giving a different ordering there. If everything looks good, let's merge!

eickenberg:
Yeah this does look good now! Thanks @janden for fixing the tests and the 1D part!!

eickenberg:
Going to squash because of the test

PR #246, opened 2018-12-16
API Change coeff order

eickenberg:
Done the change. The current state of the PR should fail on travis because tests don't reflect change

### DIFF
--- a/kymatio/scattering2d/scattering2d.py
+++ b/kymatio/scattering2d/scattering2d.py
@@ -167,7 +167,6 @@ class Scattering2D(object):
         J = self.J
         phi = self.Phi
         psi = self.Psi
-        n = 0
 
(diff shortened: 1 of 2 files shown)


lostanlen:
This is an API-breaking change. Does it make it to `0.1`? If not, it needs to be postponed to the next deprecation cycle ...

eickenberg:
With the modified test it should now pass

eickenberg:
> This is an API-breaking change. Does it make it to 0.1? If not, it needs to be postponed to the next deprecation cycle ...

Then we mess up anybody who is not doing deep learning. And even them if they reuse a saved model

lostanlen:
@eickenberg yes, i would be in favor of including this to 0.1.0-alpha

PR #249, opened 2018-12-16
API Attempt to include zeroth-order in 3D

janden:
This is currently not working, for seemingly non-trivial reasons. The current format (introduced by  #238) specifies batch index, j+order, ell, spatial as the ordering. Since orders one and two all have L+1 values for ell, whereas the zeroth order only has one, it doesn't work.

### DIFF
--- a/kymatio/scattering3d/scattering3d.py
+++ b/kymatio/scattering3d/scattering3d.py
@@ -342,12 +342,16 @@ class Scattering3D(object):
 
         compute_scattering_coefs = self._compute_scattering_coefs
 
+        s_order_0 = []
         s_order_1 = []
         s_order_2 = []
         _input = to_complex(input_array)
 
         method_args = dict(points=self.points,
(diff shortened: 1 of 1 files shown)


eickenberg:
Isn't L=0 the zeroth order?

eickenberg:
I mean `\ell=0`

janden:
> I mean `\ell=0`

I believe so. So order 0 only has one value for `ell`, while order 1 and 2 have `ell = 0, ..., L`.

eickenberg:
I think we might just want to rename this object to `SolidHarmonicScattering3d` and keep its quirks, since the output it currently gives yields reasonable performances on the qm7 dataset with regression.

We can then work on making this as compatible as possible with 3d gabor scattering, once that is added.

janden:
Ok. Works for me.

Looks like zeroth order is already included in the first-order coefficients for `ell = 0`. Closing.

Issue #250, opened 2018-12-16
`Scattering3d` should be renamed


lostanlen:
As pointed out by @eickenberg in https://github.com/kymatio/kymatio/pull/249#issuecomment-447676253, `Scattering3d` is useful for energy regression in quantum chemistry, but has many idiosyncracies (custom spatial sampling, `ell_p` pooling, meaning of `L`, etc.) that cannot be made perfectly consistent with `Scattering1d` or `Scattering2d`.

So it needs a rename ... `SolidHarmonicScattering3d` nicely matches the title of the paper. It's a bit long, but I think that it's fine. It's not a name one would type repeatedly in code, and autocompletion solves that problem in part too.

janden:
Why not `HarmonicScattering3D`? Not all of us are blessed with autocomplete...

janden:
Or like spend 33% of my line length on a class name...

eickenberg:
I'm fine with HarmonicScattering3d. That would allow us to go away from
specifically solid harmonics if we wanted to

janden:
Ok. Are these solid harmonics if it's Gaussian * Y_lm? Wouldn't that be r^l * Y_lm?

lostanlen:
I'm fine with `HarmonicScattering3d`. In fact I'm fine with anything that does not have the name `Spiral` in it :p

eickenberg:
> Wouldn't that be r^l * Y_lm?

Yes, that, sorry.

OH ....

That changes these filters to be not completely  low-pass anymore.

OK still. We are parking this thing in `HarmonicScattering3D` and will clean it up bit by bit. We will be starting from something that gives reasonable results on qm7, so we have a guideline/benchmark to work with.

PR #251, opened 2018-12-16
REF change name of Scattering3D to HarmonicScattering3D everywhere

eickenberg:
Addresses #250

### DIFF
--- a/examples/3d/README.txt
+++ b/examples/3d/README.txt
@@ -2,4 +2,4 @@
 ===========
 
 These scripts demonstrate the capabilities of the 3D scattering transform
-class, ``Scattering3D``.
+class, ``HarmonicScattering3D``.
--- a/examples/3d/compute_speed.py
+++ b/examples/3d/compute_speed.py
@@ -28,10 +28,10 @@ import time
 import kymatio.scattering3d.backend as backend
 
 ###############################################################################
-# Finally, we import the `Scattering3D` class that computes the scattering
+# Finally, we import the `HarmonicScattering3D` class that computes the scattering
 # transform.
 
-from kymatio import Scattering3D
+from kymatio import HarmonicScattering3D
 ###############################################################################
 # Benchmark setup
 # --------------------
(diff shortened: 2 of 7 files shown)


lostanlen:
LGTM. Note that this changes only the constructor, not the module.
So at the moment your PR is

```
from kymatio.scattering3d import HarmonicScattering3D
from .scattering3d.scattering3d import HarmonicScattering3D
```

And Gabor would potentially be

```
from kymatio.scattering3d import Scattering3D
```

Do I understand correctly?

eickenberg:
Yeah, that was the idea -- I think it won't clutter the module to have two files/classes in it

lostanlen:
Right. The thing that is slightly confusing is that the `3d` in scattering `3d` is used three times:
1) as a submodule name: `kymatio.scattering3d` vs `kymatio.scattering1d`
2) as a Python file name `kymayio.scattering3d.scattering3d` vs `kymatio.scattering3d.utils`
3) as a class name: `kymatio.scattering3d.scattering3d.Scattering3d` vs `kymatio.scattering3d.scattering3d.HarmoniScattering3d`

so I'm wondering if the second one of these shouldn't be renamed too

eickenberg:
This was done on purpose: so that nobody can miss that we do `Scattering`, we scattered `scattering` everywhere

lostanlen:
renamed to something like `network.py` or `transform.py` or `compute.py` or `constructors.py` or ...
you get the idea.
Something that is clearly orthogonal to `filter_bank.py`, but not necessarily orthogonal to 1d vs 2d vs 3d, nor to harmonic vs Gabor

lostanlen:
> This was done on purpose: so that nobody can miss that we do Scattering, we scattered scattering everywhere

Hahaha.
You might say I'm splitting hairs, but this has implications for further development of `TimeFrequencyScattering`, or `RotoTranslationScattering2D`, etc.

eickenberg:
Should I rename the `scattering3d.py` file to `harmonic_scattering3d.py`? I thought about it and think I shouldn't.

lostanlen:
> Should I rename the scattering3d.py file to harmonic_scattering3d.py? I thought about it and think I shouldn't.

See my comment on your PR. #251

PR #276, opened 2018-12-20
Update index.rst

eugenium:


lostanlen:
thanks but:
1) the change needs to be reflected in README
2) the sentence "if you need a library that" does not play well with "differentiable transforms". it would be good to make this grammatically correct across bullet points
3) i would prefer to avoid exclamation points

PR #293, opened 2019-01-15
FIX Make `max_order = 1` work properly for 1D

### LINKED ISSUE #292: BUG 1D scattering does not respect `max_order`
Specifically, if we set `max_order = 1`, it creates the same size output as for `max_order = 2`, but fills the remaining elements with zero.

janden:
Specifically, don't allocate a tensor of the same size as for
`max_order = 2`.

Fixes #292.

### DIFF
--- a/kymatio/scattering1d/scattering1d.py
+++ b/kymatio/scattering1d/scattering1d.py
@@ -511,7 +511,10 @@ class Scattering1D(object):
                 if j2[n2] > j1[n1]:
                     size_order2 += 1
         if detail:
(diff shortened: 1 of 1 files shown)


lostanlen:
thanks! maybe it would be good to also write a test to check that we don't ever regress on this issue?

janden:
> thanks! maybe it would be good to also write a test to check that we don't ever regress on this issue?

It's coming... in another PR. I was working on a PR to extend test coverage which is how this came up.

PR #302, opened 2019-01-16
Python 3.5 compatibility

ethiy:
The python 3.5 seems to work quite well with this module.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,5 @@ doc/source/gallery_1d
 doc/source/gallery_2d
 doc/source/gallery_3d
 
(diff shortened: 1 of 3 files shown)


eickenberg:
Hi @ethiy,
thanks for checking this! It's true that the current code works with python 3.5. However, moving forward, we will definitely break compatibility to python 3.5 (which incidentally will not be supported by future versions of many other standard packages), so we have decided not to advertise that we officially support it in the first place.
In any case, thanks for checking out the package, which I hope will be useful to you, and for reporting!

PR #313, opened 2019-01-18
Minor proofreading of user guide

f0k:
Things that bugged me while reading the user guide. (No need to add me to the contributors list for that...)


lostanlen:
Thanks!

PR #315, opened 2019-01-18
FIX Remove unused import

f0k:
Filterbanks are not cached.


lostanlen:
Thanks Jan!
Yes, this is a follow-up of #203 which closed #175
Caching those filter banks is a long-term plan, but we need to think a bit about how we want to do that. We might learn from the experience of librosa in this regard; but that's another story.
LGTM

janden:
Yes, thank you @f0k.

If you don't mind, I will add you to the contributors list.

Issue #259, opened 2018-12-17
MAINT+ENH Quality assurance via benchmarking


eickenberg:
Our codebase is prone to evolve quite a bit in the future. With every change, it is very easy to incur regressions in speed. Before making major refactorings, such as unifying backends across 1d, 2d, 3d, we need to have systematic benchmarking to check that we are not messing up.

This seems like a good framework with examples online:

https://asv.readthedocs.io/en/stable/

We already have benchmarking scripts called `compute_speed.py` in the examples. To organize this, we should
- Make a folder called `benchmarks`, just like `examples`
- Fill it with benchmarks of the relevant functionalities, including filter generation and forward (and backward) operators

edouardoyallon:
@janden maybe you could refactor following this in order to solve this issue, and add a plugin @ flatiron?

PR #316, opened 2019-01-18
Avoid creating unneeded 2D Psi filters

### LINKED ISSUE #314: Unused 2D Psi filters
The generated 2D filterbank includes filters that are not used in the computation.

Let's create a scattering transform:
```
>>> import kymatio
>>> s = kymatio.Scattering2D(J=2, shape=(32, 32))
```

Let's look at the shapes of the generated filters:
```
>>> print("\n".join('{' + ', '.join('%s: %s' % (k, getattr(v, 'shape', v)) for k, v in psi.items()) + '}' for psi in s.Psi))
{j: 0, theta: 0, 0: torch.Size([40, 40, 2])}
{j: 0, theta: 1, 0: torch.Size([40, 40, 2])}
{j: 0, theta: 2, 0: torch.Size([40, 40, 2])}
{j: 0, theta: 3, 0: torch.Size([40, 40, 2])}
{j: 0, theta: 4, 0: torch.Size([40, 40, 2])}
{j: 0, theta: 5, 0: torch.Size([40, 40, 2])}
{j: 0, theta: 6, 0: torch.Size([40, 40, 2])}
{j: 0, theta: 7, 0: torch.Size([40, 40, 2])}
{j: 1, theta: 0, 0: torch.Size([40, 40, 2]), 1: torch.Size([20, 20, 2])}
{j: 1, theta: 1, 0: torch.Size([40, 40, 2]), 1: torch.Size([20, 20, 2])}
{j: 1, theta: 2, 0: torch.Size([40, 40, 2]), 1: torch.Size([20, 20, 2])}
{j: 1, theta: 3, 0: torch.Size([40, 40, 2]), 1: torch.Size([20, 20, 2])}
{j: 1, theta: 4, 0: torch.Size([40, 40, 2]), 1: torch.Size([20, 20, 2])}
{j: 1, theta: 5, 0: torch.Size([40, 40, 2]), 1: torch.Size([20, 20, 2])}
{j: 1, theta: 6, 0: torch.Size([40, 40, 2]), 1: torch.Size([20, 20, 2])}
{j: 1, theta: 7, 0: torch.Size([40, 40, 2]), 1: torch.Size([20, 20, 2])}
```

The 20x20 filters are all unused. Demonstration:
```
>>> for p in s.Psi:
...     if 1 in p:
...        p[1] = 'foo'
...
>>> import torch
>>> x = torch.rand((1, 1, 32, 32)).cuda()
>>> s = s.cuda()
>>> y = s(x)
```

They should probably not be part of the filterbank in the first place?

f0k:
Maybe this fixes #314?

### DIFF
--- a/kymatio/scattering2d/filter_bank.py
+++ b/kymatio/scattering2d/filter_bank.py
@@ -46,7 +46,7 @@ def filter_bank(M, N, J, L=8):
                 (int(L-L/2-1)-theta) * np.pi / L,
                 3.0 / 4.0 * np.pi /2**j, 4.0/L, offset=offset_unpad)
             psi_signal_fourier = fft2(psi_signal)
-            for res in range(j + 1):
+            for res in range(min(j + 1, J - 1)):
                 psi_signal_fourier_res = periodize_filter_fft(
                     psi_signal_fourier, res)
                 psi[res] = torch.FloatTensor(


lostanlen:
yes this looks like it. thanks! could you please write a test that reproduces #314? in principle, that test would fail on the master branch and pass on your branch. once you have that, we'll rebase and merge the PR.

f0k:
I'm also not quite sure how we would test this -- directly reproducing #314 is not an option, I could only check if the total number of tensors matches the expected number, or if there are no filters with a too large resolution index. Ideally, we would check if all tensors participated in the computation, but I'm not sure how? Can we mark all of them as `requires_grad` and check if they are all referenced in the graph that Pytorch builds?

janden:
> if there are no filters with a too large resolution index

I think this would be the way to do it, but again, I don't think we should have a test for this.

> we would check if all tensors participated in the computation

This is a pretty neat idea, and may be possible to do the way you describe it. This could be combined with some sort of benchmark tests (see #259) to ensure that we're not doing something suboptimal (even if it technically works and passes the tests).

eickenberg:
> Ideally, we would check if all tensors participated in the computation, but I'm not sure how? Can we mark all of them as requires_grad and check if they are all referenced in the graph that Pytorch builds?

I think this is a super smart idea, but I wouldn't overthink it: This change seems to have passed the test of transformation output being the correct one, which is the most important one.
As said by @janden above the only other thing that would be great to have is an insurance that this is not a regression in speed, or, actually more strongly an improvement -- it should be due to the generation of fewer filters (and the current absence of filter caching)

f0k:
So to conclude, there shouldn't be any test added to this PR, but in the long run we'd want #259 to ensure that we don't introduce performance regressions (similarly to how you already test for code coverage not being reduced). Correct? Then this would be good to merge.

PR #317, opened 2019-01-21
Mark support for Python 3.5

f0k:
The README and setup.py say that Python 3.6 is required, but the Travis tests include Python 3.5, so it works fine and should be advertised as such. Ubuntu 18.04 LTS ships with Python 3.5, so installing the current version of kymatio there requires some extra hassle (using pyenv or not-quite-open commercial alternatives such as anaconda) that is actually unneeded.


eickenberg:
cf https://github.com/kymatio/kymatio/pull/302 we were not really going to support 3.5 anymore.

Though what you mention is interesting -- ubuntu18 ships with 3.5 natively?

That makes me half want to reconsider. But then again, most users will install pytorch with anaconda, which currently ships with 3.6 and 3.7.

Any other opinions on this? (I for one really want to be able to move forward with the new functionality added in 3.6, but would be OK with 3.5 if it solves some problems)

janden:
@eickenberg Just out of curiosity, what are the 3.6 features that we'd like to use?

Also, weren't we talking about 2.7 compatibility at some point? Seems like if we're cutting compatibility with 3.5, that'll never happen...

eickenberg:
> what are the 3.6 features that we'd like to use?

f-strings :)

(which is so minor that I'm willing to give up on it for the moment.)

Yes, we had talked about maintaining compatibility (to 3.5, and possibly even down to 2.7 if not too hard), but to not be bound to it by mentioning it officially. That is the main reason why I am hesitant to mention it officially - because then we'll have to drop it officially as well.

But since all this is probably just a formality, I'm happy to go with adding 3.5 into officially supported  if it makes life easier for the ubuntu18 system python case (though I doubt that many users will run into this)

f0k:
> Though what you mention is interesting -- ubuntu18 ships with 3.5 natively?

Yes.

> But then again, most users will install pytorch with anaconda

I don't. It took me some extra effort to find out how to install Python 3.6 under Ubuntu without resorting to commercial distributions. It was interesting to learn, but I would have preferred to just use the system Python and be done with it. And now I found that it was for nothing, because you're not actually dependant on Python 3.6.

> I for one really want to be able to move forward with the new functionality added in 3.6

What 3.6 functionality do you need for kymatio? If you want to build a library for a wide audience, you may want to consider keeping the requirements as low as possible.

janden:
Ok. I think this is a sufficiently strong case for marking 3.5 compatibility. From what I understand, we also need to update `setup.py`. Is there anything else that needs to be done?

Let's move the discussion on 2.7 compatibility to a separate issue.

janden:
If @eickenberg agrees, I suggest we merge this.

f0k:
> From what I understand, we also need to update `setup.py`.

(Yes, this is included in the PR.)

> Is there anything else that needs to be done?

I'll leave that question to you. I'm not aware of anything. The Travis configuration still includes `3.5` anyway. I grepped for `3.6` and found nothing else interesting.

eickenberg:
Yeah, go ahead and merge!

janden:
Sorry. For some reason I thought this PR only dealt with the README (which, curiously, already marks 3.5 as compatible). Merging.

janden:
Sorry. I meant the badge up top, not the requirements section. Everything seems to line up now.

f0k:
> I meant the badge up top, not the requirements section.

Ah, sorry! Now that you mention it, I noticed while grepping for more occurrences of the Python version :)

Issue #305, opened 2019-01-16
ENH overlap-add convolutions


eickenberg:
I think this is standard practice for 1D signals, because they often have to deal with very long signals wrt filter size, but this can be applied across the board: Imagine a very big 2D image, or, more realistically, a 3D array which is actually a video of 2D frames which you want to analyze with spatiotemporal gabors. This will require overlap-add in the time axis.

So the general setup might be something where you specify the axes that you want to have overlap-add in, and it could be a `WaveletTransform` or `OverlapAddConvolution` module that takes care of it.

lostanlen:
Yes. This is something that is available in scattering.m for example and it's really useful

janden:
The other potential advantage here, from what I understand, is that it may give us a way to compute filters independent of image size, since we're storing them in the spatial/time domain.

eickenberg:
Working with 1D data of shape `(384, 300000)` is showing me the limitations of not having this implemented. My GPU maxes out in memory at `(30, 300000)`, which, if we made this into one long line, would correspond to like 400s of mono audio at 22050. It might be nice to be able to go a bit above that without having to manually chunk data etc

lostanlen:
As a bioacoustician, i routinely work on input waveforms that are a billion samples long or more. Overlap-add is essential in this case. And 👍 for working independently of image size. This opens the path for training fully convolutional models and recurrent models

MuawizChaudhary:
Maybe one of you can explain the basic concept, and I can try to implement this one of these days?

lostanlen:
[copied from a Discord discussion]

N>>T is the typical use case of variable-length input. 

To address it, pick a buffer size M>T  (M is the maximal support of the wavelets/phi). M depends solely on the Scattering1D object, not on the input itself.

Then, given a signal x of length N>>M, divide x into C overlapping chunks of length M. I think that a default overlap of 50% is sensible.
Note that padding is only needed for the first and last chunk. For all the chunks in between, we can use overlap as a form of "natural padding".

In scattering.m i have experimented with two variants of this chunking procedure: "Hann" and "Tukey", after the name of the window functions applied to every chunk. In both variants, the number of scattering frames per chunk should be a multiple of 4. (so N=4T sounds about right?)

Here's the layout i imagine: 
```
chunk A: 0 1 2 3
chunk B:     0 1 2 3
chunk C:         0 1 2 3
chunk D:             0 1 2 3
```

[...]

PR #320, opened 2019-01-22
ENH Use real filterbank in Scattering2D

f0k:
The Morlet/Gabor filterbank in Scattering2D is actually real-valued when expressed in the spectral domain. We can exploit this in the computation. (Possibly needs a test to make sure all code paths in cdgmm are covered.)

### DIFF
--- a/kymatio/scattering2d/backend/backend_skcuda.py
+++ b/kymatio/scattering2d/backend/backend_skcuda.py
@@ -30,6 +30,10 @@ def iscomplex(input):
     return input.size(-1) == 2
 
 
+def isreal(input):
+    return input.size(-1) == 1
+
+
 class Pad(object):
     def __init__(self, pad_size, pre_pad=False):
         """
@@ -252,9 +256,9 @@ def cdgmm(A, B, inplace=False):
         Parameters
         ----------
         A : tensor
-            input tensor with size (B, C, M, N, 2)
+            A is a complex tensor of size (B, C, M, N, 2)
         B : tensor
-            B is a complex tensor of size (M, N, 2)
+            B is a complex tensor of size (M, N, 2) or real tensor of (M, N, 1)
         inplace : boolean, optional
             if set to True, all the operations are performed inplace
 
@@ -265,32 +269,34 @@ def cdgmm(A, B, inplace=False):
             C[b, c, m, n, :] = A[b, c, m, n, :] * B[m, n, :]
     """
     A, B = A.contiguous(), B.contiguous()
-    if A.size()[-3:] != B.size():
+    if A.size()[-3:-1] != B.size()[:-1]:
         raise RuntimeError('The filters are not compatible for multiplication!')
 
-    if not iscomplex(A) or not iscomplex(B):
-        raise TypeError('The input, filter and output should be complex')
+    if not iscomplex(A):
+        raise TypeError('The input must be complex')
+
+    if not iscomplex(B) and not isreal(B):
+        raise TypeError('The filter must be complex or real')
 
     if B.ndimension() != 3:
-        raise RuntimeError('The filters must be simply a complex array!')
+        raise RuntimeError('The filter must be a 3-tensor')
 
     if type(A) is not type(B):
         raise RuntimeError('A and B should be same type!')
 
-    if not A.is_cuda:
-        raise RuntimeError('Use the torch backend for cpu tensors!')
-
-    C = A.new(A.size()) if not inplace else A
-    m, n = B.nelement() // 2, A.nelement() // B.nelement()
-    lda = m
-    ldc = m
-    incx = 1
-    handle = torch.cuda.current_blas_handle()
-    stream = torch.cuda.current_stream()._as_parameter_
-    cublas.cublasSetStream(handle, stream)
-    cublas.cublasCdgmm(handle, 'l', m, n, A.data_ptr(), lda, B.data_ptr(), incx, C.data_ptr(), ldc)
-    return C
-
-
-
-
+    if isreal(B):
+        if inplace:
+            return A.mul_(B)
+        else:
+            return A * B
+    else:
+        C = A.new(A.size()) if not inplace else A
+        m, n = B.nelement() // 2, A.nelement() // B.nelement()
+        lda = m
+        ldc = m
+        incx = 1
+        handle = torch.cuda.current_blas_handle()
+        stream = torch.cuda.current_stream()._as_parameter_
+        cublas.cublasSetStream(handle, stream)
+        cublas.cublasCdgmm(handle, 'l', m, n, A.data_ptr(), lda, B.data_ptr(), incx, C.data_ptr(), ldc)
+        return C
--- a/kymatio/scattering2d/backend/backend_skcuda.py
+++ b/kymatio/scattering2d/backend/backend_skcuda.py
@@ -281,8 +281,11 @@ def cdgmm(A, B, inplace=False):
     if B.ndimension() != 3:
         raise RuntimeError('The filter must be a 3-tensor')
 
-    if type(A) is not type(B):
-        raise RuntimeError('A and B should be same type!')
+    if A.dtype is not B.dtype:
+        raise RuntimeError('A and B must be of the same dtype')
+
+    if A.device != B.device:
+        raise RuntimeError('A and B must be on the same device')
 
     if isreal(B):
         if inplace:
--- a/kymatio/scattering2d/backend/backend_skcuda.py
+++ b/kymatio/scattering2d/backend/backend_skcuda.py
@@ -268,7 +268,6 @@ def cdgmm(A, B, inplace=False):
             output tensor of size (B, C, M, N, 2) such that:
             C[b, c, m, n, :] = A[b, c, m, n, :] * B[m, n, :]
     """
-    A, B = A.contiguous(), B.contiguous()
     if A.size()[-3:-1] != B.size()[-3:-1]:
         raise RuntimeError('The filters are not compatible for multiplication!')
 
@@ -293,6 +292,7 @@ def cdgmm(A, B, inplace=False):
         else:
             return A * B
     else:
+        A, B = A.contiguous(), B.contiguous()
         C = A.new(A.size()) if not inplace else A
         m, n = B.nelement() // 2, A.nelement() // B.nelement()
         lda = m
--- a/kymatio/scattering2d/backend/backend_skcuda.py
+++ b/kymatio/scattering2d/backend/backend_skcuda.py
@@ -269,7 +269,7 @@ def cdgmm(A, B, inplace=False):
             C[b, c, m, n, :] = A[b, c, m, n, :] * B[m, n, :]
     """
     A, B = A.contiguous(), B.contiguous()
-    if A.size()[-3:-1] != B.size()[:-1]:
+    if A.size()[-3:-1] != B.size()[-3:-1]:
         raise RuntimeError('The filters are not compatible for multiplication!')
 
     if not iscomplex(A):
--- a/kymatio/scattering2d/backend/backend_skcuda.py
+++ b/kymatio/scattering2d/backend/backend_skcuda.py
@@ -268,17 +268,21 @@ def cdgmm(A, B, inplace=False):
             output tensor of size (B, C, M, N, 2) such that:
             C[b, c, m, n, :] = A[b, c, m, n, :] * B[m, n, :]
     """
-    if A.size()[-3:-1] != B.size()[-3:-1]:
-        raise RuntimeError('The filters are not compatible for multiplication!')
-
     if not iscomplex(A):
-        raise TypeError('The input must be complex')
-
-    if not iscomplex(B) and not isreal(B):
-        raise TypeError('The filter must be complex or real')
+        raise TypeError('The input must be complex, indicated by a last '
+                        'dimension of size 2')
 
     if B.ndimension() != 3:
-        raise RuntimeError('The filter must be a 3-tensor')
+        raise RuntimeError('The filter must be a 3-tensor, with a last '
+                           'dimension of size 1 or 2 to indicate it is real '
+                           'or complex, respectively')
+
+    if not iscomplex(B) and not isreal(B):
+        raise TypeError('The filter must be complex or real, indicated by a '
+                        'last dimension of size 2 or 1, respectively')
+
+    if A.size()[-3:-1] != B.size()[-3:-1]:
+        raise RuntimeError('The filters are not compatible for multiplication!')
 
     if A.dtype is not B.dtype:
         raise RuntimeError('A and B must be of the same dtype')
--- a/kymatio/scattering2d/backend/backend_torch.py
+++ b/kymatio/scattering2d/backend/backend_torch.py
@@ -178,7 +178,7 @@ def cdgmm(A, B, inplace=False):
             C[b, c, m, n, :] = A[b, c, m, n, :] * B[m, n, :]
     """
     A, B = A.contiguous(), B.contiguous()
-    if A.size()[-3:-1] != B.size()[:-1]:
+    if A.size()[-3:-1] != B.size()[-3:-1]:
         raise RuntimeError('The filters are not compatible for multiplication!')
 
     if not iscomplex(A):
--- a/kymatio/scattering2d/backend/backend_torch.py
+++ b/kymatio/scattering2d/backend/backend_torch.py
@@ -177,17 +177,21 @@ def cdgmm(A, B, inplace=False):
             output tensor of size (B, C, M, N, 2) such that:
             C[b, c, m, n, :] = A[b, c, m, n, :] * B[m, n, :]
     """
-    if A.size()[-3:-1] != B.size()[-3:-1]:
-        raise RuntimeError('The filters are not compatible for multiplication!')
-
     if not iscomplex(A):
-        raise TypeError('The input must be complex')
-
-    if not iscomplex(B) and not isreal(B):
-        raise TypeError('The filter must be complex or real')
+        raise TypeError('The input must be complex, indicated by a last '
+                        'dimension of size 2')
 
     if B.ndimension() != 3:
-        raise RuntimeError('The filter must be a 3-tensor')
+        raise RuntimeError('The filter must be a 3-tensor, with a last '
+                           'dimension of size 1 or 2 to indicate it is real '
+                           'or complex, respectively')
+
+    if not iscomplex(B) and not isreal(B):
+        raise TypeError('The filter must be complex or real, indicated by a '
+                        'last dimension of size 2 or 1, respectively')
+
+    if A.size()[-3:-1] != B.size()[-3:-1]:
+        raise RuntimeError('The filters are not compatible for multiplication!')
 
     if A.dtype is not B.dtype:
         raise RuntimeError('A and B must be of the same dtype')
--- a/kymatio/scattering2d/backend/backend_torch.py
+++ b/kymatio/scattering2d/backend/backend_torch.py
@@ -10,6 +10,10 @@ def iscomplex(input):
     return input.size(-1) == 2
 
 
+def isreal(input):
+    return input.size(-1) == 1
+
+
 class Pad(object):
     def __init__(self, pad_size, pre_pad=False):
         """
@@ -161,9 +165,9 @@ def cdgmm(A, B, inplace=False):
         Parameters
         ----------
         A : tensor
-            input tensor with size (B, C, M, N, 2)
+            A is a complex tensor of size (B, C, M, N, 2)
         B : tensor
-            B is a complex tensor of size (M, N, 2)
+            B is a complex tensor of size (M, N, 2) or real tensor of (M, N, 1)
         inplace : boolean, optional
             if set to True, all the operations are performed inplace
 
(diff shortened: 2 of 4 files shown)


f0k:
> (Possibly needs a test to make sure all code paths in cdgmm are covered.)

I've rewritten the cdgmm tests to cover everything. Some of these techniques may also be useful for other tests.

f0k:
> Out of curiosity, did you benchmark this?

Not systematically. As part of a larger pipeline it improved speed by 50% or so with the pytorch backend.

> The Morlet/Gabor filterbank in Scattering2D is actually real-valued when expressed in the spectral domain.

Note that for the filterbank generated by kymatio this is only approximately true. For the case I looked at, the real part is 7 to 8 orders of magnitude larger than the imaginary part, and dropping the imaginary part hardly affects the results (obviously they are also close enough to pass the existing scattering test).

Does anybody have an insight on whether it's generally safe to ignore the imaginary part? Should we be creating the filterbank in the spectral domain? And why does `gabor_2d` [add up 25 parts](https://github.com/kymatio/kymatio/blob/02f31a7/kymatio/scattering2d/filter_bank.py#L186-L191)? All these questions may be related :)

As a curiosity, note that the existing cdgmm test [explicitly used a real-only filter](https://github.com/kymatio/kymatio/blob/02f31a7/kymatio/scattering2d/tests/test_scattering2d.py#L116)...

janden:
> As part of a larger pipeline it improved speed by 50% or so with the pytorch backend.

That would be huge. Will run some tests.

> For the case I looked at, the real part is 7 to 8 orders of magnitude larger than the imaginary part.

Since most of these are single precision, `10**(-8)` sounds like machine precision and could just be rounding errors from FFTs. That can almost certainly be discarded.

> Does anybody have an insight on whether it's generally safe to ignore the imaginary part?

I guess this comes down to whether we expect all our filters to be Hermitian symmetric in the spatial domain. So far, they all have been, but this is not required for scattering transforms. This would be my only hesitation: locking us down in terms of the filters we can use in the future. On the other hand, if we need to, we can just put a check. Right now, real filters may be the way to go.

> Should we be creating the filterbank in the spectral domain?

This is probably not optimal, but not for the above reason. Taking an IDFT of the filters as is would actually require more space (since they're complex) if we don't do something clever (like only storing the positive half and mirroring the negative half when needed). Now the reason we may want to store things in the spatial domain has more to do with #305. We could potentially reduce computation speed and memory usage by storing these in the spatial domain with a limited support, preferably with a multirate filterbank cascade.

> And why does gabor_2d add up 25 parts?

It looks like this may be to periodize things properly. We don't actually want just a Gaussian when computing the Gabor filter, but a periodized Gaussian. Turns out doing five copies in each dimension is enough (I think?). The 1D code has this as a parameter, but I think hardcoding this as done in 2D is the way to go.

> As a curiosity, note that the existing cdgmm test explicitly used a real-only filter...

Yes the tests leave a bit to be desired. Work in progress...

edouardoyallon:
I think this is a great idea atm it still allows using complex filters: the curent set of wavelet isn't indeed real in Fourier however there are lots more wavelets that aren't!

f0k:
> This would be my only hesitation: locking us down in terms of the filters we can use in the future.

This PR still allows both. If the last dimension of the filter tensor is of size 2, it is assumed to be complex, and if it is 1, it is assumed to be real. You can even mix real and complex filters if you want to.

>> Should we be creating the filterbank in the spectral domain?

> This is probably not optimal, but not for the above reason. Taking an IDFT of the filters as is would actually require more space (since they're complex) if we don't do something clever (like only storing the positive half and mirroring the negative half when needed).

I think you may have misunderstood. Currently `gabor_2d()` creates a complex spatial-domain filter, which is later [passed through `fft2()`](https://github.com/kymatio/kymatio/blob/86a7dd6/kymatio/scattering2d/filter_bank.py#L48) to obtain "the filterbank" (the psi and phi torch tensors used in the forward pass). We could refactor this to have `gabor_2d()` directly create a real spectral-domain filter instead and not wonder about the imaginary part. Then its behaviour would also be closer to its docstring, [which says it returns a `morlet_fft`](https://github.com/kymatio/kymatio/blob/86a7dd6/kymatio/scattering2d/filter_bank.py#L177) ;)

> Turns out doing five copies in each dimension is enough (I think?).

It may even be overkill. So it's the same as computing a filter 5 times as large in each dimension and then folding it, right?

janden:
> This PR still allows both.

Ok good.

> I think you may have misunderstood. Currently `gabor_2d()` creates a complex spatial-domain filter, which is later [passed through `fft2()`](https://github.com/kymatio/kymatio/blob/86a7dd6/kymatio/scattering2d/filter_bank.py#L48) to obtain "the filterbank" (the psi and phi torch tensors used in the forward pass). We could refactor this to have `gabor_2d()` directly create a real spectral-domain filter instead and not wonder about the imaginary part. Then its behaviour would also be closer to its docstring, [which says it returns a `morlet_fft`](https://github.com/kymatio/kymatio/blob/86a7dd6/kymatio/scattering2d/filter_bank.py#L177) ;)

Ok. Yes I misunderstood. I thought `gabor_2d()` already computed things directly in Fourier. I agree that that makes more sense.

> It may even be overkill.

It may be. We should check. If the extra copies at -2 and +2 times the period contribute less than machine precision, we should discard them. That would speed things up quite a bit.

> So it's the same as computing a filter 5 times as large in each dimension and then folding it, right?

Yes, it should be.

kymatio/scattering2d/filter_bank.py
```diff
@@ -36,35 +36,39 @@ def filter_bank(M, N, J, L=8):
...
+            psi_signal_fourier = np.real(psi_signal_fourier)
             for res in range(min(j + 1, J - 1)):
                 psi_signal_fourier_res = periodize_filter_fft(
                     psi_signal_fourier, res)
-                psi[res] = torch.FloatTensor(
-                    np.stack((np.real(psi_signal_fourier_res),
-                    np.imag(psi_signal_fourier_res)), axis=2))
+                # add a trailing singleton dimension to mark it as non-complex
+                psi_signal_fourier_res = psi_signal_fourier_res[..., np.newaxis]
+                psi[res] = torch.FloatTensor(psi_signal_fourier_res)
```
edouardoyallon: the reason numpy is used here is to make the filters backend agnostic. I agree that right now, the software is not ready for a tf integration, however I believe it is a good practice to keep numpy here

kymatio/scattering2d/filter_bank.py
```diff
@@ -36,35 +36,39 @@ def filter_bank(M, N, J, L=8):
...
     filters['phi'] = {}
-    phi_signal = gabor_2d(M, N, 0.8 * 2**(J-1), 0, 0, offset=offset_unpad)
+    phi_signal = gabor_2d(M, N, 0.8 * 2**(J-1), 0, 0)
     phi_signal_fourier = fft2(phi_signal)
+    # drop the imaginary part, it is zero anyway
+    phi_signal_fourier = np.real(phi_signal_fourier)
     filters['phi']['j'] = J
     for res in range(J):
         phi_signal_fourier_res = periodize_filter_fft(phi_signal_fourier, res)
-        filters['phi'][res] = torch.FloatTensor(np.stack(
```
edouardoyallon: I didn't remember there was some tensor used here; it'd be better to employ some numpy structures for the reason mentionned above

kymatio/scattering2d/backend/backend_skcuda.py
```diff
@@ -254,7 +254,7 @@ def cdgmm(A, B, inplace=False):
         A : tensor
             input tensor with size (B, C, M, N, 2)
```
edouardoyallon: description could be phrased as for B

kymatio/scattering2d/filter_bank.py
f0k: It created a FloatTensor before and it creates a FloatTensor afterwards. I don't like it, but changing this should be another PR.

janden: Yeah I would tend to agree with @f0k. Let's keep things separate.

f0k:
Updated to address @edouardoyallon's comments (except for the numpy/torch story which is outside the goal of this PR).

f0k:
> I thought gabor_2d() already computed things directly in Fourier. I agree that that makes more sense.

I opened #322 for this.

> f the extra copies at -2 and +2 times the period contribute less than machine precision, we should discard them. That would speed things up quite a bit.

Although the filterbank creation is not really a bottleneck. It will only be created once.

janden:
> I opened #322 for this.

Thank you.
 
> Although the filterbank creation is not really a bottleneck. It will only be created once.

Right, so we can afford to at least get machine precision here, but at the same time, there's no reason to take more time than necessary.

kymatio/scattering2d/filter_bank.py
janden: If @edouardoyallon agrees, I suggest we mark this and the other comment as resolved.

edouardoyallon: goahead; i'll open an issue about that if nobody does it!

janden: Beat you to it.

kymatio/scattering2d/backend/backend_skcuda.py
```diff
@@ -265,32 +269,37 @@ def cdgmm(A, B, inplace=False):
...
+        raise RuntimeError('A and B must be of the same dtype')
 
+    if A.device != B.device:
+        raise RuntimeError('A and B must be on the same device')
 
+    if isreal(B):
+        if inplace:
+            return A.mul_(B)
+        else:
+            return A * B
```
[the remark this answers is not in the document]
eickenberg: though technically, since this is the skcuda backend, we can't really expect such high-level ideas to work here.

I'm fine with adding this here for the moment even though it is incongruent, because we have already done this with the fft modules.

When we start adding other front/backends, we might need to revisit this and have it use the `dgmm` function (could be faster anyway if pytorch isn't smart at chunking/reshaping due to the implicit broadcast)

kymatio/scattering2d/backend/backend_torch.py
```diff
@@ -174,31 +178,39 @@ def cdgmm(A, B, inplace=False):
...
+    if A.device != B.device:
+        raise RuntimeError('A and B must be on the same device')
 
-    C[..., 0].view(-1, C.size(-2)*C.size(-3))[:] = A_r * B_r - A_i * B_i
-    C[..., 1].view(-1, C.size(-2)*C.size(-3))[:] = A_r * B_i + A_i * B_r
+    if isreal(B):
+        if inplace:
+            return A.mul_(B)
+        else:
+            return A * B
```
eickenberg: In this concise formulation there could be a lesson for cleaning up the code below :)

(not for now, for another time, to be benchmarked because of the contiguity issues)

kymatio/scattering2d/filter_bank.py
```diff
@@ -188,10 +192,10 @@ def gabor_2d(M, N, sigma, theta, xi, slant=1.0, offset=0, fft_shift=False):
             [xx, yy] = np.mgrid[offset + ex * M:offset + M + ex * M, offset + ey * N:offset + N + ey * N]
             arg = -(curv[0, 0] * np.multiply(xx, xx) + (curv[0, 1] + curv[1, 0]) * np.multiply(xx, yy) + curv[
                 1, 1] * np.multiply(yy, yy)) + 1.j * (xx * xi * np.cos(theta) + yy * xi * np.sin(theta))
-            gab = gab + np.exp(arg)
+            gab += np.exp(arg)
```
eickenberg: :clap: :pray:

kymatio/scattering2d/tests/test_scattering2d.py
```diff
@@ -107,38 +107,61 @@ def test_SubsampleFourier():
...
+        # call cdgmm
+        if inplace:
+            x = x.clone()
+        z = backend.cdgmm(x, filter, inplace=inplace)
+        if inplace:
+            z = x
+        # compare
+        assert (y - z).abs().max() < 1e-6
+
+    @pytest.mark.parametrize("backend", backends)
```
eickenberg: :open_mouth: :open_mouth: 
I did not know this was a thing.
This solves so much awkwardness.
(It looks like it would pay for me to look at the docs of pytest more thoroughly!)

kymatio/scattering2d/filter_bank.py
```diff
@@ -36,35 +36,39 @@ def filter_bank(M, N, J, L=8):
...
             psi = {}
             psi['j'] = j
             psi['theta'] = theta
             psi_signal = morlet_2d(M, N, 0.8 * 2**j,
                 (int(L-L/2-1)-theta) * np.pi / L,
-                3.0 / 4.0 * np.pi /2**j, 4.0/L, offset=offset_unpad)
+                3.0 / 4.0 * np.pi /2**j, 4.0/L)
             psi_signal_fourier = fft2(psi_signal)
+            # drop the imaginary part, it is zero anyway
+            psi_signal_fourier = np.real(psi_signal_fourier)
```
eickenberg: Q: Do we ever not want this?

I can't think of a reason other than:

We are inside a function that creates a filter bank by calling a filter function repeatedly.

If we had more than one type of filters, this function would likely be reused, pointing to the fact that either it should be optional to set the imaginary part to 0, or it should already be taken care of by the filters.

Which leads to modifying the filter generator function to ask to be able to have a fourier version of it returned (regardless of how the function generates it)

So we have two options here: Keep this logic as it is, and definitely be required to move it upon refactoring for multiple wavelet types, or push the fft into the filter generation now because we know it needs to happen eventually/soon.

Any thoughts on this? I might be willing to go with the former, but it is pretty clear that the abovementioned refactoring will have to happen pretty soon

kymatio/scattering2d/backend/backend_skcuda.py
```diff
@@ -265,32 +269,37 @@ def cdgmm(A, B, inplace=False):
...
-    lda = m
-    ldc = m
-    incx = 1
-    handle = torch.cuda.current_blas_handle()
-    stream = torch.cuda.current_stream()._as_parameter_
-    cublas.cublasSetStream(handle, stream)
-    cublas.cublasCdgmm(handle, 'l', m, n, A.data_ptr(), lda, B.data_ptr(), incx, C.data_ptr(), ldc)
-    return C
+    if not iscomplex(B) and not isreal(B):
+        raise TypeError('The filter must be complex or real')
```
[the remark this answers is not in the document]
janden: I may be missing something, but what happens if `(M, N) = (M, 2)`? Granted, it's a weird example, but maybe should be checked.

kymatio/scattering2d/backend/backend_skcuda.py
```diff
@@ -265,32 +269,37 @@ def cdgmm(A, B, inplace=False):
             C[b, c, m, n, :] = A[b, c, m, n, :] * B[m, n, :]
     """
     A, B = A.contiguous(), B.contiguous()
-    if A.size()[-3:] != B.size():
+    if A.size()[-3:-1] != B.size()[-3:-1]:
```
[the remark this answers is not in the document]
janden: This is yet another thing that's in the "unofficial" style guide. We really need to put all of this in one place.

kymatio/scattering2d/backend/backend_skcuda.py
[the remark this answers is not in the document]
f0k: > `"The filter must be complex or real, which means that its last dimension must be of size 2 or 1 to indicate it"`

Sounds good to me. I kept the style of the existing error messages, but I also prefer to be more explicit.

> what happens if `(M, N) = (M, 2)`?

The very next line checks that B is a 3-tensor.

kymatio/scattering2d/backend/backend_skcuda.py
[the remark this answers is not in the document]
f0k: > since this is the skcuda backend, we can't really expect such high-level ideas to work here.

But the skcuda backend is still pytorch. There's no need to go low-level. At least for now, the skcuda backend means "you *also* need scikit-cuda", not "you need scikit-cuda instead of pytorch". Is this planned to change?

> `dgmm` function (could be faster anyway

That would be an argument. (But then pytorch should be using it as well.)

kymatio/scattering2d/backend/backend_skcuda.py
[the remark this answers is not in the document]
f0k: I also prefer `shape`, but I didn't want to mess with the existing style. A style guide and cleanup sounds good. (Also a guide for the commit prefixes you use.)

kymatio/scattering2d/backend/backend_torch.py
```diff
@@ -174,31 +178,39 @@ def cdgmm(A, B, inplace=False):
             C[b, c, m, n, :] = A[b, c, m, n, :] * B[m, n, :]
     """
     A, B = A.contiguous(), B.contiguous()
-    if A.size()[-3:] != B.size():
+    if A.size()[-3:-1] != B.size()[-3:-1]:
```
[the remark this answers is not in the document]
f0k: Agreed!

kymatio/scattering2d/backend/backend_torch.py
f0k: Definitely. I found the code below barely readable, but now it's unused so I had no strong motivation to rewrite it.

kymatio/scattering2d/filter_bank.py
f0k: (Note the commit message to it. This was not for performance, but for keeping the dtype.)

kymatio/scattering2d/filter_bank.py
eickenberg: > I don't understand, are you saying that we'd have morlet_2d generate the (real) Fourier transform directly?

Up to (future) discussion, but yes, essentially that.

The idea would be to delegate the appearance in Fourier to the filter generation function in a kwarg `fourier=True`. If that kwarg isn't present we could also come up with a decorator to add it and make a normal complex fft be returned.
The reason for this is that it seems that the filter function would "know best" how to present its fourier transform, i.e. whether its fourier transform is purely real or not. (We could envision something similar for filters that are purely real in signal domain as well, but first things first)

The filterbank generation function could then just ask the filter function to return its fourier version and not worry about knowing/guessing the properties of the filters.

janden: Ok. Taking this idea to its extreme, it seems that we need a `Filter` object. At the end of the day, we don't care about how a filter is represented, only that we can apply it to some array efficiently. This application can be in Fourier (real or complex) or it can be spatial (when we get to that). Depending on the size of the input, we may want to do one or the other. Ultimately, this logic should reside with the filter, not with the `Scattering` objects.

janden:
Testing this with `compute_speed.py` on a V100, I get:
```
==> Testing Float32 with torch backend, on GPU, forward
Elapsed time: 27.925186 [s / 10 evals], avg: 2.79 (s/batch)
==> Testing Float32 with skcuda backend, on GPU, forward
Elapsed time: 4.951094 [s / 10 evals], avg: 0.50 (s/batch)
```
on the current master (86a7dd68b1d9b7b524334e93d2e13af43da671cf) and
```
==> Testing Float32 with torch backend, on GPU, forward
Elapsed time: 20.333319 [s / 10 evals], avg: 2.03 (s/batch)
==> Testing Float32 with skcuda backend, on GPU, forward
Elapsed time: 4.972970 [s / 10 evals], avg: 0.50 (s/batch)
```
for this PR (29fc49549ec7bfc294b4d9308cdeb2d793a20897). Does this seem consistent with the modifications (e.g., we should not expect an improvement for the skcuda backend)?

f0k:
> good to merge for me, but would appreciate the error message pedagogy improvement

Finally done!

> Does this seem consistent with the modifications (e.g., we should not expect an improvement for the skcuda backend)?

Hmm. Actually I would have expected an improvement for skcuda as well. If there isn't any, then the skcuda complex cdgmm is as fast as the torch noncomplex multiplication (which is possible; maybe none of them are saturating the GPU yet).

janden:
> the skcuda complex cdgmm is as fast as the torch noncomplex multiplication (which is possible; maybe none of them are saturating the GPU yet).

I think you're right. I've been playing around a bit (increasing the size of the filters, replacing the torch multiplication with `cublas.sdgmm`, etc), and I don't see an improvement for the real multiplication. Since it's just an elementwise multiplication, it's entirely possible that we're already memory-bound. Performing a multiplication between tensors of size `(1024, 3, 256, 256, 2)` and `(256, 256, 2)` takes about 4 ms for me, which translates to an effective memory bandwidth of around 800 GB/s. The GPU is rated at 900 GB/s, so we may very well be saturating (the measurement is done using `%timeit`, which may account for the 10% discrepancy).

Could you (or someone else) run the benchmark on your end to make sure there's not something weird about my configuration? Then I think we're good to merge.

eickenberg:
> Could you (or someone else) run the benchmark on your end to make sure there's not something weird about my configuration? Then I think we're good to merge.

You mean `compute_speed.py` on master and this branch?

janden:
> You mean compute_speed.py on master and this branch?

Correct.

janden:
Also, before merging, we need to decide how we're going to do the branching for `0.2.X`. Seems like this would qualify for that?

janden:
@edouardoyallon, can you run `compute_speed.py` to make sure you see the speedups I'm seeing?

edouardoyallon:
> @edouardoyallon, can you run `compute_speed.py` to make sure you see the speedups I'm seeing?

Alright, it'll be done by friday 6pm et

janden:
> Alright, it'll be done by friday 6pm et

Wonderful. By that time, we will hopefully know whether this goes in 0.1 or 0.2... @lostanlen?

edouardoyallon:
Alright, so, with the version developed by @f0k:

```
==> Testing Float32 with skcuda backend, on GPU, forward
Elapsed time: 10.545391 [s / 10 evals], avg: 1.05 (s/batch)
==> Testing Float32 with torch backend, on GPU, forward
Elapsed time: 26.511131 [s / 10 evals], avg: 2.65 (s/batch)
```

with the previous one,

```
==> Testing Float32 with skcuda backend, on GPU, forward
Elapsed time: 10.701863 [s / 10 evals], avg: 1.07 (s/batch)
==> Testing Float32 with torch backend, on GPU, forward
Elapsed time: 43.686787 [s / 10 evals], avg: 4.37 (s/batch)
```

Can we first merge #345 and then rebase this one?

janden:
:+1:

> Can we first merge #345, #346 and then rebase this one?

Sounds like a good plan. I will take a look tomorrow, if that's ok.

janden:
Rebasing on `master` to remove conflicts and force Jenkins build. If you need to work on this branch, please reset.

edouardoyallon:
> Rebasing on `master` to remove conflicts and force Jenkins build. If you need to work on this branch, please reset.

okay, it starts to be critic to include those PRs..

eickenberg:
So this seems good to go other than merge ordering then

f0k:
Thanks for merging!

PR #327, opened 2019-02-05
ENH+WIP Make skcuda backend differentiable for 2D

janden:
Integrates the 2D skcuda backend into PyTorch's autodifferentiation framework to allow for differentiating scattering transforms with respect to the input signal.

### DIFF
--- a/kymatio/scattering2d/backend/backend_skcuda.py
+++ b/kymatio/scattering2d/backend/backend_skcuda.py
@@ -248,24 +248,6 @@ def fft(input, direction='C2C', inverse=False):
 class cdgmmMul(Function):
     @staticmethod
     def forward(ctx, A, B):
-        """
-        Complex pointwise multiplication between (batched) tensor A and tensor B.
-
-        Parameters
-        ----------
-        A : tensor
-            input tensor with size (B, C, M, N, 2)
-        B : tensor
-            B is a complex tensor of size (M, N, 2)
-        inplace : boolean, optional
-            if set to True, all the operations are performed inplace
-
-        Returns
-        -------
-        C : tensor
-            output tensor of size (B, C, M, N, 2) such that:
-            C[b, c, m, n, :] = A[b, c, m, n, :] * B[m, n, :]
-        """
         A, B = A.contiguous(), B.contiguous()
         if A.size()[-3:] != B.size():
             raise RuntimeError('The filters are not compatible for multiplication!')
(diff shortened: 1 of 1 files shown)


janden:
So we need a test for differentiability here to make sure we're getting the right results. The 1D code has this, but neither 2D or 3D do. I suggest adding a 2D test be part of this PR (3D can wait).

sixin-zh:
This is another test that you may use:
import torch
from torch.autograd import gradcheck

nb=1
nc=1
M=2
N=2
A = torch.randn((nb,nc,M,N,2),dtype=torch.float32).cuda()
B = torch.randn((M,N,2),dtype=torch.float32).cuda()

from kymatio.scattering2d import backend

#C = cdgmm(A,B)
A.requires_grad_(True)
B.requires_grad_(True)
result = gradcheck(cdgmm, (A,B), eps=1e-4, atol=1e-2, rtol=1e-3)
if result:
    print('pass')
else:
    print('not pass')

janden:
Thank you. That seems to run ok (although it gives me a warning). It still doesn't explain why the other test fails, though. This operation is basically linear in its inputs, so if I take the sum of the output, the gradient of one variable should be equal to the value of the other variable. That is what's verified in the check that I added.

More to the point, why am I getting a different gradient when running `backend_torch.cdgmm` compared to `backend_skcuda.cdgmm` on the same data?

sixin-zh:
I looked into the problem, there is actually a storage continuity bug when using cdgmm, 
I fixed this problem at
https://github.com/sixin-zh/kymatio/blob/master/kymatio/scattering2d/backend/backend_skcuda.py#L294
Can you take this code and test it again?

janden:
Great! Can you commit the change to this PR? You first have to add my repository and check out the corresponding branch (`diff_skcuda_2d`):
```
git remote add janden git@github.com:janden/kymatio.git
git checkout diff_skcuda_2d
```
Then make the changes, commit, and run `git push`. After that, the new commit should appear in this PR.

You should also be able to run the tests using
```
KYMATIO_BACKEND=skcuda pytest
```
if you have this branch checked out.

janden:
Ok. This seems to work now (passes the new test).

There's still some work to be done for the other primitives (modulus, subsampling), so I will leave this open for now. If any brave soul wants to take a crack at it, be my guest.

sixin-zh:
You may take a look at 
https://github.com/sixin-zh/kymatio/blob/master/kymatio/scatrec2d/scattering2d_pp.py#L132
for the modulus, subsampling which are differentiable 

best

janden:
Ok. From what I can tell, that file uses the `Modulus` and `SubsampleFourier` classes in [backend_skcuda.py](https://github.com/sixin-zh/kymatio/blob/69b77e229b053a7bc228474aedc429b2d2ce0b9e/kymatio/scatrec2d/backend/backend_skcuda.py), which don't seem to implement the autodifferentiation functionality. Am I missing something here?

sixin-zh:
I’d say it works, e.g.
https://github.com/sixin-zh/kymatio/blob/master/examples/rec2d/test_rec_dirac2d_lbfgsb.py

在 2019年2月8日，上午11:14，Joakim Andén <notifications@github.com<mailto:notifications@github.com>> 写道：


Ok. From what I can tell, that file uses the Modulus and SubsampleFourier classes in backend_skcuda.py<https://github.com/sixin-zh/kymatio/blob/69b77e229b053a7bc228474aedc429b2d2ce0b9e/kymatio/scatrec2d/backend/backend_skcuda.py>, which don't seem to implement the autodifferentiation functionality. Am I missing something here?

janden:
> I’d say it works, e.g. https://github.com/sixin-zh/kymatio/blob/master/examples/rec2d/test_rec_dirac2d_lbfgsb.py

I don't understand. Where is the gradient calculated? There's no `backward` function defined for `Modulus` or `SubsampleFourier`.

sixin-zh:
It is computed by autograd.

在 2019年2月9日，上午7:17，Joakim Andén <notifications@github.com<mailto:notifications@github.com>> 写道：


I’d say it works, e.g. https://github.com/sixin-zh/kymatio/blob/master/examples/rec2d/test_rec_dirac2d_lbfgsb.py

I don't understand. Where is the gradient calculated? There's no backward function defined for Modulus or SubsampleFourier.

eickenberg:
> It is computed by autograd.

Not in `backend_skcuda`

sixin-zh:
you are right, i did not realize we wrote cuda kernel for these operations.
that’s going to be tricky to write backprop. i do not have a good solution for the moment.

在 2019年2月9日，上午7:38，eickenberg <notifications@github.com<mailto:notifications@github.com>> 写道：


It is computed by autograd.

Not in backend_skcuda

edouardoyallon:
Any progress done? If you isolate the problem with the new code, I'd be happy to have a look into it.

janden:
> Any progress done? If you isolate the problem with the new code, I'd be happy to have a look into it.

Might be a while before I get back to this. Essentially, what's missing is the `backward()` function for `Modulus` and `SubsampleFourier`. Shouldn't take more than a few hours of work if someone wants to take a stab at it.

edouardoyallon:
FYI this is a low priority PR/issue, my hope is that torch will become faster later :)

janden:
Given that torch_skcuda will be deprecated when complex numbers are supported in torch, this is no longer relevant and I suggest we close the PR.

Issue #319, opened 2019-01-22
Non-Anaconda testing


janden:
Currently, all CI scripts run through Anaconda. Given that this is not the only way to run Kymatio (supposedly), should we test in other environments as well (i.e., `pip`)?

janden:
For what it's worth, I'm able to get it running using `pipenv`:
```
pipenv --three
pipenv install torch torchvision pytest
pipenv install
pipenv run pytest kymatio
```
Do we think this makes enough of a difference for testing purposes to add another CI script?

edouardoyallon:
I think it is worth to add. Can you add it?
Thanks!
Edouard

PR #328, opened 2019-02-05
TST+WIP Make Travis test a bare `pip` environment

janden:
For compatibility reasons, we also run the test suite in a bare Python environment, installing all the packages using `pip`. This is done in addition to the existing tests, which are run using the `conda` package manager.

Addresses #319.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,36 +1,15 @@
 language: python
+dist: xenial
 python:
-  # We don't actually use the Travis Python, so just put 3.6.
+  - "3.5"
   - "3.6"
+  - "3.7"
 env:
-  - PYTHON_VERSION=3.5
-  - PYTHON_VERSION=3.6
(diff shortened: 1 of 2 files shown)


eickenberg:
very cool. Thanks for taking care of this!

LGTM

janden:
No problem. Thinking about it further, should we even bother testing it on conda? If it runs on a bare `pip` environment, it should run on conda as well, no? If we merge this, all the tests will just take (at least) twice as long, with marginal benefit.

edouardoyallon:
@janden I think it's great and worth to merge, just in case and to simulate a standard user behavior!

janden:
Ok. Want to rebase a little bit to put the script inside a `tools` directory, then will merge.

janden:
Ok. Will make sure it passes and then merge.

PR #329, opened 2019-02-11
ENH adding the number of DL on README.md

edouardoyallon:


janden:
I don't see why not. If no one has any objections, I will merge tomorrow.

lostanlen:
yes, let's do this 👍

PR #338, opened 2019-02-16
ENH Remove unused `getDtype` in `backend_torch` (3D)

janden:


eickenberg:
Yeah this is also very non-standard and brittle-looking. I think we might have more of this in other places where it is actually used.

LGTM

janden:
Yes it's normally part of the skcuda backend code. I agree, it should be looked over.

PR #341, opened 2019-02-17
ENH+WIP Add backend benchmarks

janden:
Lets us benchmark the performance of the backends by measuring their throughput.

### DIFF
--- a/benchmarks/bench_backend_2d.py
+++ b/benchmarks/bench_backend_2d.py
@@ -1,0 +1,65 @@
+import kymatio.scattering2d.backend as backend
+
+import torch
+
+B = 128
+N = 256
+
+trials = 10
+repeats = 3
+
+x = torch.randn(B, 1, N, N, 2)
+y = torch.randn(N, N, 2)
+
+x = x.cuda()
+y = y.cuda()
+
+x = x.contiguous()
+y = y.contiguous()
+
+subsample_factors = (8, 16, 32, 64)
+
+cdgmm = backend.cdgmm
+modulus = backend.Modulus()
+subsample = backend.SubsampleFourier()
+
+name_list = ['cdgmm', 'modulus']
+func_list = [cdgmm, modulus]
+args_list = [(x, y), (x,)]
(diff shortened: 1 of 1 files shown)


janden:
Getting these results for the torch backend on V100:
```
cdgmm:              104.2 GB/s
modulus:            10.83 GB/s
subsample(8):       459.4 GB/s
subsample(16):      567.7 GB/s
subsample(32):      619.5 GB/s
subsample(64):      605.3 GB/s
```
while the skcuda backend gets
```
cdgmm:              769.5 GB/s
modulus:            783   GB/s
subsample(8):       765.9 GB/s
subsample(16):      797.5 GB/s
subsample(32):      650.8 GB/s
subsample(64):      231.4 GB/s
```

So the torch modulus is pretty bad, but it's an easy fix, it looks like (just take the modulus for from the 1D torch backend). The torch cdgmm is also slow, but speeds up quite a bit when we incorporate #320. As @edouardoyallon predicted, the subsample kernel in skcuda slows down when we increase the subsampling factor, but surprisingly, the torch backend has the opposite behavior.

eickenberg:
Thanks for making this benchmark. That's going to be really useful. I think we might want to have a whole folder `benchmarks` for all sorts of benchmark scripts and figure out how to use [asv](https://asv.readthedocs.io/en/stable/) for this

janden:
No problem!

> I think we might want to have a whole folder benchmarks for all sorts of benchmark scripts and figure out how to use asv for this

I thought about that, but there's also #328, which introduces another script that needs to be put in a separate directory (right now, it's just in the root, which I don't think is appropriate). That's why I tentatively put `tools`, where we could put both of them. Having a separate `benchmarks` folder might make sense once we have a few more of these.

janden:
Actually, on second thought, maybe a separate `benchmarks` folder would be better. It's fine if it's not very well-populated for now.

janden:
Putting this in a benchmarks folder and merging. We can add other ones later.

PR #342, opened 2019-02-17
ENH Optimize 2D torch modulus

janden:
Efforts to improve the 2D torch backend go here.

### DIFF
--- a/kymatio/scattering2d/backend/backend_torch.py
+++ b/kymatio/scattering2d/backend/backend_torch.py
@@ -104,10 +104,12 @@ class Modulus(object):
         output: a tensor with imaginary part set to 0, real part set equal to
         the modulus of x.
     """
-    def __call__(self, input):
+    def __call__(self, x):
 
-        norm = input.norm(p=2, dim=-1, keepdim=True)
-        return torch.cat([norm, torch.zeros_like(norm)], -1)
+        norm = torch.zeros_like(x)
+        norm[...,0] = (x[...,0]*x[...,0] +
+                       x[...,1]*x[...,1]).sqrt()
+        return norm
 
 
 


kymatio/scattering2d/backend/backend_torch.py
```diff
@@ -104,10 +104,12 @@ class Modulus(object):
...
         the modulus of x.
     """
-    def __call__(self, input):
+    def __call__(self, x):
 
-        norm = input.norm(p=2, dim=-1, keepdim=True)
-        return torch.cat([norm, torch.zeros_like(norm)], -1)
+        norm = torch.zeros_like(x)
+        norm[...,0] = (x[...,0]*x[...,0] +
+                       x[...,1]*x[...,1]).sqrt()
```
eickenberg: Could it be useful to do some parts inplace?

Either
```
norm[..., 0] = x[..., 0]
norm[..., 0] *= norm[..., 0]
norm[..., 0] += x[..., 1] * x[..., 1]
norm.sqrt_()
```

or

```
norm[:] = x
norm *= norm
norm[..., 0] += norm[..., 1]
norm.sqrt_()
```

kymatio/scattering2d/backend/backend_torch.py
```diff
@@ -104,10 +104,12 @@ class Modulus(object):
         output: a tensor with imaginary part set to 0, real part set equal to
         the modulus of x.
     """
-    def __call__(self, input):
+    def __call__(self, x):
 
-        norm = input.norm(p=2, dim=-1, keepdim=True)
-        return torch.cat([norm, torch.zeros_like(norm)], -1)
```
eickenberg: yeah well this has always been something i cringed at

kymatio/scattering2d/backend/backend_torch.py
janden: Yeah I thought about this. The risk is that splitting it up forces PyTorch to make each a separate kernel, which would slow things down. On the other hand, it may not. Worth benchmarking.

kymatio/scattering2d/backend/backend_torch.py
janden: Yes, and, as it turns out, for good reason!

kymatio/scattering2d/backend/backend_torch.py
janden: Original:
```
modulus:            177.3 GB/s
```
Solution #1:
```
modulus:            99.84 GB/s
```
Solution #2:
```
modulus:            122.1 GB/s
```

So it looks like it's better to try to combine things, when possible.

janden:
Scaling things back, it looks like it's the interleaved addressing's slowing things down for `Modulus()`. Specifically, if we have
```
norm = x[...,0]*x[...,0]
```
we get 789 GB/s, while
```
norm = (x[...,0]*x[...,0] + x[...,1]*x[...,1])
```
gives 263.1 GB/s. Adding in `sqrt` to get
```
norm = (x[...,0]*x[...,0] + x[...,1]*x[...,1]).sqrt()
```
brings us down to 213.8 GB/s. Finally, adding in the zeros with
```
norm = torch.zeros_like(x)
norm[...,0] = (x[...,0]*x[...,0] + x[...,1]*x[...,1]).sqrt()
```
gives 177.2 GB/s.

Seems like torch is having trouble combining the two strided memory accesses.

eickenberg:
Not sure if this explains the differences well, but the different expressions create different numbers of intermediate arrays under the hood.

> `norm = (x[...,0]*x[...,0] + x[...,1]*x[...,1])`

This creates 3 intermediate arrays

> `(x[...,0]*x[...,0] + x[...,1]*x[...,1]).sqrt()`

This creates 4.

The kernel from the `skcuda` backend basically does loop fusion and makes stuff a lot faster.

> norm[...,0] = (x[...,0]*x[...,0] + x[...,1]*x[...,1]).sqrt()

The fact that this is a bit slower than writing to a contiguous array is normal, I think. Definitely something to be aware of, though.


One thing we could do is to check out what exactly `torch.norm` does and maybe, just maybe, think about proposing a specialized `torch.complex_magnitude(x, complex_output=False)` function directly into `pytorch`. I think complex numbers will be used more and more, so it might be interesting to them.

edouardoyallon:
Seems there is this: https://github.com/pytorch/pytorch/issues/755

janden:
> the different expressions create different numbers of intermediate arrays under the hood.

Yeah that makes sense.

> Seems there is this: pytorch/pytorch#755

Ok so once this is merged. We can get to work. Is there some way to watch an issue without getting spammed with all of PyTorch?

janden:
So I had some grandiose plans about doing all sorts of optimizations here but never got around to it. I'm just going to merge this modulus thing and call it a day.

Issue #335, opened 2019-02-16
`Pad` object in torch backend is not tested with `pre_pad = True`


janden:
As a result, [this padding variant](https://codecov.io/gh/kymatio/kymatio/src/6e5c430d71a1562a93931eee0f568a9aba489220/kymatio/scattering2d/backend/backend_torch.py#L31) is not tested.

Issue #284, opened 2019-01-12
Scattering2D doesn't work if 2**J == image_size


Jonas1312:
Hi,

I would like to use Scattering2D with `2**J == img_size` in order to get a (1, K, 1, 1) tensor with K scattering coefficients. However, when I run this script https://gist.github.com/Jonas1312/a0a03c9d3a5307f13ba32348ad8eed5b#file-main-py I get the following error (input image 64x64):

```
  File "D:/Cours/3A/PFE/Python/coefficients_visualization/main.py", line 35, in main
    scattering_coefficients = scattering(src_img_tensor)
  File "C:\Python36\lib\site-packages\kymatio\scattering2d\scattering2d.py", line 234, in __call__
    return self.forward(input)
  File "C:\Python36\lib\site-packages\kymatio\scattering2d\scattering2d.py", line 187, in forward
    U_r = pad(input)
  File "C:\Python36\lib\site-packages\kymatio\scattering2d\backend\backend_torch.py", line 34, in __call__
    out_ = self.padding_module(input)
  File "C:\Python36\lib\site-packages\torch\nn\modules\module.py", line 489, in __call__
    result = self.forward(*input, **kwargs)
  File "C:\Python36\lib\site-packages\torch\nn\modules\padding.py", line 172, in forward
    return F.pad(input, self.padding, 'reflect')
  File "C:\Python36\lib\site-packages\torch\nn\functional.py", line 2685, in pad
    ret = torch._C._nn.reflection_pad2d(input, pad)
RuntimeError: invalid argument 4: Padding size should be less than the corresponding input dimension, but got: padding (64, 64) at dimension 3 of input [1 x 1 x 64 x 64] at c:\a\w\1\s\windows\pytorch\aten\src\thcunn\generic/SpatialReflectionPadding.cu:36

Process finished with exit code 1
```

Thanks

eugenium:
Hi Jonas, we'll have to discuss a bit internally if we will cover this case. However, you can work around this for now by just padding the input image.

PR #346, opened 2019-02-22
FIX bug on the input sizes

### LINKED ISSUE #306: Padding size problem
I was trying the Scattering2D on my data and it did not work. I had this error message: 
>RuntimeError: The filters are not compatible for multiplication!

While debugging, I found out that the padding does not work well with some input shapes. Here is a minimal example:

``` python
import torch
r = torch.randn(1, 1, 45, 45)

scatterer = Scattering2D(J=3, shape=tuple(r.size()[2:]))

from kymatio.scattering2d.backend import fft
print(scatterer.Phi[0].size()) # size -> torch.Size([56, 56, 2])
print(fft(scatterer.pad(r), 'C2C').size()) # this is the 'U_0_c' size -> torch.Size([1, 1, 61, 61, 2])
print(scatterer.forward(r))
```

I have tried other matrix shapes and it seems that the `U_0_c` size, in axis `d=2,3`, is equal to the size of `Phi[0]`, in axis `d-2`, plus the rest of the Euclidian division of the original size, in axis `d-2`, on `2^J`. This causes the later uncatched exception.

Was this unintended? or was it the contrary (which implies that I have to choose a good `J` value)?

If it is the last case, how to choose `J` when I have no guaranty over the input sizes?

edouardoyallon:
Close #306, #284, #335

- [X] Add a short test on several hyper parameters  (common ones e.g., 32, 224 and random values odd)
- [X]  Padding size is now correct even for non-power of 2 sizes !
- [X]  Easier "pre_pad" option management

### DIFF
--- a/kymatio/scattering2d/backend/backend_skcuda.py
+++ b/kymatio/scattering2d/backend/backend_skcuda.py
@@ -47,13 +47,10 @@ class Pad(object):
         self.padding_module = ReflectionPad2d(pad_size)
 
     def __call__(self, input):
-        if(self.pre_pad):
-            output = input.new_zeros(input.size(0), input.size(1), input.size(2), input.size(3), 2)
-            output.narrow(output.ndimension()-1, 0, 1)[:] = input
-        else:
-            out_ = self.padding_module(input)
-            output = input.new_zeros(*(out_.size() + (2,)))
-            output.select(4, 0)[:] = out_
+        if(not self.pre_pad):
+            input = self.padding_module(input)
+        output = input.new_zeros(*(input.size() + (2,)))
+        output.select(4, 0)[:] = input
         return output
 
 def unpad(in_):
--- a/kymatio/scattering2d/backend/backend_skcuda.py
+++ b/kymatio/scattering2d/backend/backend_skcuda.py
@@ -1,6 +1,6 @@
 # Authors: Edouard Oyallon, Sergey Zagoruyko
 
-from collections import defaultdict, namedtuple
+from collections import namedtuple
 import torch
 from skcuda import cublas
 import cupy
@@ -44,13 +44,18 @@ class Pad(object):
                 if set to true, then there is no padding, one simply adds the imaginarty part.
         """
         self.pre_pad = pre_pad
-        self.padding_module = ReflectionPad2d(pad_size)
+        self.pad_size = pad_size
 
-    def __call__(self, input):
-        if(not self.pre_pad):
-            input = self.padding_module(input)
-        output = input.new_zeros(*(input.size() + (2,)))
-        output.select(4, 0)[:] = input
+        self.build()
+
+    def build(self):
+        self.padding_module = ReflectionPad2d(self.pad_size)
+
+    def __call__(self, x):
+        if not self.pre_pad:
+            x = self.padding_module(x)
+        output = x.new_zeros(*(x.size() + (2,)))
+        output.select(4, 0)[:] = x
         return output
 
 def unpad(in_):
--- a/kymatio/scattering2d/backend/backend_skcuda.py
+++ b/kymatio/scattering2d/backend/backend_skcuda.py
@@ -54,7 +54,7 @@ class Pad(object):
     def __call__(self, x):
         if not self.pre_pad:
             x = self.padding_module(x)
-        output = torch.zeros(x.shape + (2,), dtype=x.dtype)
+        output = x.new_zeros(x.shape + (2,))
         output[...,0] = x
         return output
 
--- a/kymatio/scattering2d/backend/backend_skcuda.py
+++ b/kymatio/scattering2d/backend/backend_skcuda.py
@@ -38,7 +38,7 @@ class Pad(object):
 
             Parameters
             ----------
-            pad_size : int
+            pad_size : list of 4 integers
                 size of padding to apply.
             pre_pad : boolean
                 if set to true, then there is no padding, one simply adds the imaginarty part.
@@ -54,8 +54,8 @@ class Pad(object):
     def __call__(self, x):
         if not self.pre_pad:
             x = self.padding_module(x)
-        output = x.new_zeros(*(x.size() + (2,)))
-        output.select(4, 0)[:] = x
+        output = torch.zeros(x.shape + (2,), dtype=x.dtype)
+        output[...,0] = x
         return output
 
 def unpad(in_):
--- a/kymatio/scattering2d/backend/backend_torch.py
+++ b/kymatio/scattering2d/backend/backend_torch.py
@@ -24,13 +24,18 @@ class Pad(object):
                 if set to true, then there is no padding, one simply adds the imaginarty part.
         """
         self.pre_pad = pre_pad
-        self.padding_module = ReflectionPad2d(pad_size)
+        self.pad_size = pad_size
 
-    def __call__(self, input):
-        if(not self.pre_pad):
-            input = self.padding_module(input)
-        output = input.new_zeros(*(input.size() + (2,)))
-        output.select(4, 0)[:] = input
+        self.build()
+
+    def build(self):
+        self.padding_module = ReflectionPad2d(self.pad_size)
+
+    def __call__(self, x):
+        if not self.pre_pad:
+            x = self.padding_module(x)
+        output = x.new_zeros(*(x.size() + (2,)))
+        output.select(4, 0)[:] = x
         return output
 
 def unpad(in_):
--- a/kymatio/scattering2d/backend/backend_torch.py
+++ b/kymatio/scattering2d/backend/backend_torch.py
@@ -34,7 +34,7 @@ class Pad(object):
     def __call__(self, x):
         if not self.pre_pad:
             x = self.padding_module(x)
-        output = torch.zeros(x.shape + (2,), dtype=x.dtype)
+        output = x.new_zeros(x.shape + (2,))
         output[...,0] = x
         return output
 
--- a/kymatio/scattering2d/backend/backend_torch.py
+++ b/kymatio/scattering2d/backend/backend_torch.py
@@ -27,13 +27,10 @@ class Pad(object):
         self.padding_module = ReflectionPad2d(pad_size)
 
     def __call__(self, input):
-        if(self.pre_pad):
-            output = input.new_zeros(input.size(0), input.size(1), input.size(2), input.size(3), 2)
-            output.narrow(output.ndimension()-1, 0, 1)[:] = input
-        else:
-            out_ = self.padding_module(input)
-            output = input.new_zeros(*(out_.size() + (2,)))
-            output.select(4, 0)[:] = out_
+        if(not self.pre_pad):
+            input = self.padding_module(input)
+        output = input.new_zeros(*(input.size() + (2,)))
+        output.select(4, 0)[:] = input
         return output
 
 def unpad(in_):
--- a/kymatio/scattering2d/backend/backend_torch.py
+++ b/kymatio/scattering2d/backend/backend_torch.py
@@ -18,7 +18,7 @@ class Pad(object):
 
             Parameters
             ----------
-            pad_size : int
+            pad_size : list of 4 integers
                 size of padding to apply.
             pre_pad : boolean
                 if set to true, then there is no padding, one simply adds the imaginarty part.
@@ -34,8 +34,8 @@ class Pad(object):
     def __call__(self, x):
         if not self.pre_pad:
             x = self.padding_module(x)
-        output = x.new_zeros(*(x.size() + (2,)))
-        output.select(4, 0)[:] = x
+        output = torch.zeros(x.shape + (2,), dtype=x.dtype)
+        output[...,0] = x
         return output
 
 def unpad(in_):
--- a/kymatio/scattering2d/scattering2d.py
+++ b/kymatio/scattering2d/scattering2d.py
@@ -98,10 +98,10 @@ class Scattering2D(object):
     def build(self):
         self.M, self.N = self.shape
         self.modulus = Modulus()
-        self.pad = Pad(2**self.J, pre_pad = self.pre_pad)
+        self.M_padded, self.N_padded = compute_padding(self.M, self.N, self.J)
+        self.pad = Pad([(self.N_padded-self.N)//2,(self.N_padded-self.N+1)//2,(self.M_padded-self.M)//2,(self.M_padded-self.M+1)//2], pre_pad = self.pre_pad)
         self.subsample_fourier = SubsampleFourier()
         # Create the filters
-        self.M_padded, self.N_padded = compute_padding(self.M, self.N, self.J)
         filters = filter_bank(self.M_padded, self.N_padded, self.J, self.L)
         self.Psi = filters['psi']
         self.Phi = [filters['phi'][j] for j in range(self.J)]
--- a/kymatio/scattering2d/scattering2d.py
+++ b/kymatio/scattering2d/scattering2d.py
@@ -99,7 +99,8 @@ class Scattering2D(object):
         self.M, self.N = self.shape
         self.modulus = Modulus()
         self.M_padded, self.N_padded = compute_padding(self.M, self.N, self.J)
-        self.pad = Pad([(self.N_padded-self.N)//2,(self.N_padded-self.N+1)//2,(self.M_padded-self.M)//2,(self.M_padded-self.M+1)//2], pre_pad = self.pre_pad)
+        # pads equally on a given side if the amount of padding to add is an even number of pixels, otherwise it adds an extra pixel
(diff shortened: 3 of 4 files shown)


edouardoyallon:
- [x] Add a short test on several hyper parameters (common ones e.g., 32, 224 and random values odd)
- [x] Padding size is now correct even for non-power of 2 sizes!

edouardoyallon:
One issue of testing several sizes is that the test becomes substantially longer. On the other hand, to me this is not an issue as those tests have no reasons to go fast..

janden:
> One issue of testing several sizes is that the test becomes substantially longer.

Yeah this is important enough that we can't skip it. On the other hand, it may be enough to do it on a few small sizes (e.g., 31, 32, 33)?

edouardoyallon:
@janden what's the guideline in term of testing time? would be good to have some rules about that I guess (s.t. "a test alpha shouldn't be slower than test beta..")

janden:
> what's the guideline in term of testing time?

I don't know, really. They should be as short as possible while still testing all relevant cases. We should look over the tests in general to see how long they take. I suspect some are a bit too long.

janden:
> You can accelerate your test using python comprehensions

Thank you for the suggestion. What's the difference in performance with comprehensions compared to for loops?

ethiy:
It is usually faster, according to my experience than for loops. To be tested for this suggestions. Otherwise you can parellelize your for loop using `multiprocessing.Pool().map` but that seems a bit too far fetched for a unit test.

janden:
Running `timeit` on
```
def test_input_size_agnostic(): 
    for N in [31,32,33]: 
        for J in [2,4]: 
            scattering = Scattering2D(J, shape=(N, N)) 
            x = torch.zeros(3,3,N,N) 
            S = scattering(x)
```
gets me `5.79 s ± 78.1 ms` while the list comprehension variant gives `6.32 s ± 250 ms`, so they have very similar performance. This seems in line with results reported elsewhere: https://stackoverflow.com/questions/22108488/are-list-comprehensions-and-functional-functions-faster-than-for-loops

I find the for loop variant easier to follow, so I suggest we leave that as is.

edouardoyallon:
@ethiy, @janden Thanks a lot! So since it is slower maybe it nos a good idea to include it here.. yet it's a good tips to know(I ignored this). Also, we should define some days a guideline w.r.t. the style of coding.. However, maybe it could be a great idea to add a multiprocessing library if it is not too difficult?(like running each tests in parallel)

Also, maybe we could merge this branch?

janden:
> Also, maybe we could merge this branch?

Yes we just need a review. I can do it later today.

ethiy:
@janden @edouardoyallon : Can you give us any time table for the merge. I kind of need it for my own research.

kymatio/scattering2d/backend/backend_skcuda.py
```diff
@@ -47,13 +47,10 @@ def __init__(self, pad_size, pre_pad=False):
...
 
     def __call__(self, input):
-        if(self.pre_pad):
-            output = input.new_zeros(input.size(0), input.size(1), input.size(2), input.size(3), 2)
-            output.narrow(output.ndimension()-1, 0, 1)[:] = input
-        else:
-            out_ = self.padding_module(input)
-            output = input.new_zeros(*(out_.size() + (2,)))
-            output.select(4, 0)[:] = out_
+        if(not self.pre_pad):
```
eickenberg: remove parentheses

kymatio/scattering2d/backend/backend_skcuda.py
```diff
@@ -47,13 +47,10 @@ def __init__(self, pad_size, pre_pad=False):
...
+            input = self.padding_module(input)
```
eickenberg: why is there still there variable `input` here. I thought we had purged that. Can you switch to `input_array` or `x` ?

kymatio/scattering2d/backend/backend_torch.py
```diff
@@ -27,13 +27,10 @@ def __init__(self, pad_size, pre_pad=False):
...
 
     def __call__(self, input):
-        if(self.pre_pad):
-            output = input.new_zeros(input.size(0), input.size(1), input.size(2), input.size(3), 2)
-            output.narrow(output.ndimension()-1, 0, 1)[:] = input
-        else:
-            out_ = self.padding_module(input)
-            output = input.new_zeros(*(out_.size() + (2,)))
-            output.select(4, 0)[:] = out_
+        if(not self.pre_pad):
```
eickenberg: remove parentheses

kymatio/scattering2d/backend/backend_torch.py
```diff
@@ -27,13 +27,10 @@ def __init__(self, pad_size, pre_pad=False):
...
+            input = self.padding_module(input)
```
eickenberg: Would be great to not overwrite a python word

kymatio/scattering2d/scattering2d.py
```diff
@@ -98,10 +98,10 @@ def __init__(self, J, shape, L=8, max_order=2, pre_pad=False):
     def build(self):
         self.M, self.N = self.shape
         self.modulus = Modulus()
-        self.pad = Pad(2**self.J, pre_pad = self.pre_pad)
+        self.M_padded, self.N_padded = compute_padding(self.M, self.N, self.J)
+        self.pad = Pad([(self.N_padded-self.N)//2,(self.N_padded-self.N+1)//2,(self.M_padded-self.M)//2,(self.M_padded-self.M+1)//2], pre_pad = self.pre_pad)
```
eickenberg: remove spaces around keyword arguments, add spaces around operators

eickenberg: Can you add a comment explaining what this does?
It looks to me like to pads equally on both sides if the amount of padding to add is an even number of pixels, and will add the extra pixel on the right and the bottom if it's an odd number of pixels. Is this correct? If yes, can you add a comment to that effect? If not, can you add a comment explaining what is correct?

kymatio/scattering2d/backend/backend_torch.py
```diff
@@ -27,13 +27,10 @@ def __init__(self, pad_size, pre_pad=False):
...
+        output = input.new_zeros(*(input.size() + (2,)))
+        output.select(4, 0)[:] = input
```
eickenberg: what does this do? Can you make it less torch-obscure and more numpy-like?

kymatio/scattering2d/backend/backend_torch.py
```diff
@@ -27,13 +27,10 @@ def __init__(self, pad_size, pre_pad=False):
...
-        if(self.pre_pad):
-            output = input.new_zeros(input.size(0), input.size(1), input.size(2), input.size(3), 2)
-            output.narrow(output.ndimension()-1, 0, 1)[:] = input
-        else:
-            out_ = self.padding_module(input)
-            output = input.new_zeros(*(out_.size() + (2,)))
-            output.select(4, 0)[:] = out_
+        if(not self.pre_pad):
+            input = self.padding_module(input)
+        output = input.new_zeros(*(input.size() + (2,)))
```
eickenberg: prefer `.shape`

kymatio/scattering2d/scattering2d.py
```diff
@@ -98,10 +98,10 @@ def __init__(self, J, shape, L=8, max_order=2, pre_pad=False):
     def build(self):
         self.M, self.N = self.shape
         self.modulus = Modulus()
-        self.pad = Pad(2**self.J, pre_pad = self.pre_pad)
```
eickenberg: So the padding module was able to take a number, but below it seems to be taking a list - was this always possible? Asking because I don't see a modification to the padding module itself

kymatio/scattering2d/backend/backend_skcuda.py
```diff
@@ -47,13 +47,10 @@ def __init__(self, pad_size, pre_pad=False):
         self.padding_module = ReflectionPad2d(pad_size)
```
eickenberg: this should technically be in `build`, not `__init__`.

janden:
> Can you give us any time table for the merge.

@eickenberg just reviewed it, so once @edouardoyallon makes the changes, it will be merged. My guess is that it'll be done by the end of the week.

> I kind of need it for my own research.

If you just need to run the code, you don't need to wait for the merge. If you clone the main repository and add @edouardoyallon's repository as a remote, you should be able to check out the branch corresponding to this PR:
```
git clone git@github.com:kymatio/kymatio.git
cd kymatio
git remote add edouardoyallon git@github.com:edouardoyallon/kymatio.git
git fetch edouardoyallon
git checkout -b fixing_multisize_issue edouardoyallon/fixing_multisize_issue
python setup.py develop
```
This is assuming that you already have the dependencies installed.

kymatio/scattering2d/backend/backend_torch.py
edouardoyallon: if s is the size, it creates an array of size (s,2) to handle complex number and set the imaginary part to 0.. not sure how to do make it more numpy-like, I'm not familiar with the standards

kymatio/scattering2d/backend/backend_torch.py
edouardoyallon: well, so i'll need to create a torch.Size() tensor after, no?

kymatio/scattering2d/scattering2d.py
edouardoyallon: Fixed!

kymatio/scattering2d/scattering2d.py
edouardoyallon: it was always allowed by pytorch's standard.

janden:
@eickenberg Let me know if this looks good to you and I'll merge.

edouardoyallon:
OK,that's fixed!

kymatio/scattering2d/backend/backend_skcuda.py
```diff
@@ -38,22 +38,24 @@ def __init__(self, pad_size, pre_pad=False):
 
             Parameters
             ----------
-            pad_size : int
+            pad_size : list of 4 integers
```
eickenberg: or one `int`? Or is that not an option anymore

eickenberg:
Looks good to me!

eickenberg:
Wonder what jenkins has to say -- is the error log public or not?

janden:
> Wonder what jenkins has to say -- is the error log public or not?

It's not, unfortunately.

```
============================= test session starts ==============================
platform linux -- Python 3.6.7, pytest-4.3.0, py-1.8.0, pluggy-0.9.0
rootdir: /home/jenkins/workspace/kymatio_PR-346, inifile:
plugins: cov-2.6.1
collected 40 items

kymatio/scattering1d/tests/test_filters.py .........                     [ 22%]
kymatio/scattering1d/tests/test_scattering1d.py .........                [ 45%]
kymatio/scattering1d/tests/test_utils.py .....                           [ 57%]
kymatio/scattering2d/tests/test_scattering2d.py ...F..                   [ 72%]
kymatio/scattering3d/tests/test_scattering3d.py ...........              [100%]

=================================== FAILURES ===================================
______________________________ test_Scattering2D _______________________________

    def test_Scattering2D():
        test_data_dir = os.path.dirname(__file__)
        data = torch.load(os.path.join(test_data_dir, 'test_data_2d.pt'))
    
        x = data['x']
        S = data['Sx']
        J = data['J']
    
        # we need to reorder S from interleaved (how it's saved) to o0, o1, o2
        # (which is how it's now computed)
    
        o0, o1, o2 = reorder_coefficients_from_interleaved(J, L=8)
        reorder = torch.from_numpy(np.concatenate((o0, o1, o2)))
        S = S[..., reorder, :, :]
    
        pre_pad = data['pre_pad']
    
        M = x.shape[2]
        N = x.shape[3]
    
        import kymatio.scattering2d.backend as backend
    
        if backend.NAME == 'skcuda':
            print('skcuda backend tested!')
            # First, let's check the Jit
            scattering = Scattering2D(J, shape=(M, N), pre_pad=pre_pad)
            scattering.cuda()
            x = x.cuda()
            S = S.cuda()
            y = scattering(x)
            assert ((S - y)).abs().max() < 1e-6
        elif backend.NAME == 'torch':
            # Then, let's check when using pure pytorch code
            scattering = Scattering2D(J, shape=(M, N), pre_pad=pre_pad)
            Sg = []
    
            for device in devices:
                if device == 'gpu':
                    print('torch-gpu backend tested!')
                    x = x.cuda()
                    scattering.cuda()
                    S = S.cuda()
>                   Sg = scattering(x)

kymatio/scattering2d/tests/test_scattering2d.py:209: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
kymatio/scattering2d/scattering2d.py:235: in __call__
    return self.forward(input)
kymatio/scattering2d/scattering2d.py:192: in forward
    U_1_c = subsample_fourier(cdgmm(U_0_c, phi[0]), k=2**J)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

A = tensor([[[[[ 1.1445e+01,  0.0000e+00],
           [ 1.9044e+01, -1.1014e+02],
           [ 4.6450e+01, -4.3023e+00],
 ...          [-4.8062e+01,  2.3332e+01],
           [-7.2866e+00,  9.8200e-01],
           [-3.9617e+01, -5.3813e+00]]]]])
B = tensor([[[6.2502e-02, 0.0000e+00],
         [5.1306e-02, 2.1395e-18],
         [2.8378e-02, 2.3744e-18],
         ...,...     [8.6822e-03, 1.6500e-18],
         [2.3295e-02, 6.8711e-18],
         [4.2115e-02, 1.2082e-17]]], device='cuda:0')
inplace = False

    def cdgmm(A, B, inplace=False):
        """
            Complex pointwise multiplication between (batched) tensor A and tensor B.
    
            Parameters
            ----------
            A : tensor
                input tensor with size (B, C, M, N, 2)
            B : tensor
                B is a complex tensor of size (M, N, 2)
            inplace : boolean, optional
                if set to True, all the operations are performed inplace
    
            Returns
            -------
            C : tensor
                output tensor of size (B, C, M, N, 2) such that:
                C[b, c, m, n, :] = A[b, c, m, n, :] * B[m, n, :]
        """
        A, B = A.contiguous(), B.contiguous()
        if A.size()[-3:] != B.size():
            raise RuntimeError('The filters are not compatible for multiplication!')
    
        if not iscomplex(A) or not iscomplex(B):
            raise TypeError('The input, filter and output should be complex')
    
        if B.ndimension() != 3:
            raise RuntimeError('The filters must be simply a complex array!')
    
        if type(A) is not type(B):
            raise RuntimeError('A and B should be same type!')
    
    
        C = A.new(A.size())
    
        A_r = A[..., 0].contiguous().view(-1, A.size(-2)*A.size(-3))
        A_i = A[..., 1].contiguous().view(-1, A.size(-2)*A.size(-3))
    
        B_r = B[...,0].contiguous().view(B.size(-2)*B.size(-3)).unsqueeze(0).expand_as(A_i)
        B_i = B[..., 1].contiguous().view(B.size(-2)*B.size(-3)).unsqueeze(0).expand_as(A_r)
    
>       C[..., 0].view(-1, C.size(-2)*C.size(-3))[:] = A_r * B_r - A_i * B_i
E       RuntimeError: expected type torch.FloatTensor but got torch.cuda.FloatTensor

kymatio/scattering2d/backend/backend_torch.py:200: RuntimeError
```

edouardoyallon:
Sorry, this PR isn't passing the test? I can't see the report

janden:
> Sorry, this PR isn't passing the test? I can't see the report

It didn't after you addressed some of @eickenberg's comments. I fixed the bugs, though, so it should be good now.

@eickenberg, are we ready to merge? If so, I'll do a quick rebase.

edouardoyallon:
definintely can be merged!

janden:
Ok. Wasn't asking you, but sure. Merging.

kymatio/scattering2d/backend/backend_torch.py
```diff
@@ -18,22 +18,24 @@ def __init__(self, pad_size, pre_pad=False):
...
-            output.select(4, 0)[:] = out_
+        self.build()
+
+    def build(self):
+        self.padding_module = ReflectionPad2d(self.pad_size)
+
+    def __call__(self, x):
+        if not self.pre_pad:
+            x = self.padding_module(x)
+        output = x.new_zeros(x.shape + (2,))
```
eickenberg: For the record, this could have been fixed with `torch.zeros(x.shape + (2,), dtype=x.dtype, device=x.device)`, but this is fine.

janden: I had that originally, but figured this was cleaner. After all, one of the bugs was caused by `torch.zeros(x.shape + (2,), dtype=x.dtype)`, where we didn't specify `device`. Before the PR, we also had `new_zeros`.

PR #364, opened 2019-03-03
FIX BUG(merge?) J=ceil(log(N)/log(2))

### LINKED ISSUE #363: Scattering2D doesn't work with 2**J == image_size

```
import torch

x = torch.randn(1, 1, 32, 32)
Sx = scattering(x)
print(Sx.size())
```
gives:
```
    Sx = scattering(x)

```
Can you reproduce the error ?

Thanks

edouardoyallon:
Close #363

### DIFF
--- a/kymatio/scattering2d/scattering2d.py
+++ b/kymatio/scattering2d/scattering2d.py
@@ -93,7 +93,7 @@ class Scattering2D(object):
         self.max_order = max_order
         self.shape = shape
         if 2**J>shape[0] or 2**J>shape[1]:
-            raise (RuntimeError('The largest dimension should be larger than 2^J'))
+            raise (RuntimeError('The smallest dimension should be larger than 2^J'))
 
         self.build()
 
--- a/kymatio/scattering2d/scattering2d.py
+++ b/kymatio/scattering2d/scattering2d.py
@@ -102,7 +102,7 @@ class Scattering2D(object):
         self.modulus = Modulus()
         self.M_padded, self.N_padded = compute_padding(self.M, self.N, self.J)
         # pads equally on a given side if the amount of padding to add is an even number of pixels, otherwise it adds an extra pixel
-        self.pad = Pad([(self.N_padded - self.N) // 2, (self.N_padded - self.N+1) // 2, (self.M_padded - self.M) // 2, (self.M_padded - self.M + 1) // 2], pre_pad=self.pre_pad)
+        self.pad = Pad([(self.N_padded - self.N) // 2, (self.N_padded - self.N+1) // 2, (self.M_padded - self.M) // 2, (self.M_padded - self.M + 1) // 2], [self.N, self.M], pre_pad=self.pre_pad)
         self.subsample_fourier = SubsampleFourier()
         # Create the filters
         filters = filter_bank(self.M_padded, self.N_padded, self.J, self.L)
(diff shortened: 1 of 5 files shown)


kymatio/scattering2d/scattering2d.py
```diff
@@ -92,6 +92,8 @@ def __init__(self, J, shape, L=8, max_order=2, pre_pad=False):
         self.pre_pad = pre_pad
         self.max_order = max_order
         self.shape = shape
+        if 2**J>shape[0] or 2**J>shape[1]:
+            raise (RuntimeError('The largest dimension should be smaller than 2^J'))
```
Jonas1312: should be greater than

edouardoyallon: you mean like "2**J>=shape[0] or 2**J>=shape[1]" ?! Isn't it precisely the setting you're trying to use(and that was available with prior scatnet software)?

edouardoyallon: my bad

Jonas1312:
Right now:
```
import torch

x = torch.randn(1, 1, 32, 32)
Sx = scattering(x)
print(Sx.size())
a = Sx.view(-1)
print(a.size())
```

gives the following output:
```
torch.Size([1, 1, 681, 0, 0])
torch.Size([])
```

I think the proper output size should be:
```
torch.Size([1, 1, 681, 1, 1])
```

edouardoyallon:
@Jonas1312 thanks a lot, I would have missed it.. this is due to the unpad function. So here, I see 3 options:

1. Maintaining a padding by 3*N=3*2^J as initially done.
2. Modifying the unpad function such that it handles non-symmetric padding. The issue that comes to my mind is that it will lead to an assymetry which is not really desirable (it was actually a bug previously introduced in scatnet-2D ..)
3. Editing special routines for the the setting where N=2^J. Indeed, in this setting, the aliasing will be potentially strong and that could be an issue... For instance, using a stronger oversampling and a simple sum as a low-pass filtering could be an improvement.

@janden @lostanlen @eickenberg  any suggestions? This should strategy could be common to 1-2-3-D signals..

eickenberg:
So while looking at this, I went to verify that we have the option of doing the padding ourselves. We do seem to have that, but in a weird way in which the fact of "not padding" is given to the padding object which then doesn't pad :)

https://github.com/kymatio/kymatio/blob/master/kymatio/scattering2d/backend/backend_torch.py#L41

While we're at it, the casting to complex performed in the padding object seems completely out of scope for that function. Also `unpad` seems surprisingly simple as a function - I'm surprised unpadding is always achieved by just removing a pixel everywhere, but if it works, cool!

I have no clear opinion about this issue as of now, because I don't really know all the things that had been done up to now, but I do think that

- signal sizes of size `2**J` will probably happen a lot, especially for researchers and people from signal processing that want to control exactly what is going on, so this case should be treated with care
- cases that are not of size `2**J` shouldn't be treated radically differently without reason/explanation from the former. If suddenly the padding doubles, it's weird and completely unpredictable for a user

edouardoyallon:
okay, for now I'll go to solution 1. I'll try to reproduce the @joanbruna 's experiments with textures

lostanlen:
i'm a bit late to the party, but why does this have to be a special case? 
can't we have `J=Inf` for expected scattering and `J` finite for everything that requires spatial localization?

edouardoyallon:
@lostanlen the main reason is that pytorch knows how to pad a signal of size N up to... 2N-1, but it launches an error on 2N ... it's not a difficult fix but it has to be done carefully.. this is why it's still not done yet.

edouardoyallon:
hi @eickenberg is it mergeable? if yes, please merge.

eickenberg:
if travis passes, we can merge this

PR #368, opened 2019-03-19
fix 2D example: plot_filters.py

grlee77:
This PR fixes one of the 2D examples that became outdated after PRs #345 and #320 

I can squash to one commit if you prefer, but left it separated for now in case you don't want the stylistic changes in the third commit.

I added `plt.show()` so that the plots will stay around for users running the example from the command line rather than an interactive session. Otherwise the figures pop up and immediately disappear when the script reaches the end.

### DIFF
--- a/examples/2d/plot_filters.py
+++ b/examples/2d/plot_filters.py
@@ -47,9 +47,7 @@ plt.rc('text', usetex=True)
 plt.rc('font', family='serif')
 i=0
 for filter in filters_set['psi']:
-    f_r = filter[0][...,0]
-    f_i = filter[0][..., 1]
-    f = f_r + 1j*f_i
+    f = filter[0][...,0]
     filter_c = fft2(f)
     filter_c = np.fft.fftshift(filter_c)
     axs[i // L, i % L].imshow(colorize(filter_c))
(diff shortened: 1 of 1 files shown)


janden:
Hi @grlee77. Thank you for bringing this to our attention. As you can see, we're not consistently testing our examples, which we should be doing. I'll take a closer look at this tomorrow and will hopefully merge it by the end of the week.

examples/2d/plot_filters.py
```diff
@@ -71,13 +73,14 @@ def colorize(z):
...
 filter_c = np.fft.fftshift(filter_c)
-plt.suptitle("The corresponding low-pass filter, also known as scaling function."
-"Color saturation and color hue respectively denote complex magnitude and complex phase", fontsize=13)
+plt.suptitle(("The corresponding low-pass filter, also known as scaling "
+              "function.\nColor saturation and color hue respectively denote "
+              "complex magnitude and complex phase"), fontsize=13)
 filter_c = np.abs(filter_c)
 plt.imshow(filter_c)
+
+plt.show()
```
janden: @edouardoyallon Why wasn't this here in the first place?

edouardoyallon: > I added plt.show() so that the plots will stay around for users running the example from the command line rather than an interactive session. Otherwise the figures pop up and immediately disappear when the script reaches the end.

Because I didn't expect an user to run this script in a shell..

edouardoyallon:
Thanks again!

PR #370, opened 2019-03-23
FIX (merge?) Updates for new jenkins server

dylex:
Along with some other cleanups and efficiency improvements to test framework.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -10,19 +10,18 @@ pipeline {
       agent {
 	dockerfile {
 	  dir 'tools'
-	  args '--device /dev/nvidia0:/dev/nvidia0 --device /dev/nvidiactl:/dev/nvidiactl --device /dev/nvidia-uvm:/dev/nvidia-uvm'
 	}
       }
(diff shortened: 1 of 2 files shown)


edouardoyallon:
I think it'd be awesome if we could have back this continuous-integration software!

dylex:
Thanks for the reminder.  I think I've gotten this working again the nvidia way -- nvidia-container-runtime and nvidia/cuda:9.2 docker image -- which is simpler anyway.  (The current python packages don't work with cuda 10.)  Should be safe to merge/rebase at this point, @janden.

eickenberg:
This LGTM.

Wondering how is best to merge given that some of the commits fail the Jenkins tests.

Should we 
a) squash them to remove the failed commits (so that in the future the only failed commits are in WIP pull requests)

or

b) rebase and keep the commit history because it is informative/useful in some way?

dylex:
Squashing is fine -- I've rebased now to clean up history.

edouardoyallon:
Thanks so much!

Issue #373, opened 2019-03-27
Hello, I have a question about inverse


strawberry1996:
I reading your code about scatter inverse, and I train the model use your way, but I don't know, how can I test the model?  for instance, when I scattering, I set L =8, J=2 then I get 17 feature images, What should I input when I invert? Is it 17 scattering images?

edouardoyallon:
I'll add soon a PR with an example of scattering inversion via gradient descent.

Jonas1312:
Not sure if it's what you're looking for but maybe this can help you: [https://github.com/Jonas1312/scattering-networks/blob/master/Python/image_reconstruction/main.py](https://github.com/Jonas1312/scattering-networks/blob/master/Python/image_reconstruction/main.py)

edouardoyallon:
Hi, thanks a lot! Maybe you could do a PR?

PR #380, opened 2019-05-12
ENH+WIP Code for inverting scattering via mse

### LINKED ISSUE #374: Code for inverting scattering via mse
Should be added in examples..

Close https://github.com/kymatio/kymatio/issues/373

Jonas1312:
Close #374 and #373 

TODO:
- [x] Add plots
- [x] Add comments
- [x] Add image
- [x] Replace cv2 with PIL
- [x] Tune hyperparams


Jonas1312:
@edouardoyallon Is it fine if I add an image in `kymatio\examples\2d\images\` or should I put it somewhere else?

edouardoyallon:
> @edouardoyallon Is it fine if I add an image in `kymatio\examples\2d\images\` or should I put it somewhere else?

@eickenberg @janden @lostanlen any suggestions of a folder in which we should put some extra material? I guess also it should have no copyright.

Jonas1312:
Are those images copyrights free? (from "Compressing the Input for CNNs with the first order scattering transform" paper)
Capture

edouardoyallon:
> Are those images copyrights free? (from "Compressing the Input for CNNs with the first order scattering transform" paper)

AFAIR, it is, yes, but maybe it is better here to use a more standard set of images from computer vision softwares: https://www.hlevkin.com/06testimages.htm for instance, 

Capture

That being said, I can also dig and provide the original images from this paper.

eickenberg:
> > @edouardoyallon Is it fine if I add an image in `kymatio\examples\2d\images\` or should I put it somewhere else?
> 
> @eickenberg @janden @lostanlen any suggestions of a folder in which we should put some extra material? I guess also it should have no copyright.

Yeah, I think `examples/2d/images` is OK. I'd be happy with that solution.

Alternatively, we could do something like `skimage` does (`from skimage.data import chelsea` gives you a function that loads this `https://github.com/scikit-image/scikit-image/blob/master/skimage/data/chelsea.png`). And these functions could be stored in `kymatio.datasets.images`, requiring  the creation of a folder `datasets`, to which `datasets.py` needs to be moved, some functions of which need to be imported in the `datasets/__init__.py` to preserve current functionality, and then creating `images.py` which should contain functions for loading specific images.

examples/2d/invert_scattering.py
```diff
@@ -0,0 +1,80 @@
...
+from kymatio import Scattering2D
+
+device = "cuda" if torch.cuda.is_available() else "cpu"
+
+# Load src image
+img_name = "./images/lena_64.png"
+src_img = Image.open(img_name).convert("RGB")
+src_img = np.array(src_img).astype(np.float32)
+src_img = src_img / 255.0
+y_true = np.array(src_img)  # copy of src_img
```
eickenberg: `src_img.copy()` but unclear to me why you do this? (You already created two copies by the previous two operations - in case you were afraid of anything modifying the png file :))

examples/2d/invert_scattering.py
```diff
@@ -0,0 +1,80 @@
...
+        best_loss = float("inf")
+        for epoch in range(1, 300):
+            new_coefficients = scattering(input_tensor)
+            loss = F.mse_loss(input=new_coefficients, target=scattering_coefficients)
+            print("Epoch {}, loss: {}".format(epoch, loss.item()), end="\r")
+            optimizer.zero_grad()
+            loss.backward()
+            optimizer.step()
+            if loss < best_loss:
+                best_loss = loss
```
eickenberg: I would also detach this loss (this is not as bad as keeping all losses in a list, but I think it might still retain a graph in certain cases)

So do `best_loss = loss.detach().cpu().numpy()` or something

examples/2d/invert_scattering.py
```diff
@@ -0,0 +1,80 @@
...
+                best_img = input_tensor.clone().detach()
```
eickenberg: Is the clone necessary when you detach? If you want to store this image for looking at later, why not just go `best_img = input_tensor.detach().cpu().numpy()`

examples/2d/invert_scattering.py
```diff
@@ -0,0 +1,80 @@
...
+
+        y_pred = best_img.cpu().detach().numpy()
```
eickenberg: If you do the abovementioned thing, then this here can be shortened/removed

examples/2d/invert_scattering.py
```diff
@@ -0,0 +1,80 @@
...
+        y_pred = np.moveaxis(y_pred, 0, -1)  # CHW to HWC
+
+        y_pred = np.clip(y_pred, 0.0, 1.0)
```
eickenberg: Is this the prefered way for reconstruction? (Could also imagine rescaling, but it's true that this could be very uinstable wrt weird high values)

examples/2d/invert_scattering.py
```diff
@@ -0,0 +1,80 @@
...
+
+        # PSNR
+        print("")
```
eickenberg: hmm, is this for a newline or something? Maybe just add `\n` to the beginning of the next print statement?

examples/2d/invert_scattering.py
```diff
@@ -0,0 +1,80 @@
...
+        mse = np.mean((y_true - y_pred) ** 2)
+        psnr = 20 * np.log10(1.0 / np.sqrt(mse))
```
eickenberg: Here it is assumed that signal size is `1.0`? Wouldn't that then be in `\infty` norm, whereas the noise is measured in L2 norm?

I never know how to do these estimates so if you say this is the way to do it, fine, otherwise maybe others can chime in here?

examples/2d/invert_scattering.py
Jonas1312: Changed to `best_loss = loss.detach().cpu().item()`

Jonas1312:
Preview with https://www.hlevkin.com/TestImages/baboon.bmp:

Figure_1
Figure_2
Figure_3

I will add the image in the next commit if it suits you

examples/2d/invert_scattering.py
Jonas1312: Yes rescaling doesn't work properly with outliers, any suggestions?

examples/2d/invert_scattering.py
edouardoyallon: that's fine here

examples/2d/invert_scattering.py
edouardoyallon: I also used this method and it's not clear to me what's the best, yet at least, this is working.

edouardoyallon:
I've reduced the number of gradient step for CPU use and it builds correctly with the documentation. @Jonas1312, let me know if one can merge.

Jonas1312:
Thanks @edouardoyallon ! LGTM

PR #381, opened 2019-05-13
FIX(merge?) different devices message error for 2D scattering

### LINKED ISSUE #357: `cdgmm` doesn't check that both inputs are on the same device
This is at least true for the 2D torch `cdgmm`. It only complains when you try to multiply tensors. We should throw an error instead.

### LINKED ISSUE #363: Scattering2D doesn't work with 2**J == image_size

edouardoyallon:
Closes #357  

This PR ensures that 2 cuda tensors are on the same GPU.

### DIFF
--- a/kymatio/scattering2d/backend/backend_skcuda.py
+++ b/kymatio/scattering2d/backend/backend_skcuda.py
@@ -302,8 +302,12 @@ def cdgmm(A, B, inplace=False):
     if A.dtype is not B.dtype:
         raise RuntimeError('A and B must be of the same dtype')
 
-    if A.device != B.device:
-        raise RuntimeError('A and B must be on the same device')
+    if A.device.type != B.device.type:
+        raise RuntimeError('A and B must be of the same device type')
+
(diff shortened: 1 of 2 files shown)


eickenberg:
Huh - can you explain a bit?

Does this mean that `A.device == B.device` does not necessarily imply that they are the same GPU?

(Trying to understand how this change fixes something -- I definitely appreciate the more fine-grained error messages though)

edouardoyallon:
> Huh - can you explain a bit?
> 
> Does this mean that `A.device == B.device` does not necessarily imply that they are the same GPU?
> 
> (Trying to understand how this change fixes something -- I definitely appreciate the more fine-grained error messages though)

I'm not sure what was the issue yet it solves Joakim concerns.. device is an object, not a string, it might be the origin of this bug..

PR #384, opened 2019-05-13
ENH(merge?) support of J=1 for 2D

### LINKED ISSUE #379: Unexpected crush with J=1
Hi, 

I faced the following problem: I get unexpected and unexplained crush when J=1 regardless of other parameters values. Please, find the code and the error message below. 

Here's the code:

```
data = torch.Tensor(np.random.rand(128,3,32,32))
print("shape before scattering: {}".format(data.shape))

scatModel = Scattering2D(J=1,L=8, shape = (32,32), max_order=2)
a = scatModel(data)

print("shape after scattering: {}".format(a.shape))
```

And the error message: 

> shape before scattering: torch.Size([128, 3, 32, 32])
> Traceback (most recent call last):
>   File "/user/HS221/dm00314/PycharmProjects/ScatterNetsTestFramework/venv/mixingMallatTest.py", line 24, in <module>
>     a = scatModel(data)
>   File "/user/HS221/dm00314/PycharmProjects/ScatterNetsTestFramework/venv/lib/python3.5/site-packages/kymatio-0.2.0.dev0-py3.5.egg/kymatio/scattering2d/scattering2d.py", line 235, in __call__
>   File "/user/HS221/dm00314/PycharmProjects/ScatterNetsTestFramework/venv/lib/python3.5/site-packages/kymatio-0.2.0.dev0-py3.5.egg/kymatio/scattering2d/scattering2d.py", line 202, in forward
> KeyError: 0

Is it me or something is wrong?

Thank you, 
Dmitry

edouardoyallon:
Closes #379

### DIFF
--- a/kymatio/scattering2d/filter_bank.py
+++ b/kymatio/scattering2d/filter_bank.py
@@ -46,14 +46,24 @@ def filter_bank(M, N, J, L=8):
             psi_signal_fourier = fft2(psi_signal)
             # drop the imaginary part, it is zero anyway
             psi_signal_fourier = np.real(psi_signal_fourier)
-            for res in range(min(j + 1, J - 1)):
+            if J > 1:
+                for res in range(min(j + 1, J - 1)):
+                    psi_signal_fourier_res = periodize_filter_fft(
+                        psi_signal_fourier, res)
+                    # add a trailing singleton dimension to mark it as non-complex
+                    psi_signal_fourier_res = psi_signal_fourier_res[..., np.newaxis]
+                    psi[res] = psi_signal_fourier_res
+                    # Normalization to avoid doing it with the FFT.
+                    psi[res] /= M*N// 2**(2*j)
+            else:
                 psi_signal_fourier_res = periodize_filter_fft(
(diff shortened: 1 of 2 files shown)


kymatio/scattering2d/filter_bank.py
```diff
@@ -46,14 +46,24 @@ def filter_bank(M, N, J, L=8):
             psi_signal_fourier = fft2(psi_signal)
             # drop the imaginary part, it is zero anyway
             psi_signal_fourier = np.real(psi_signal_fourier)
-            for res in range(min(j + 1, J - 1)):
+            if J > 1:
+                for res in range(min(j + 1, J - 1)):
```
eickenberg: OK so this block looks like a copy/paste of block below.

Is it not possible to find a way of parametrizing the loop such that the J=1 case is contained within it the way we want?

If not, then I'm OK with duplicating code (I'd like to avoid it if it is easily avoided though)

edouardoyallon:
The coverage isn't accurate but jenkin works. I'm merging.

PR #385, opened 2019-05-21
[DO NOT MERGE]  ENH(WIP) Add 1d wavelets: Battle-Lemarie and Meyer

rleonarduzzi:
This pull request adds (cubic) Battle-Lemarie and Meyer wavelets in 1d.

Changes in the example `plot_filters.py` show the new filters.

I've tried to minimize the intrusiveness of the changes as much as possible, trying to preserve structure as much as possible. This leads to a clear inefficiency in one place: the maximum dyadic subsamplings are computed numerically by precomputing the filters. These extra computations could be avoided in the future by restructuring the factory function, but I've kept it simple for now.

These filters work only for Q=1. I'm not sure about how to extend them for Q>1, but they are still useful in cases where dyadic wavelets are enough.

### DIFF
--- a/examples/1d/plot_filters.py
+++ b/examples/1d/plot_filters.py
@@ -25,6 +25,12 @@ import matplotlib.pyplot as plt
 ###############################################################################
 # Filter parameters and generation
 # --------------------------------
+# Kymatio implements different wavelets: 'morlet', 'battle-lemarie' and 'meyer'.
+# We start by choosing morlet, and we will compare it with other filter banks
+# later.
+wavelet_type = 'morlet'
+
+
 # The filters are defined for a certain support size `T` which corresponds to
 # the size of the input signal. The only restriction is that `T` must be a
 # power of two. Since we are not computing any scattering transforms here, we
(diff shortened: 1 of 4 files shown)


edouardoyallon:
@rleonarduzzi where are we? The scattering community is watching you! 😄

lostanlen:
> These filters work only for Q=1. I'm not sure about how to extend them for Q>1, but they are still useful in cases where dyadic wavelets are enough.

Yes, you're totally right. It's not completely unthinkable, but it would probably be quite cumbersome, and i don't think that there is a reference textbook or article to do it. And writing an implementation from an unpublished idea is probably a bad idea.
For Q>1, we can redirect the user to the Morlet wavelet.
I will also port my implementation of adjustable-Q Gammatone wavelet to Kymatio at some point.
(https://static-content.springer.com/esm/art%3A10.1186%2Fs13636-018-0138-4/MediaObjects/13636_2018_138_MOESM1_ESM.pdf)

edouardoyallon:
This is a difficult feature to add, but it would be great to do it...

janden:
> This is a difficult feature to add, but it would be great to do it...

Agreed. It would also be a useful exercise to find out what a good generalization of the filter bank looks like.

Regarding Q = 1 vs. Q > 1, I don't think we should (or need to) extend these for the Q > 1 case. There's plenty of applications in Q = 1 that would benefit from these wavelets.

janden:
Is this something we want to look at for the sprint? May need to restart the PR given the number of changes made in the meantime. All the same, the code in here should be a useful starting point.

janden:
Is there any interest in merging these anytime soon?

lostanlen:
i could have it either way. i don't need them in the near future

janden:
Ok. Closing.

PR #388, opened 2019-05-29
ENH adding to() and _apply to the 1D code

edouardoyallon:
This PR does the same modifications from the 2D code ( #384 ).

### DIFF
--- a/kymatio/scattering1d/scattering1d.py
+++ b/kymatio/scattering1d/scattering1d.py
@@ -10,7 +10,7 @@ from .backend import (fft1d_c2c, ifft1d_c2c, modulus_complex, pad, real,
     subsample_fourier, unpad)
 from .filter_bank import (calibrate_scattering_filters,
     scattering_filter_factory)
-from .utils import cast_phi, cast_psi, compute_border_indices, compute_padding
+from .utils import _apply_phi, _apply_psi, compute_border_indices, compute_padding
 
 
 __all__ = ['Scattering1D']
@@ -237,9 +237,9 @@ class Scattering1D(object):
         """
             Mimicates the behavior of the function _apply() of a nn.Module()
         """
-        _apply_psi(self.psi1_f, target_type)
-        _apply_psi(self.psi2_f, target_type)
-        _apply_phi(self.phi_f, target_type)
+        _apply_psi(self.psi1_f, fn)
+        _apply_psi(self.psi2_f, fn)
+        _apply_phi(self.phi_f, fn)
         return self
 
     def cuda(self, device=None):
--- a/kymatio/scattering1d/scattering1d.py
+++ b/kymatio/scattering1d/scattering1d.py
@@ -233,32 +233,24 @@ class Scattering1D(object):
         self.phi_f = phi_f
         self._type(torch.FloatTensor)
 
-    def _type(self, target_type):
-        """Change the datatype of the filters
-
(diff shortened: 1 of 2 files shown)


kymatio/scattering1d/scattering1d.py
```diff
@@ -233,36 +233,42 @@ def build(self):
...
-        ----------
-        target_type : type
-            The desired type of the filters, typically `torch.FloatTensor`
-            or `torch.cuda.FloatTensor`.
+    def _apply(self, fn):
         """
-        cast_psi(self.psi1_f, target_type)
-        cast_psi(self.psi2_f, target_type)
-        cast_phi(self.phi_f, target_type)
+            Mimicates the behavior of the function _apply() of a nn.Module()
```
eickenberg: *Mimics

Can't you just register `self.psi` etc as parameters and it would do it automatically?

I guess you would need to inherit from `nn.Module` and I guess we are trying to avoid that?

kymatio/scattering1d/scattering1d.py
```diff
@@ -233,36 +233,42 @@ def build(self):
...
+        _apply_psi(self.psi1_f, fn)
+        _apply_psi(self.psi2_f, fn)
+        _apply_phi(self.phi_f, fn)
         return self
 
-    def cpu(self):
-        """Move to the CPU
+    def cuda(self, device=None):
+        """
+            Mimicates the behavior of the function cuda() of a nn.Module()
```
eickenberg: *Mimics

kymatio/scattering1d/scattering1d.py
```diff
@@ -233,36 +233,42 @@ def build(self):
...
+        """
+        return self._apply(lambda t: t.cuda(device))
 
-        This function prepares the object to accept input Tensors on the CPU.
+    def to(self, *args, **kwargs):
+        """
+            Mimicates the behavior of the function to() of a nn.Module()
```
eickenberg: *Mimics

kymatio/scattering1d/scattering1d.py
```diff
@@ -233,36 +233,42 @@ def build(self):
...
         """
-        return self._type(torch.FloatTensor)
+        device, dtype, non_blocking = torch._C._nn._parse_to(*args, **kwargs)
```
eickenberg: OK, so here we are using torch. I guess we have to talk about how to separate this away to have different backends/frontends?

kymatio/scattering1d/scattering1d.py
```diff
@@ -233,36 +233,42 @@ def build(self):
...
-        """Move to the GPU
+        def convert(t):
+            return t.to(device, dtype if t.is_floating_point() else None, non_blocking)
 
-        This function prepares the object to accept input Tensors on the GPU.
+        return self._apply(convert)
+
+    def cpu(self):
+        """
+            Mimicates the behavior of the function cpu() of a nn.Module()
```
eickenberg: Mimics

kymatio/scattering1d/scattering1d.py
edouardoyallon: Absolutely correct. This is for a next PR.

kymatio/scattering1d/scattering1d.py
edouardoyallon: Yes, it will happen in some future commits!!

Issue #339, opened 2019-02-16
Test for raised error and warnings in 3D torch backend


janden:
Specifically, [`fft` may raise a `TypeError` for non-complex input](https://codecov.io/gh/kymatio/kymatio/src/6e5c430d71a1562a93931eee0f568a9aba489220/kymatio/scattering3d/backend/backend_torch.py#L48) and [`cdgmm3d` may issue a warning for non-contiguous arrays](https://codecov.io/gh/kymatio/kymatio/src/6e5c430d71a1562a93931eee0f568a9aba489220/kymatio/scattering3d/backend/backend_torch.py#L68). These need to be tested for.

Issue #337, opened 2019-02-16
Test for raised errors in `Scattering3D`


janden:
In particular, `ValueError`s, `TypeErrors`s, and `RuntimeError`s may be raised by `_compute_scattering_coefs` and `_convolution_and_modulus`, but these are not tested for. For example, see [here](https://codecov.io/gh/kymatio/kymatio/src/6e5c430d71a1562a93931eee0f568a9aba489220/kymatio/scattering3d/scattering3d.py#L214).

MuawizChaudhary:
@janden @edouardoyallon Is this still a problem?

PR #390, opened 2019-06-12
ENH(merge?) adding to() and _apply to the 3D code 

### LINKED ISSUE #348: Remove PyTorch from 1D/3D filter generation
- [x] 1D
- [x] 3D

Reproduce #345 in 1D and 3D.

### LINKED ISSUE #351: `to` function
TODO (edited by EO):

- [x] method `.to()` for 1D
- [x] method `.to()` for 2D
- [x] method `.to()` for 3D

PyTorch tensors have a `to` function (see https://pytorch.org/docs/stable/tensors.html#torch.Tensor.to), which combines datatype conversion and device copy (i.e., `cpu()` and `cuda()` functions). It might be good to have something similar on our end. This would, for example, simplify device-agnostic code that could simply call
```
scattering.to(x.device)
Sx = scattering(x)
```
instead of
```
if x.is_cuda:
    scattering.cuda()
else:
    scattering.cpu()
Sx = scattering(x)
```
It would also simplify our test suite which is full of such if statements.

A (slightly) related question is how to handle multiple GPUs. Right now, we're restricted to the current GPU as defined by PyTorch.

Finally, all of this needs to be considered with our goal of backend agnosticism in mind. We don't want to commit to hard to the PyTorch infrastructure. Still, I think there could be a way to do this in an appropriate manner.

edouardoyallon:
This PR does the same modifications from the 2D code ( #384 ), and also remove torch from the filter bank building procedure

Close #351
Close #348

### DIFF
--- a/kymatio/scattering3d/scattering3d.py
+++ b/kymatio/scattering3d/scattering3d.py
@@ -79,19 +79,42 @@ class HarmonicScattering3D(object):
         self.gaussian_filters = gaussian_filter_bank(
                                 self.M, self.N, self.O, self.J + 1, self.sigma_0)
 
-    def cuda(self):
-        """Move to the GPU
-
-        This function prepares the object to accept input Tensors on the GPU.
+    def _apply(self, fn):
         """
-        self.is_cuda = True
+            Mimics the behavior of the function _apply() of a nn.Module()
+        """
+        _apply_psi(self.psi1_f, fn)
+        _apply_psi(self.psi2_f, fn)
+        _apply_phi(self.phi_f, fn)
+        return self
+
+    def cuda(self, device=None):
+        """
+            Mimics the behavior of the function cuda() of a nn.Module()
+        """
+        return self._apply(lambda t: t.cuda(device))
+
+    def to(self, *args, **kwargs):
+        """
+            Mimics the behavior of the function to() of a nn.Module()
+        """
+        device, dtype, non_blocking = torch._C._nn._parse_to(*args, **kwargs)
+
+        if dtype is not None:
+            if not dtype.is_floating_point:
+                raise TypeError('nn.Module.to only accepts floating point '
+                                'dtypes, but got desired dtype={}'.format(dtype))
+
+        def convert(t):
+            return t.to(device, dtype if t.is_floating_point() else None, non_blocking)
+
+        return self._apply(convert)
 
     def cpu(self):
-        """Move to the CPU
-
-        This function prepares the object to accept input Tensors on the CPU.
         """
-        self.is_cuda = False
+            Mimics the behavior of the function cpu() of a nn.Module()
+        """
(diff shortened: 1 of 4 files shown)


edouardoyallon:
I removed some tests that do not make any sens given the fact that the variable is_cuda got removed.

I think more test should be added by @louity via #337 #339

edouardoyallon:
I removed a test which is outdated given the current backend.

kymatio/scattering3d/scattering3d.py
```diff
@@ -79,19 +77,46 @@ def build(self):
         self.gaussian_filters = gaussian_filter_bank(
                                 self.M, self.N, self.O, self.J + 1, self.sigma_0)
 
-    def cuda(self):
-        """Move to the GPU
+        # transfer the filters from numpy to torch
+        for k in range(len(self.filters)):
+            self.filters[k] = torch.from_numpy(self.filters[k]).type(torch.Tensor)
```
eickenberg: why `.type(torch.Tensor)`? What does that do? How can the output of `torch.from_numpy(self.filters[k])` not be a `torch.Tensor` ?

kymatio/scattering3d/scattering3d.py
```diff
@@ -79,19 +77,46 @@ def build(self):
...
+        self.gaussian_filters = torch.from_numpy(self.gaussian_filters).type(torch.Tensor)
```
eickenberg: same as above

kymatio/scattering3d/tests/test_scattering3d.py
```diff
@@ -274,42 +274,6 @@ def test_scattering_methods():
     scattering.rotation_covariant = False
     Sx = scattering(x)
 
-def test_cpu_cuda():
```
eickenberg: Can't you modify this test to work? All you have to do is remove the `assert S.is_cuda`

kymatio/scattering3d/scattering3d.py
edouardoyallon: Ok, so the point is that by default, the type of the returned tensor(here desired float32) corresponds to the type of the numpy array(here np.64). On the other hand, I think it makes sens to use the default type set up in torch(eg torch.Tensor, which is normally float32). It will be helpful the day we will be float64 compatible!

kymatio/scattering3d/scattering3d.py
edouardoyallon: same thing as above

edouardoyallon: same thing as above

kymatio/scattering3d/tests/test_scattering3d.py
edouardoyallon: This test is redundant(since I removed the local variable is_cuda) with most of the other subparts of the code, I don't want to keep it!

PR #396, opened 2019-06-14
ENH(WIP) adding meta to 2D scattering

### LINKED ISSUE #376: Accessing coefficients layer wise
I am trying to compare the energy of the signal at different levels of the scattering response. However, I have trouble pointing out which index value corresponds to which level?
Please refer to the code below taken from MNIST classification:

```python
import torch.nn as nn
import torch.optim
from torchvision import datasets, transforms
import torch.nn.functional as F
import kymatio.datasets as scattering_datasets
import numpy as np 
myDataSet = datasets.MNIST(scattering_datasets.get_dataset_dir('MNIST'),train=True, download=True,transform=transforms.Compose([ transforms.ToTensor()]))
train_loader = torch.utils.data.DataLoader(myDataSet,shuffle=True,batch_size=1)
dataiter = iter(train_loader)
images, labels = dataiter.next()
scattering = Scattering2D(J=2, shape=(28, 28),max_order=2,L=8)
val = scattering.forward(images)
scatterImage = val.detach().numpy()
print(scatterImage.shape)
```
This gives the scatterinf response (1, 1, 81,7,7). However, I am not sure which index corresponds to which level.

Also, I would like to apply SVM to scattering response but am not sure how to handle the data point with dimensions (81,7,7) for each datapoint. Can someone point me in the right direction ?

edouardoyallon:
Close #376 

Adding a routine similar to 1D's meta to 2D.

### DIFF
--- a/kymatio/scattering2d/scattering2d.py
+++ b/kymatio/scattering2d/scattering2d.py
@@ -149,6 +149,96 @@ class Scattering2D(object):
         """
         return self._apply(lambda t: t.cpu())
 
+    def meta(self):
+        """Get metadata on the transform.
+
+        This information specifies the content of each scattering coefficient,
+        which order, which frequencies, which filters were used, and so on.
+
+        Parameters
+        ----------
+        J : int
+            The maximum log-scale of the scattering transform.
+            In other words, the maximum scale is given by `2**J`.
+        L : int
+            The number of rotation applied to the mother wavelet.
+        max_order : int, optional
+            The maximum order of scattering coefficients to compute.
+            Must be either equal to `1` or `2`. Defaults to `2`.
+
+        Returns
+        -------
+        meta : dictionary
+            A dictionary with the following keys:
+
+            - `'order`' : tensor
+                A Tensor of length `C`, the total number of scattering
+                coefficients, specifying the scattering order.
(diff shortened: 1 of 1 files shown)


eickenberg:
The way this should be done is by outputting a dictionary with tuple indices that point to the right places

eickenberg:
Why does the array output need to be torch arrays?

As far as I can tell, these dictionaries don't point to the actual data. So this seems quite awkward to me (I am aware we have something like this in another part of the code base)

edouardoyallon:
> The way this should be done is by outputting a dictionary with tuple indices that point to the right places

Do you have an idea of output you would like to get? I'm not sure to understand, as here meta[index] would have info about the coordinate meta.

janden:
IIRC the reason this is done in 1D is to simplify logical indexing of the output. In other words, you can do something like `S[order==1, ...]` and have it work without barking at you. Now whether it's a good idea or not is a different question. We should maybe default all of these fields to numpy arrays or (gasp) Python lists.

edouardoyallon:
I'm up to follow any clear methodology for this! Any suggestions?

edouardoyallon:
@eickenberg @lostanlen @janden any suggestions?

PR #397, opened 2019-06-15
FIX (try to merge after #415) for fft normalization bug in backend and unittest

### LINKED ISSUE #386: Inverse fft in 2D backend
EDIT by EO:

In 2D backend code, in
`

```
import torch

# -- parameters
M = 8
N = 8
L = 8
J = 4

# -- synthesise data
y = torch.rand(1, 1, M, N)
for i in range(M):
    for j in range(N):
        y[:,:,i,j] = i+j

# -- print
print(y)
print(y_fft_inv_un)
```

nshervt:
Close #386

This PR addresses the normalization bug in backend and modifies the fft unittest to avoid such bugs in future. In addition, in regularized inverse 2D example, tqdm was added for better illustration of enumerating batches


edouardoyallon:
Hi @nshervt, thanks a lot! For now the PR fails the tests. You can verify locally that a PR satisfy the test constraints by running `pytest` in the kymatio main folder.

examples/2d/regularized_inverse_scattering_MNIST.py
```diff
@@ -24,6 +24,7 @@
 import torch.nn as nn
 import torch.optim as optim
 from PIL import Image
```
edouardoyallon: I'm pros incorporating this change and adding "tqdm" to the requirements_optional.txt

kymatio/scattering2d/tests/test_scattering2d.py
```diff
@@ -181,6 +181,17 @@ def test_FFT():
...
+
+    # -- synthesise data
+    y = torch.rand(1, 1, 8, 8)
+
+    y_pad = pad(y)                                          # pad to add complex dimension
+    y_fft = backend.fft(y_pad, 'C2C')                       # fft
+    y_fft_inv = backend.fft(y_fft, 'C2C', inverse=True)     # inverse fft
+    y_fft_inv_un = y_fft_inv[..., 1:-1, 1:-1, 0]            # unpad
+
+    assert torch.allclose(y, y_fft_inv_un)
```
edouardoyallon: great, I'm gonna re-use this function!

kymatio/scattering2d/tests/test_scattering2d.py
```diff
@@ -181,6 +181,17 @@ def test_FFT():
         backend.fft(y)
     assert ('must be contiguous' in record.value.args[0])
 
+    pad = backend.Pad(2 ** 0, pre_pad=False)
+
+    # -- synthesise data
+    y = torch.rand(1, 1, 8, 8)
+
+    y_pad = pad(y)                                          # pad to add complex dimension
```
edouardoyallon: I think avoiding using pad to add a the last coordinate would be great and that right now it's mroe confusing than helping.(there is no need to pad for applying a fft)

kymatio/scattering2d/tests/test_scattering2d.py
```diff
@@ -181,6 +181,17 @@ def test_FFT():
...
+    y_fft = backend.fft(y_pad, 'C2C')                       # fft
+    y_fft_inv = backend.fft(y_fft, 'C2C', inverse=True)     # inverse fft
+    y_fft_inv_un = y_fft_inv[..., 1:-1, 1:-1, 0]            # unpad
```
edouardoyallon: Again, remove. Simply consider a tensor with last dimension 2.

kymatio/scattering2d/tests/test_scattering2d.py
nshervt: Agreed!

nshervt:
Hi @edouardoyallon. Thanks for the reviews. In test_Scattering2D, which currently fails, the code is loading data file `test_data_2d.pt` containing a tensor and its scattering output, then applies scattering transform on the original data and expects the result to match the loaded scattering output. But I assume that the scattering output was computed using this code when fft was not being normalized correctly, as a result, fixing normalization would make a different scattering output and I believe that this is why the test fails. Based on this, I think to fix the test fail we need to create a new data file `test_data_2d.pt`.

janden:
> tqdm was added for better illustration of enumerating batches
Not sure about this. We used to have `tqdm` as a dependency but removed it since it was only used in a few places. Is it worth putting it back in?

nshervt:
I think if you download a code for the first time and try to run it, it would be inconvenient without seeing a response from the code and knowing how long it takes to run. But also as you said, simplicity and less requirements for the package would be nice as well. Might be a good idea to make it optional and have a recommended dependency list and a minimalist dependency list as you prefer.

eickenberg:
We went to some lengths to remove `tqdm` so it would be great not to add it back.
That said, having some form of progress report is good, which can be achieved by `print(status_string, end='\r')` periodically. Eventually that might also crystallize into a utility function that we can keep in `kymatio.utils`

edouardoyallon:
@eickenberg  's suggestion sounds good to me

>  But I assume that the scattering output was computed using this code when fft was not being normalized correctly,

Absolutely, that's why I suggested to change the normalization of the filters here: https://github.com/kymatio/kymatio/blob/master/kymatio/scattering2d/filter_bank.py#L56
That should fix this issue.

> Based on this, I think to fix the test fail we need to create a new data file test_data_2d.pt.

No!

edouardoyallon:
Hi @nshervt - Thanks again for your PR, we really do appreciate.

I proposed a fix to the comparison with scatnet coefficients here:
https://github.com/nshervt/kymatio/pull/3
(it's mainly for legacy purpose that I do wish to leave the normalization as it is done in scatnet, I agree that one might be willing to find another solution later)

In order to finally submit your PR, I propose the following: creating two different PRs, one **only** with the fix for the FFT, with the corresponding unit test. A second PR with the add-on that you propose. Currently, the code is a bit hard to review; and combined with the summer break, this is why we didn't merge it yet. In general, I think it's better to open several small PRs when the purposes are quite different rather than having a big patch.

Thanks again, I'm looking forward your PRs! (I can also do them if you wish - just lmk - but I think we're just one step for the merge and I really did like the use of new functions you introduced)

nshervt:
@edouardoyallon Thanks for the fix, will do in the next few days.

edouardoyallon:
> @edouardoyallon Thanks for the fix, will do in the next few days.

Thanks! we're slightly changing the core of 2D-scattering, just ignore that and work with it as if it was the current version, I'll merge your fix in the new mechanisms!

nshervt:
Hi @edouardoyallon, I merged your edits and removed other suggestions to include in another PR. unittests run fine for `test_FFT` and `test_Scattering2D`, but not for `TestCDGMM`. Here is the error that I get for that:

```
======================================================================
ERROR: test_scattering2d.TestCDGMM.test_cdgmm_exceptions
----------------------------------------------------------------------
  File "/Users/Nsherv/anaconda3/lib/python3.7/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
TypeError: test_cdgmm_exceptions() missing 1 required positional argument: 'backend'

======================================================================
ERROR: test_scattering2d.TestCDGMM.test_cdgmm_forward
----------------------------------------------------------------------
  File "/Users/Nsherv/anaconda3/lib/python3.7/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
TypeError: test_cdgmm_forward() missing 4 required positional arguments: 'data', 'backend', 'device', and 'inplace'
```

It doesn't seem to have anything to do with my changes. I have to look into this.

edouardoyallon:
Hi thanks a lot, I'll handle the remaining issues and wille merge this with #415 

Thanks!

PR #398, opened 2019-06-17
ENH (merge?) build a front-end for the scattering for 2D

edouardoyallon:
TODO (non-exaustive):
- [x] front end should ensure that the appropriate backend is loaded
- [x] checking folders and the hiearchy of files
- [x] incorporating the backend as a namedtuple
- [x] adding the scattering function in `core` folder
- [x] frontend/backend dynamic are handled in an userfriendly manner

The goal of this PR is to delimit the 2D scattering computations(the 2 for loops across frequencies) into a simple function and to incorporate sub-routines (fft, cdgmm, ..) in a module.


edouardoyallon:
@eickenberg it passes the test, the doc needs to be improved, I'm about to do it but feel free to propose some changes. Thanks

kymatio/__init__.py
```diff
@@ -13,7 +16,14 @@
             ]
 
 from .scattering1d.scattering1d import Scattering1D
-from .scattering2d.scattering2d import Scattering2D
+
+# if torch is loaded
+if 'torch' in sys.modules:
```
eickenberg: Can you explain what you intend here?

This looks to me like the user has to import torch? And then this gets detected by kymatio?

If this is the case, then I don't think it's a good idea.

kymatio/frontend/torch_f.py
```diff
@@ -0,0 +1,23 @@
+import torch.nn as nn
+
+class Scattering(nn.Module):
+   def __init__(self, J, shape, max_order = None):
+       super(Scattering, self).__init__()
+       return
```
eickenberg: Why this statement?

Also please explain the naming of the file

kymatio/frontend/torch_f.py
```diff
@@ -0,0 +1,23 @@
...
+       self.J = J
+       self.shape = shape
+       self.max_order = max_order
+       self.build()
+
+   def build(self):
+       return
```
eickenberg: remove - this will not error even in the current state

kymatio/frontend/torch_f.py
```diff
@@ -0,0 +1,23 @@
...
+       self.filters = self.create_and_register_filters()
+
+   def create_and_register_filters(self):
+       pass
+
+   def scattering(self):
+       pass
+
+   def forward(self, x):
+      return self.scattering(x)
```
eickenberg: add newline

kymatio/scattering2d/backend/__init__.py
```diff
@@ -61,9 +61,9 @@
 
 
 if BACKEND == 'torch':
-    from .backend_torch import *
+    from .torch_b import *
```
eickenberg: why this naming change?

kymatio/scattering2d/scattering2d.py
```diff
@@ -2,255 +2,69 @@
...
-        scattering_shape = S.shape[-3:]
-        S = S.reshape(batch_shape + scattering_shape)
-
-        return S
-
-    def __call__(self, input):
-        return self.forward(input)
+__all__ = ['scattering2d']
+
+def scattering2d(input, J, L, subsample_fourier, pad, modulus, fft, cdgmm, unpad, phi, psi, max_order, M_padded, N_padded):
```
eickenberg: hmm. Do you think this is the right abstraction?

I am a bit worried about requiring `subsample_fourier` given that scattering isn't inherently bound to being computed in Fourier space.

If possible, I'd prefer `convolution` as a primitive, or even `wavelet transform`, but I see how that can cause its own problems

lostanlen: i 100% agree

kymatio/__init__.py
edouardoyallon: Okay. I agree, I will remove this.

kymatio/frontend/torch_f.py
edouardoyallon: So - I wanted to do a "template" for a recurrent type of class as it's done in standard OOP. Maybe it's a bad idea here... any suggestions?

kymatio/frontend/torch_f.py
edouardoyallon: The idea is that this function should be erased. In C++ there is an instruction(virtual AFAIR) to say it should be overloead

kymatio/frontend/torch_f.py
edouardoyallon: will be done.

kymatio/scattering2d/backend/__init__.py
edouardoyallon: b for backend. I'd like that the backends related to a frontend (they will be hard to disentangle)  start with a similar prefix, what do you think?

kymatio/scattering2d/scattering2d.py
edouardoyallon: So, IMHO, that's fine, we could simply do a second function scattering2d. What do you think?

edouardoyallon:
I would really appreciate if we had some guidelines concerning my attempt to do a template(is someone familiar with it? To me the interesting thing is to propose an uniform style of front tend for a given framework) here: kymatio/frontend/torch_f.py, as well as some files' names and the functional style I intended to use. I think it's close to be mergeable. Thanks!

kymatio/scattering2d/frontend/torch_f.py
```diff
@@ -0,0 +1,178 @@
+__all__ = ['Scattering2D']
+
+import torch
+from ..backend import cdgmm, Modulus, SubsampleFourier, fft, Pad, unpad, NAME
```
janden: Why not import the whole backend as a module? It gets a little confusing here since `NAME` now refers to the name of the backend used, not the frontend.

kymatio/scattering2d/frontend/torch_f.py
```diff
@@ -0,0 +1,178 @@
...
+from ..filter_bank import filter_bank
+from ..utils import compute_padding
+from ..scattering2d import scattering2d
```
janden: We should probably change the name of this function to distinguish it from the class `Scattering2D`. It's easy to confuse the two right now.

Another thing to think about is the directory structure. It may be better for the `scattering2d` function to be in a `core` submodule (or something like that). Right now, it's a little curious that the main user interface resides in a submodule, while the internal function implementing the transform resides in the module itself.

kymatio/scattering2d/frontend/torch_f.py
```diff
@@ -0,0 +1,178 @@
...
+
+class Scattering2D(Scattering):
+    """Main module implementing the scattering transform in 2D.
+        The scattering transform computes two wavelet transform followed
+        by modulus non-linearity.
+        It can be summarized as::
+
+            S_J x = [S_J^0 x, S_J^1 x, S_J^2 x]
+
+        where::
```
janden: Replace `where` with `for` (we have another `where` just after).

kymatio/scattering2d/frontend/torch_f.py
```diff
@@ -0,0 +1,178 @@
...
+from ..filter_bank import filter_bank
+from ..utils import compute_padding
+from ..scattering2d import scattering2d
+from ...frontend.torch_f import Scattering
+
+if NAME != 'torch' and NAME != 'skcuda':
+    raise (RuntimeError('The only supported backend by the torch frontend are torch and skcuda.'))
+
+class Scattering2D(Scattering):
+    """Main module implementing the scattering transform in 2D.
```
janden: Fix justification of the body text here. Probably best to fit it to 80 characters wide. (Does PEP8 say anything about this?)

kymatio/scattering2d/frontend/torch_f.py
```diff
@@ -0,0 +1,178 @@
...
+
+            S_J x = [S_J^0 x, S_J^1 x, S_J^2 x]
+
+        where::
+
+            S_J^0 x = x * phi_J
+            S_J^1 x = [|x * psi^1_lambda| * phi_J]_lambda
+            S_J^2 x = [||x * psi^1_lambda| * psi^2_mu| * phi_J]_{lambda, mu}
+
+        where * denotes the convolution (in space), phi_J is a low pass
```
janden: `lowpass` and `bandpass`

kymatio/scattering2d/frontend/torch_f.py
```diff
@@ -0,0 +1,178 @@
...
+        filter, psi^1_lambda is a family of band pass
+        filters and psi^2_mu is another family of band pass filters.
+        Only Morlet filters are used in this implementation.
+        Convolutions are efficiently performed in the Fourier domain
+        with this implementation.
```
janden: Remove `with this implementation`.

kymatio/scattering2d/frontend/torch_f.py
```diff
@@ -0,0 +1,178 @@
...
+
+        Example
+        -------
+            # 1) Define a Scattering object as:
```
janden: `Scattering2D`

kymatio/scattering2d/frontend/torch_f.py
```diff
@@ -0,0 +1,178 @@
...
+            s = Scattering2D(J, shape=(M, N))
+            #    where (M, N) are the image sizes and 2**J the scale of the scattering
```
janden: `are` → `is`

kymatio/scattering2d/frontend/torch_f.py
```diff
@@ -0,0 +1,178 @@
...
+            # 2) Forward on an input Tensor x of shape B x M x N,
+            #     where B is the batch size.
+            result_s = s(x)
+
+        Parameters
+        ----------
+        J : int
+            logscale of the scattering
+        shape : tuple of int
```
janden: `ints`?

kymatio/scattering2d/frontend/torch_f.py
```diff
@@ -0,0 +1,178 @@
...
+            s = Scattering2D(J, shape=(M, N))
+            #    where (M, N) are the image sizes and 2**J the scale of the scattering
+            # 2) Forward on an input Tensor x of shape B x M x N,
+            #     where B is the batch size.
+            result_s = s(x)
+
+        Parameters
+        ----------
+        J : int
+            logscale of the scattering
```
janden: `Log-2 of the scattering scale.`

kymatio/scattering2d/frontend/torch_f.py
```diff
@@ -0,0 +1,178 @@
...
+        shape : tuple of int
+            spatial support (M, N) of the input
```
janden: Capitalize and add period (goes for below as well).

kymatio/scattering2d/frontend/torch_f.py
```diff
@@ -0,0 +1,178 @@
...
+            `1` or `2`. Defaults to `2`.
+        pre_pad : boolean, optional
+            controls the padding: if set to False, a symmetric padding is applied
+            on the signal. If set to true, the software will assume the signal was
+            padded externally.
+
+        Attributes
+        ----------
+        J : int
+            logscale of the scattering
```
janden: See above.

kymatio/scattering2d/frontend/torch_f.py
```diff
@@ -0,0 +1,178 @@
...
+        shape : tuple of int
+            spatial support (M, N) of the input
+        L : int, optional
+            number of angles used for the wavelet transform
+        max_order : int, optional
+            The maximum order of scattering coefficients to compute.
+            Must be either equal to `1` or `2`. Defaults to `2`.
+        pre_pad : boolean
+            controls the padding
```
janden: Be more specific.

kymatio/scattering2d/frontend/torch_f.py
```diff
@@ -0,0 +1,178 @@
...
+        Psi : dictionary
+            containing the wavelets filters at all resolutions. See
```
janden: `containing` → `contains` (same below)

kymatio/scattering2d/frontend/torch_f.py
```diff
@@ -0,0 +1,178 @@
...
+            filter_bank.filter_bank for an exact description.
+        Phi : dictionary
+            containing the low-pass filters at all resolutions. See
+            filter_bank.filter_bank for an exact description.
+        M_padded, N_padded : int
+             spatial support of the padded input
+
+        Notes
+        -----
+        The design of the filters is optimized for the value L = 8
```
janden: Add period.

kymatio/scattering2d/frontend/torch_f.py
```diff
@@ -0,0 +1,178 @@
...
+
+        pre_pad is particularly useful when doing crops of a bigger
```
janden: `when doing crops of a bigger image` → `when cropping bigger images`

Also, how is padding more or less accurate?

kymatio/scattering2d/frontend/torch_f.py
```diff
@@ -0,0 +1,178 @@
...
+            result_s = s(x)
+
+        Parameters
+        ----------
+        J : int
+            logscale of the scattering
+        shape : tuple of int
+            spatial support (M, N) of the input
+        L : int, optional
+            number of angles used for the wavelet transform
```
janden: What's the default?

kymatio/scattering2d/frontend/torch_f.py
```diff
@@ -0,0 +1,178 @@
...
+        max_order : int, optional
+            The maximum order of scattering coefficients to compute. Must be either
+            `1` or `2`. Defaults to `2`.
+        pre_pad : boolean, optional
+            controls the padding: if set to False, a symmetric padding is applied
+            on the signal. If set to true, the software will assume the signal was
+            padded externally.
```
janden: What's the default?

kymatio/scattering2d/frontend/torch_f.py
```diff
@@ -0,0 +1,178 @@
...
+        M_padded, N_padded : int
+             spatial support of the padded input
+
+        Notes
+        -----
+        The design of the filters is optimized for the value L = 8
+
+        pre_pad is particularly useful when doing crops of a bigger
+         image because the padding is then extremely accurate. Defaults
+         to False.
```
janden: Default shouldn't be here but in the documentation of the constructor arguments.

kymatio/scattering2d/frontend/torch_f.py
```diff
@@ -0,0 +1,178 @@
...
+
+        """
+    def __init__(self, J, shape, L=8, max_order=2, pre_pad=False):
+        super(Scattering2D, self).__init__(J, shape, max_order = max_order)
```
janden: No space around `=`.

Also, what do we expect this to do? I know that right now it does nothing, but is there some common functionality that needs to be reproduced independent of dimension?

kymatio/scattering2d/frontend/torch_f.py
```diff
@@ -0,0 +1,178 @@
...
+        self.L, self.J, self.max_order = L, J, max_order
```
janden: Isn't `self.J`, `self.max_order` (and `self.shape`, below) assigned in `super`?

kymatio/scattering2d/frontend/torch_f.py
```diff
@@ -0,0 +1,178 @@
...
+        self.pre_pad = pre_pad
+        self.shape = shape
+        if 2 ** J > shape[0] or 2 ** J > shape[1]:
+            raise (RuntimeError('The smallest dimension should be larger than 2^J'))
```
janden: Parentheses around `RuntimeError`?

kymatio/scattering2d/frontend/torch_f.py
```diff
@@ -0,0 +1,178 @@
...
+        self.M, self.N = self.shape
+        self.modulus = Modulus()
+        self.M_padded, self.N_padded = compute_padding(self.M, self.N, self.J)
+        # pads equally on a given side if the amount of padding to add is an even number of pixels, otherwise it adds an extra pixel
+        self.pad = Pad([(self.N_padded - self.N) // 2, (self.N_padded - self.N + 1) // 2, (self.M_padded - self.M) // 2,
+                       (self.M_padded - self.M + 1) // 2], [self.N, self.M], pre_pad=self.pre_pad)
+        self.subsample_fourier = SubsampleFourier()
+        self.filters = self.create_and_register_filters()
+
+    def create_and_register_filters(self):
```
janden: See comment in parent class.

kymatio/scattering2d/frontend/torch_f.py
```diff
@@ -0,0 +1,178 @@
...
+
+        for j in range(len(self.psi)):
+            for k, v in self.psi[j].items():
+                if isinstance(k, int):
+                    self.psi[j][k] = torch.from_numpy(v)
+                    self.register_buffer('tensor' + str(n), self.psi[j][k])
+                    n += 1
+
+    def scattering(self, input):
+        # each time scattering is run, one needs to make sure self.psi and self.phi point to
```
janden: Why can't this be done once and for all? Do the buffers move around? If so, what's to guarantee that they stay the same during the call to `scattering2d` below?

kymatio/scattering2d/frontend/torch_f.py
```diff
@@ -0,0 +1,178 @@
...
+        self.filters = self.create_and_register_filters()
+
+    def create_and_register_filters(self):
+        # Create the filters
+        filters = filter_bank(self.M_padded, self.N_padded, self.J, self.L)
+        n = 0
+        self.phi, self.psi = filters['phi'], filters['psi']
+        for c, phi in self.phi.items():
+            if isinstance(c, int):
+                self.phi[c] = torch.from_numpy(self.phi[c])
```
janden: This loop does two things: convert from numpy to torch, and register the torch arrays. Should this be two different loops (or two different functions)? Maybe the former should happen in a `create_filters` function and the latter in a `register_filters` function?

kymatio/scattering2d/frontend/torch_f.py
```diff
@@ -0,0 +1,178 @@
...
+                self.register_buffer('tensor' + str(n), self.phi[c])
+                n += 1
+
+        for j in range(len(self.psi)):
+            for k, v in self.psi[j].items():
+                if isinstance(k, int):
+                    self.psi[j][k] = torch.from_numpy(v)
+                    self.register_buffer('tensor' + str(n), self.psi[j][k])
```
janden: Is there some more elegant (and safer) way to do this registration? It seems that this linear encoding is quite brittle.

kymatio/scattering2d/frontend/torch_f.py
```diff
@@ -0,0 +1,178 @@
...
+                self.phi[c] =  buffer_dict['tensor' + str(n)]
+                n += 1
+
+        for j in range(len(self.psi)):
+            for k, v in self.psi[j].items():
+                if isinstance(k, int):
+                    self.psi[j][k] = buffer_dict['tensor' + str(n)]
+                    n += 1
+
+        return scattering2d(input, self.J, self.L, self.subsample_fourier, self.pad, self.modulus, fft, cdgmm, unpad, self.phi, self.psi, self.max_order, self.M_padded, self.N_padded)
```
janden: Why not just pass `self` and `backend`? Long argument lists like this are rarely a good idea.

kymatio/scattering2d/frontend/torch_f.py
```diff
@@ -0,0 +1,178 @@
...
+
+    def forward(self, input):
+        """Forward pass of the scattering.
+
+        Parameters
+        ----------
+        input : tensor
+           tensor with 3 dimensions :math:`(B, C, M, N)` where :math:`(B, C)` are arbitrary.
```
janden: Capitalize.

Also, it says three but we list four. What is this, the Spanish Inquisition?

kymatio/scattering2d/frontend/torch_f.py
```diff
@@ -0,0 +1,178 @@
...
+           :math:`B` typically is the batch size, whereas :math:`C` is the number of input channels.
+
+        Returns
+        -------
+        S : tensor
+           scattering of the input, a 4D tensor :math:`(B, C, D, Md, Nd)` where :math:`D` corresponds
```
janden: Capitalize.

kymatio/scattering2d/frontend/torch_f.py
```diff
@@ -0,0 +1,178 @@
...
+           to a new channel dimension and :math:`(Md, Nd)` are downsampled sizes by a factor :math:`2^J`.
+
+        """
+        if not torch.is_tensor(input):
+            raise (
+            TypeError('The input should be a torch.cuda.FloatTensor, a torch.FloatTensor or a torch.DoubleTensor'))
```
janden: Why not just say `The input should be a PyTorch Tensor` (or something like that)?

kymatio/scattering2d/frontend/torch_f.py
```diff
@@ -0,0 +1,178 @@
...
+
+        if len(input.shape) < 2:
+            raise (RuntimeError('Input tensor must have at least two '
```
janden: Remove parentheses.

kymatio/scattering2d/frontend/torch_f.py
```diff
@@ -0,0 +1,178 @@
...
+                               'dimensions'))
+
+        if (not input.is_contiguous()):
+            raise (RuntimeError('Tensor must be contiguous!'))
```
janden: Remove parentheses.

kymatio/scattering2d/frontend/torch_f.py
```diff
@@ -0,0 +1,178 @@
...
+
+        if ((input.size(-1) != self.N or input.size(-2) != self.M) and not self.pre_pad):
+            raise (RuntimeError('Tensor must be of spatial size (%i,%i)!' % (self.M, self.N)))
```
janden: Remove parentheses.

kymatio/scattering2d/frontend/torch_f.py
```diff
@@ -0,0 +1,178 @@
...
+
+        if ((input.size(-1) != self.N_padded or input.size(-2) != self.M_padded) and self.pre_pad):
+            raise (RuntimeError('Padded tensor must be of spatial size (%i,%i)!' % (self.M_padded, self.N_padded)))
```
janden: Remove parentheses.

kymatio/scattering2d/frontend/torch_f.py
```diff
@@ -0,0 +1,178 @@
...
+
+        return self.scattering(input)
```
janden: Add newline.

kymatio/scattering2d/tests/test_torch_scattering2d.py
```diff
@@ -11,14 +11,14 @@
 backends = []
```
janden: Don't think we need to rename this script (at least not yet).

Ideally, we'd like to test all the frontends in the same way, so some clever way or parametrizing the tests with respect to this will be needed.

kymatio/scattering2d/scattering2d.py
```diff
@@ -2,255 +2,69 @@
...
+
+    order0_size = 1
+    order1_size = L * J
+    order2_size = L ** 2 * J * (J - 1) // 2
+    output_size = order0_size + order1_size
+
+    if max_order == 2:
+        output_size += order2_size
+
+    S = input.new(input.size(0),
```
janden: Is calling `input.new` generally safe across frontends (same for `input.reshape`, `input.shape`, and `input.size`)?

Also, we should stick to one of `shape` and `size`. (I believe `shape` is preferred.)

kymatio/scattering2d/scattering2d.py
```diff
@@ -2,255 +2,69 @@
...
+                  input.size(1),
+                  output_size,
+                  M_padded//(2**J)-2,
+                  N_padded//(2**J)-2)
+    U_r = pad(input)
+    U_0_c = fft(U_r, 'C2C')  # We trick here with U_r and U_2_c
```
janden: What's the trick?

kymatio/scattering2d/scattering2d.py
```diff
@@ -2,255 +2,69 @@
...
+
+    # First low pass filter
+    U_1_c = subsample_fourier(cdgmm(U_0_c, phi[0]), k=2**J)
```
janden: The naming confuses me here. This is the Fourier transform of `S_0`, correct? Same below with `U_J_r`.

kymatio/scattering2d/scattering2d.py
```diff
@@ -2,255 +2,69 @@
...
+        j1 = psi[n1]['j']
+        U_1_c = cdgmm(U_0_c, psi[n1][0])
+        if(j1 > 0):
+            U_1_c = subsample_fourier(U_1_c, k=2 ** j1)
+        U_1_c = fft(U_1_c, 'C2C', inverse=True)
+        U_1_c = fft(modulus(U_1_c), 'C2C')
+
+        # Second low pass filter
+        U_2_c = subsample_fourier(cdgmm(U_1_c, phi[j1]), k=2**(J-j1))
+        U_J_r = fft(U_2_c, 'C2R')
```
janden: Why not `inverse=True`?

Also naming again.

kymatio/scattering2d/scattering2d.py
```diff
@@ -2,255 +2,69 @@
...
+    for n1 in range(len(psi)):
+        j1 = psi[n1]['j']
+        U_1_c = cdgmm(U_0_c, psi[n1][0])
+        if(j1 > 0):
+            U_1_c = subsample_fourier(U_1_c, k=2 ** j1)
+        U_1_c = fft(U_1_c, 'C2C', inverse=True)
+        U_1_c = fft(modulus(U_1_c), 'C2C')
+
+        # Second low pass filter
+        U_2_c = subsample_fourier(cdgmm(U_1_c, phi[j1]), k=2**(J-j1))
```
janden: Naming is off again here. Shouldn't it be `S_1_c`?

kymatio/scattering2d/scattering2d.py
```diff
@@ -2,255 +2,69 @@
...
+        U_J_r = fft(U_2_c, 'C2R')
+        S[..., n_order1, :, :] = unpad(U_J_r)
+        n_order1 += 1
+
+        if max_order == 2:
+            for n2 in range(len(psi)):
+                j2 = psi[n2]['j']
+                if(j1 < j2):
+                    U_2_c = subsample_fourier(cdgmm(U_1_c, psi[n2][j1]), k=2 ** (j2-j1))
```
janden: Naming.

kymatio/scattering2d/scattering2d.py
```diff
@@ -2,255 +2,69 @@
...
+                    U_2_c = fft(U_2_c, 'C2C', inverse=True)
+                    U_2_c = fft(modulus(U_2_c), 'C2C')
+
+                    # Third low pass filter
+                    U_2_c = subsample_fourier(cdgmm(U_2_c, phi[j2]), k=2 ** (J-j2))
+                    U_J_r = fft(U_2_c, 'C2R')
```
janden: Naming. `inverse=True`?

janden:
So if I've understood this correctly, this PR includes a frontend class `Scattering2D`, which takes torch tensors in and splits torch tensors out. Internally, it preps all the torch-specific stuff (mostly filters) and sends these off to `scattering2d` when doing the forward pass. That function, in turn, calls the various backend primitives (which are also supplied by `Scattering2D` when calling `scattering2d`).

The plan is then to have different frontends for different types of inputs (in other words, depending on what types of tensors the user supplied) that pass different types of tensors, filters, and backend primitives to `scattering2d`? How is the selection of the frontend going to happen? Right now, it looks like it just takes the torch one.

edouardoyallon:
> So if I've understood this correctly, this PR includes a frontend class `Scattering2D`, which takes torch tensors in and splits torch tensors out. Internally, it preps all the torch-specific stuff (mostly filters) and sends these off to `scattering2d` when doing the forward pass. That function, in turn, calls the various backend primitives (which are also supplied by `Scattering2D` when calling `scattering2d`).

Absolutely!
 
> The plan is then to have different frontends for different types of inputs (in other words, depending on what types of tensors the user supplied) that pass different types of tensors, filters, and backend primitives to `scattering2d`?

So, if by type of tensors you mean `half`, `float` or `double`, then, this should be handled by the internal routines of `nn.Module()` of pytorch, given that the backend supports it. There might be a short discussion about the default type of tensor we use in the whole software and documentation(e.g., np, float32, cuda.float64?)

> How is the selection of the frontend going to happen? Right now, it looks like it just takes the torch one.

I think this should be specified by the user when loading the module, with for instance a command like:
```from kymatio import torch.frontend```

We might be willing to rethink the way we pass backends. We might also be willing to propose a default frontend for a given backend(that might make some sens), or the inverse. We might also be willing to specify as an external variable the frontend (e.g., `KYMATIO_FRONTEND=...`)

I'm up to anything.. even changing it later, depending on the usages.

File: kymatio/frontend/torch_f.py (hunk 2 above)
Comment (edouardoyallon) [reply kymatio/frontend/torch_f.py]: fixed

File: kymatio/frontend/torch_f.py [hunk 50]
```diff
@@ -0,0 +1,23 @@
...
+       self.J = J
+       self.shape = shape
+       self.max_order = max_order
+       self.build()
+
+   def build(self):
+       return
+       self.filters = self.create_and_register_filters()
+
+   def create_and_register_filters(self):
```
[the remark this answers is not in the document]
edouardoyallon: I added a docstring, indeed, it wasn't clear

kymatio/scattering2d/frontend/torch_f.py
edouardoyallon: Not atm the parameters are overloaded.. the `super` is simply not used.(again this cass is only a template)

kymatio/scattering2d/frontend/torch_f.py
edouardoyallon: I quite disagree here! It is the simplest way to make the function `scattering2d` self-suffisant. Otherwise, for other frontends/backends, we might be missing some modules. Also, frontends which do not have a "class" structure might suffer from this.(e.g., tf)

kymatio/scattering2d/scattering2d.py
edouardoyallon: No, it is not. I was thinking to leave this for a future fix.(e.g., including numpy backend)

I also think shape is preferred.

kymatio/scattering2d/frontend/torch_f.py
edouardoyallon: I've improved slightly the doc

kymatio/scattering2d/frontend/torch_f.py
edouardoyallon: Yeah, for now, at least initializing the parameters. I think shape, J and max order are inherent to any scattering transform...

kymatio/scattering2d/frontend/torch_f.py
edouardoyallon: Do you know an automatic way to switch to 79 characters wide? (like, I don't see myself doing it manually for the whole pkg)

kymatio/scattering2d/frontend/torch_f.py
edouardoyallon: It wasn't done cleanly, it's indeed done now. Sorry, I got confused with python's oop

kymatio/scattering2d/scattering2d.py
edouardoyallon: This is linked to the functional aspect of the pkg, not the frontend's one. I think there are several arguments in favor of one or the other, yet it might be discussed in another future PR if someone is interested to change this...

kymatio/scattering2d/frontend/torch_f.py
edouardoyallon: Hmm that would work, it requires a little more work, and also, it sort of "hardcodes" what is a backend.(right now, we could evolve freely to what you propose or something simpler)

@lostanlen @eickenberg any opinions?

kymatio/scattering2d/frontend/torch_f.py
edouardoyallon: > We should probably change the name of this function to distinguish it from the class Scattering2D. It's easy to confuse the two right now.

So, I proposed to use the same convention as pytorch. As an example: https://pytorch.org/docs/stable/_modules/torch/nn/modules/linear.html

> Another thing to think about is the directory structure. It may be better for the `scattering2d` function to be in a `core` submodule (or something like that). Right now, it's a little curious that the main user interface resides in a submodule, while the internal function implementing the transform resides in the module itself.

I do agree. For this one, I'm not sure what's the best solution. I'm up to any solution... maybe we could call it functional like pytorch? But then we do copy pytorch and tf people might not be happy.. I don't know what to do..!

kymatio/scattering2d/frontend/torch_f.py
edouardoyallon: They do move around. The array buffer is an array of "pointer". Here is an url to an example where it is nicely used: https://pytorch.org/docs/stable/_modules/torch/nn/modules/batchnorm.html (for BNs) I can't think of a simpler way to use it because we need an array/dictionary and the buffers do not seem able to track each tensors of a dictionary. Thus currently, I artifically expand it and track it..

The only guarantee is that the registered pointer will point to the current buffer. An example of failure is when one does `.cuda()`, then the memory changes..

kymatio/scattering2d/frontend/torch_f.py
edouardoyallon: I think that's fine because one sees I had to trick to track the registered buffers. Those functions will be updated when torch will improve its buffer routines.

kymatio/scattering2d/frontend/torch_f.py
edouardoyallon: I wish there was.

kymatio/scattering2d/frontend/torch_f.py
edouardoyallon: Also another point is that this initialization will be different for 1/3-D Scattering. Ultimately, we would like this manner to instantiate the buffers to be identical for every scattering. This can't be done curretnly because we need more harmonization between 1-2-3 D.

kymatio/scattering2d/tests/test_torch_scattering2d.py
edouardoyallon: Curretnly, we want to maintain this testing script for making sure that the introduced frontend does not change anything to the documentation and the highlevel use of advanced users.

Later, we would like a minimal script to test the frontend, and we would like to separate backend and frontend tests. Also, we will need to avoid being redundant with torch, and probably to use torch set of testing routines.(maybe there is like in `sklearn` a routine `test_user_module()` or similar) **Very soon, we would like to ensure that the multi-GPU works with the current frontend without any additional changes.**

kymatio/scattering2d/scattering2d.py
edouardoyallon: it doesn't apply here anymore.

kymatio/scattering2d/scattering2d.py
edouardoyallon: This naming was due to former optimization of memory(in order to re-use intermediary variables to save memory) I think this is not necessary anymore!

kymatio/scattering2d/scattering2d.py
edouardoyallon: I've added an error message in the fft's backend

kymatio/scattering2d/backend/__init__.py
edouardoyallon: I've chosen the last naming convention (i.e., adding *_backend* at the end)

edouardoyallon:
@janden I think I've done all the changes requested up to some discussions:

1. I'm in favor to add the functional scattering in a `core` folder as you suggested. **_Should I do it?_**
2. I'm not sure how the frontend should be imported (cf discussion).. @eickenberg @lostanlen ?
3. An array of buffer isn't currently well handled by torch.. so I'm ok to let my trick like it is. 
4. All minor modifications linked to style, writing, coding style or variable naming have been applied.

Please resolve any comments I left or re-open some I closed. Thanks!

kymatio/scattering2d/frontend/torch_frontend.py
```diff
@@ -0,0 +1,181 @@
...
+
+import torch
+from ..backend import cdgmm, Modulus, SubsampleFourier, fft, Pad, unpad, BACKEND_NAME
+from ..filter_bank import filter_bank
+from ..utils import compute_padding
+from ..scattering2d import scattering2d
+from ...frontend.torch_frontend import Scattering
+
+if BACKEND_NAME != 'torch' and BACKEND_NAME != 'skcuda':
+    raise (RuntimeError('The only supported backend by the torch frontend are torch and skcuda.'))
```
eickenberg: This line might lower code coverage, because I don't think it can ever be reached (I think the check in `..backend` will always raise before this raises)

kymatio/scattering2d/frontend/torch_frontend.py
```diff
@@ -0,0 +1,181 @@
...
+        The design of the filters is optimized for the value L = 8.
+
+        pre_pad is particularly useful when cropping bigger images because
+        this does not introduce border effects inherent to padding.
+
+        """
+    def __init__(self, J, shape, L=8, max_order=2, pre_pad=False):
+        super(Scattering2D, self).__init__(J, shape, max_order=max_order)
+        self.pre_pad, self.L = pre_pad, L
+        if 2 ** J > shape[0] or 2 ** J > shape[1]:
```
eickenberg: move to `build`

kymatio/scattering2d/frontend/torch_frontend.py
```diff
@@ -0,0 +1,181 @@
...
+
+    def create_and_register_filters(self):
+        """ This function run the filterbank function that
+            will create the filters as numpy array, and then, it
+            saves those arrays as module's buffers."""
+
+        # Create the filters
+        filters = filter_bank(self.M_padded, self.N_padded, self.J, self.L)
+        n = 0
+        self.phi, self.psi = filters['phi'], filters['psi']
```
eickenberg: cosmetic: These should probably not be stored in `self` here, but rather just in names `phi`, `psi`, and then stored in a new dictionary lower down. probably don't change, since it might require adding boiler plate code

kymatio/scattering2d/frontend/torch_frontend.py
```diff
@@ -0,0 +1,181 @@
...
+        for j in range(len(self.psi)):
+            for k, v in self.psi[j].items():
+                if isinstance(k, int):
+                    self.psi[j][k] = torch.from_numpy(v)
+                    self.register_buffer('tensor' + str(n), self.psi[j][k])
+                    n += 1
+
+    def scattering(self, input):
+        # each time scattering is run, one needs to make sure self.psi and self.phi point to
+        # the correct buffers
```
eickenberg: Is this because of potential `.to` and `.cuda` calls in the meantime?

If so, wouldn't it be semantically more correct to overwrite `.cuda` and `.to` to first call `super().to/cuda` and then make the modification in the dict?

kymatio/scattering2d/frontend/torch_frontend.py
```diff
@@ -0,0 +1,181 @@
...
+                n += 1
+
+        for j in range(len(self.psi)):
+            for k, v in self.psi[j].items():
+                if isinstance(k, int):
+                    self.psi[j][k] = torch.from_numpy(v)
+                    self.register_buffer('tensor' + str(n), self.psi[j][k])
+                    n += 1
+
+    def scattering(self, input):
```
eickenberg: What is the main reason to have a method `.scattering` independent of `forward`?

kymatio/scattering2d/scattering2d.py
eickenberg: 1. pragmatic question: Does this work with any other setting than 2d + torch? Would this framework be able to accommodate tf? (probably yes)

2. I get that the depth-first search nature of our scattering implementations requires being a bit more low-level than one would want. I'm still not 1000% convinced this is the right abstraction, but it might be worth going with for the moment. If it hits a road block, we'll have to change it

kymatio/scattering2d/scattering2d.py
eickenberg: `shape` and `reshape` should work for most frameworks.
If there is divergence, then we can write backend-specific methods called `safe_*` like `safe_new` which can make a new tensor of whatever the backend is.

Now I am wondering about how this whole thing will be impacted by the fact that TF builds compute graphs. Keras managed to handle it, so hopefully, with the build method, we should able to do so as well.

kymatio/scattering2d/scattering2d.py
eickenberg: is this now unconfused?

kymatio/scattering2d/tests/test_torch_scattering2d.py
eickenberg: > Very soon, we would like to ensure that the multi-GPU works with the current frontend without any additional changes.

that seems like a good goal and fire test for this approach

kymatio/frontend/torch_frontend.py
```diff
@@ -0,0 +1,25 @@
...
+class Scattering(nn.Module):
+   def __init__(self, J, shape, max_order = None):
+       super(Scattering, self).__init__()
+       self.J = J
+       self.shape = shape
+       self.max_order = max_order
+
+   def build(self):
+       self.filters = self.create_and_register_filters()
+       raise NotImplementedError
```
eickenberg: I think you might have to instantiate the error. Otherwise you'll get an error while raising the error :D

eickenberg:
Would it be possible to start including one other backend/frontend, e.g. TF in order to see that this works and is the right way?

eickenberg:
> I'm not sure how the frontend should be imported (cf discussion)..

e.g. `from kymatio.torch import Scattering2D`

I guess this would only be the frontend then. Maybe we need

`from kymatio.torch_torch import Scattering2D`
`from kymatio.torch_skcuda import Scattering2D`

and `kymatio.torch.Scattering2D` points to whatever default backend was specified?

Meh. not sure

janden:
> So, if by type of tensors you mean half, float or double, then, this should be handled by the internal routines of nn.Module() of pytorch, given that the backend supports it. There might be a short discussion about the default type of tensor we use in the whole software and documentation(e.g., np, float32, cuda.float64?)

Actually, by “type of tensors” I meant torch vs. tf vs. numpy, and so on. I like the way you suggest to handle it with explicit imports, though.

janden:
> 1. I'm in favor to add the functional scattering in a `core` folder as you suggested. **_Should I do it?_**

Yes.

> 2-3. Blah blah blah.

Ok.

janden:
> `from kymatio.torch_torch import Scattering2D`

Ugh no. Can we just keep backend selection as is? I guess that won't work if the user selects a frontend that's incompatible with the backend. Maybe just have it bark at the user to change their backend for now.

I'm leaning towards doing this programmatically, in the end, selecting `backend='torch'` or something when creating the `Scattering2D` object. This should be a different PR, though.

lostanlen:
Not sure i follow the rationale for having a frontend different from the backend?
Does that mean we have a quadratic explosion of frontend-backend pairs?

janden:
> Not sure i follow the rationale for having a frontend different from the backend?
> Does that mean we have a quadratic explosion of frontend-backend pairs?

Not necessarily since there are compatibility considerations (e.g., TF with a torch backend is not allowed). It should still be *roughly* linear.

kymatio/frontend/torch_frontend.py
janden: Both work: https://stackoverflow.com/questions/13052991/proper-way-to-raise-exception-in-python

kymatio/scattering2d/backend/__init__.py
janden: :ok_hand:

kymatio/scattering2d/backend/torch_b.py
```diff
@@ -178,10 +168,10 @@ def fft(input, direction='C2C', inverse=False):
         raise (RuntimeError('Tensors must be contiguous!'))
 
     if direction == 'C2R':
-        output = torch.irfft(input, 2, normalized=False, onesided=False)*input.size(-2)*input.size(-3)
+        output = torch.irfft(input, 2, normalized=False, onesided=False) * input.size(-2) * input.size(-3)
```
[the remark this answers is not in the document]
janden: Ok fine.

kymatio/scattering2d/frontend/torch_f.py
janden: > it sort of "hardcodes" what is a backend.

I don't understand?

At the very least, we should do `from ..backend import NAME as BACKEND_NAME`.

kymatio/scattering2d/frontend/torch_f.py
janden: > So, I proposed to use the same convention as pytorch.

Ok. Makes sense (although `L` and `l` are more easily distinguishable than `S` and `s`).

kymatio/scattering2d/frontend/torch_f.py
janden: In vim it's `set tw=79`, then it wraps as you type. You can type `gq` to reformat old lines. Most other editors have something similar.

kymatio/scattering2d/frontend/torch_f.py
janden: Right. Saw that later.

kymatio/scattering2d/frontend/torch_f.py
janden: So why do we have to fetch them during the `scattering()` call? When we register them, we get the pointers to the registered buffers and assign them to our filters already.

kymatio/scattering2d/frontend/torch_f.py
janden: > I think that's fine because one sees I had to trick to track the registered buffers. 

I don't understand. What's the trick?

kymatio/scattering2d/frontend/torch_f.py
janden: Ok for now, but it would be good to have some function that just registers all the tensors in a dictionary in a standard way and another that extracts them.

kymatio/scattering2d/frontend/torch_f.py
janden: Ok sure. But can we at least package things into a dict or something instead of a long argument list?

janden:
> Would it be possible to start including one other backend/frontend, e.g. TF in order to see that this works and is the right way?

I think this is a great idea. Numpy might be a better candidate, though.

kymatio/scattering2d/tests/test_torch_scattering2d.py
janden: Ok so we'll just copy this file for each frontend for now? Seems like a mess...

kymatio/scattering2d/scattering2d.py
janden: Ok makes sense.

edouardoyallon:
> > Would it be possible to start including one other backend/frontend, e.g. TF in order to see that this works and is the right way?
> 
> I think this is a great idea. Numpy might be a better candidate, though.

I'll add a PR for a numpy frontend and backend on top of this one asap

kymatio/scattering2d/frontend/torch_frontend.py
edouardoyallon: that might not be necessary.. I think the forward might need some intermediary work that the scattering function does not.(first example in mind, the batch size is currently handled for 4D tensor, not beyond but this could be easily done in the forward // multi-GPU things maybe)

janden:
> > > Would it be possible to start including one other backend/frontend, e.g. TF in order to see that this works and is the right way?
> > 
> > 
> > I think this is a great idea. Numpy might be a better candidate, though.
> 
> I'll add a PR for a numpy frontend and backend on top of this one asap

Why don't we finish working on this one first and merge it into a dev branch, then you can PR on top of that one. PRs on top of PRs just make my head hurt.

kymatio/scattering2d/frontend/torch_frontend.py
edouardoyallon: NO! We should never handle those functions on our own, except for raising an error for types not supported yet by backends(eg, `.half()`)

edouardoyallon: (also pytorch has some local optimization we do not wish to deal with)

kymatio/scattering2d/frontend/torch_frontend.py
edouardoyallon: well, for now, I view it more as hack: for instance, it allows to (possibly) memorize in which order the buffers were stored

kymatio/scattering2d/scattering2d.py
edouardoyallon: 1. AFAIR, yes. If I encounter a necessary change for 1/3-D, we'll do it.
2. OK, well, see 1/.

kymatio/scattering2d/frontend/torch_f.py
edouardoyallon: I've switched it to "BACKEND_NAME". What I meant is that if the backend is a NN module, it can be restrictive... I think this could be changed in a future PR?

kymatio/scattering2d/frontend/torch_f.py
edouardoyallon: Well, for the future!

kymatio/scattering2d/frontend/torch_f.py
edouardoyallon: at the moment we do a .cuda(), the location changes - it's in order to guarantee that the pointer does pooint to the correct memory location.

kymatio/scattering2d/frontend/torch_f.py
edouardoyallon: Sure..

kymatio/scattering2d/frontend/torch_f.py
edouardoyallon: storing each element of the array as a variable 'buffer_$i' where $i is incremented.

kymatio/scattering2d/tests/test_torch_scattering2d.py
edouardoyallon: for now, let's maintain the test. In a future PR(on the new branch `kymatio-dev`), we'll have to do a serie of test for the `frontend`, `core` and `backend` folders. I think it's the simplest.

edouardoyallon:
> > `from kymatio.torch_torch import Scattering2D`
> 
> Ugh no. Can we just keep backend selection as is? I guess that won't work if the user selects a frontend that's incompatible with the backend. Maybe just have it bark at the user to change their backend for now.

How do you want to set up the frontend and/or the backend?

PR #405, opened 2019-07-09
FIX+TST Emit proper warnings in 3D backends and test for them

### LINKED ISSUE #339: Test for raised error and warnings in 3D torch backend

janden:
The first part ensures that the skcuda backend emits the same warnings as the torch backend for non-contiguous arrays in `cdgmm`. The second part ensures that these (and other errors) are appropriately tested.

Fixes #339 and #408.

### DIFF
--- a/kymatio/scattering3d/backend/backend_skcuda.py
+++ b/kymatio/scattering3d/backend/backend_skcuda.py
@@ -1,4 +1,5 @@
 import torch
+import warnings
 from skcuda import cublas
 
 NAME = 'skcuda'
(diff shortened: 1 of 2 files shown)


janden:
Why doesn't codecov comment on this? Did we do something to make it upset?

PR #412, opened 2019-07-19
Fix two bugs for the max scale case (final output size of 1).

### LINKED ISSUE #363: Scattering2D doesn't work with 2**J == image_size

Kegnarok:
In addition to what I discussed in https://github.com/kymatio/kymatio/issues/363#issuecomment-513037055, I fixed a bug for the case `M != N`, because pytorch expects its padding sizes in the reverse order.

Hopefully fixes #363 for real.

### DIFF
--- a/kymatio/scattering2d/backend/backend_skcuda.py
+++ b/kymatio/scattering2d/backend/backend_skcuda.py
@@ -56,7 +56,7 @@ class Pad(object):
             Parameters
             ----------
             pad_size : list of 4 integers
-                size of padding to apply [left, right, top, bottom].
+                size of padding to apply [top, bottom, left, right].
             input_size : list of 2 integers
                 size of the original signal [height, width].
             pre_pad : boolean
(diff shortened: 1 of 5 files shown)


kymatio/scattering2d/backend/backend_skcuda.py
```diff
@@ -69,23 +69,25 @@ def __init__(self, pad_size, input_size, pre_pad=False):
         self.build()
 
     def build(self):
-        pad_size_tmp = self.pad_size
+        pad_size_tmp = list(self.pad_size)
```
eickenberg: What is this for? What other than list/tuple/array can be expected as input here requiring this? An iterator?

eickenberg: OH. it's supposed to make a copy? And you want to modify it lower down, so tuples are out

kymatio/scattering2d/backend/backend_skcuda.py
```diff
@@ -69,23 +69,25 @@ def __init__(self, pad_size, input_size, pre_pad=False):
...
 
         # This allow to handle the case where the padding is equal to the image size
-        if pad_size_tmp[0] == self.input_size[0]:
-            pad_size_tmp[0] -= 1
-            pad_size_tmp[1] -= 1
-        if pad_size_tmp[2] == self.input_size[1]:
+        # We need to take care that the order of the pad sizes is reversed with
```
eickenberg: (Yeah, uh ... shouldn't we just make this more consistent? Especially given 3D. Probably axis ordering would be the easiest)

kymatio/scattering2d/backend/backend_skcuda.py
```diff
@@ -69,23 +69,25 @@ def __init__(self, pad_size, input_size, pre_pad=False):
...
+        # respect to the input size.
+        if pad_size_tmp[2] == self.input_size[0]:
             pad_size_tmp[2] -= 1
             pad_size_tmp[3] -= 1
+        if pad_size_tmp[0] == self.input_size[1]:
+            pad_size_tmp[0] -= 1
```
eickenberg: For my own edification: I have been wondering through a bunch of potential reasons for the `-1` here. What is the actual reason?

eickenberg: (My No. 1: Because it's reflection padding and you don't want to copy the one right at the border?)

kymatio/scattering2d/backend/backend_skcuda.py
Kegnarok: Yes this is to make a copy. Besides, this means you now can pass tuples in, which I think doesn't hurt.

kymatio/scattering2d/backend/backend_skcuda.py
eickenberg: Actually, I'd prefer this to error instead of making such a modification (which will confuse users, who will be left with the feeling that the software doesn't do what they asked of it.). Either error or continue with a new reflection beyond the first reflection.

Kegnarok: The problem is that reflection padding isn't defined for a padding equal to the input size: you can extend `0 1 2` to `2 1 | 0 1 2 | 1 0` but no further. So in order to add that extra padding we manually do a second reflection: `1 2 1 | 0 1 2 | 2 1 2`.

kymatio/scattering2d/backend/backend_skcuda.py
edouardoyallon: I'm confused too. Please fix

kymatio/scattering2d/backend/backend_skcuda.py
edouardoyallon: Sounds good to me

PR #413, opened 2019-07-20
ENH (merge?) backend/frontend for numpy

edouardoyallon:


edouardoyallon:
Do we agree that for this backend, we do use complex numbers as implemented in numpy?

lostanlen:
> Do we agree that for this backend, we do use complex numbers as implemented in numpy?

yes, that would be great

eickenberg:
> Do we agree that for this backend, we do use complex numbers as implemented in numpy?

What property are you talking about specifically? That numpy by default uses double precision? Otherwise it's the same as in pytorch:

try:
`a = (np.arange(10) + 1j * np.arange(10, 20)).astype('complex64')`

Then:
`b = a.view()`
`b.shape = b.shape + (1,)`
`b.dtype='float32'`
`print(b.shape)`

kymatio/scattering2d/backend/numpy_backend.py
```diff
@@ -0,0 +1,235 @@
...
+            integer such that x is subsampled by 2**k along the spatial variables.
+        Returns
+        -------
+        res : tensor_like
+            tensor such that its fourier transform is the Fourier
+            transform of a subsampled version of x, i.e. in
+            FFT^{-1}(res)[u1, u2] = FFT^{-1}(x)[u1 * (2**k), u2 * (2**k)]
+    """
+    def __call__(self, input, k):
+        out = input.new(input.size(0), input.size(1), input.size(2) // k, input.size(3) // k, 2)
```
eickenberg: All this is still torch code, we agree?

edouardoyallon:
I'm trying to find a pure - simple np padding to match torch... and then we're good!

Feel free to comment

PR #424, opened 2019-08-19
ENH - attempt to initiate a 3D frontend

### LINKED ISSUE #423: Frontend/backend for 3D is missing
- [x] numpy
- [x] tf

edouardoyallon:
Closes #423 

The code requires a lot of refactoring. Simple todo for this PR:

- [x] Removing torch dependency in https://github.com/edouardoyallon/kymatio/blob/attempt_3D/kymatio/scattering3d/utils.py as well as in https://github.com/edouardoyallon/kymatio/blob/attempt_3D/kymatio/scattering3d/core/scattering3d.py .
- [x] Clearly specifying which functions are part of the backend, which are not. For instance, https://github.com/edouardoyallon/kymatio/blob/attempt_3D/kymatio/scattering3d/utils.py#L54-L55 or https://github.com/edouardoyallon/kymatio/blob/attempt_3D/kymatio/scattering3d/utils.py#L59-L83 are beyond utility functions and should be declared in the appropriate file (probably a `backend.py`).
- [x] Removing complex interlacements between simple operations. For instance, https://github.com/edouardoyallon/kymatio/blob/attempt_3D/kymatio/scattering3d/core/scattering3d.py#L123-L159 could be refactored, in order to clearly show what is done in the code.(if I wasn't familiar with 3D scattering, I would have had a hard time to follow)

- [x] As https://github.com/edouardoyallon/kymatio/blob/attempt_3D/kymatio/scattering3d/core/scattering3d.py#L162-L194 is an elementary function(similarly to a complex modulus), shouldn't it be part of the backend?
- [x] https://github.com/edouardoyallon/kymatio/blob/attempt_3D/kymatio/scattering3d/core/scattering3d.py#L228-L243 is pretty analogous to standard scattering (e.g., current 1-2D or future 3D with Gaussian filters), maybe it could be nice to emphasize it.
- [x] The model of https://github.com/edouardoyallon/kymatio/blob/attempt_3D/kymatio/scattering3d/filter_bank.py is slightly different from what is done in other part of the software. While this is not a bad behaviour for future work, this does not follow the current template, and thus adding filters(hopefully some contributors will add some wavelets beyond Gabor) might be complex here.
- [x] **When everything above is done, add first a numpy frontend,...**
- [x] ...and then a tensorflow frontend.
- [x] ~~The documentation should be drastically improved.~~ to improve everywhere
- [x] ~~It seems to me that the current code can not support super large array on GPU. I think this could be however easily handled with an appropriate backend, for instance by unloading from GPU's memory large tensors. Actually, one could think of two backends: one when large memory is available, and one for contexts of limited resources.~~ too much work for now

This time, that's too much work for me! 😉


kymatio/scattering3d/frontend/torch_frontend.py
```diff
@@ -0,0 +1,160 @@
...
+            self.averaging =lambda x,j: self.backend.compute_integrals(self.backend.fft(x, inverse=True)[...,0],
+                                                                       self.integral_powers)
+        elif self.method == 'local':
+            self.averaging = lambda x,j:\
+                self.backend._compute_local_scattering_coefs(x,
+                        self.tensor_gaussian_filter[j+1], self.points)
+        elif self.method == 'standard':
+            self.averaging = lambda x, j:\
+                self.backend._compute_standard_scattering_coefs(x,
+                        self.tensor_gaussian_filter[j], self.J, self.backend.subsample)
```
edouardoyallon: I'm a bit suspicious about this line. Shouldn't it be `self.tensor_gaussian_filter[J]` or a similar instruction? Similarly, I find it is too bad to pass self.J in this one and not in the previous function. I think it is better to pass self.tensor_gaussian_filter and then to select the j-th filter.

edouardoyallon: @MuawizChaudhary

edouardoyallon:
Currently the np backend doesn't work, I've introduced a bug somewhere.. I modified the filter bank part such that it does return a `np.complex64` array instead of a `np.float32` with a final unsqueezed dimension of 2, it is probably linked to this modification!

kymatio/scattering3d/frontend/torch_frontend.py
MuawizChaudhary: It should indeed be the J'th filter and not the j'th filter, if we are keeping in line with the current version of the software: https://github.com/kymatio/kymatio/blob/master/kymatio/scattering3d/scattering3d.py#L163.

I believe that the previous function we want to pass in the lower case j, and not capital J, see: https://github.com/kymatio/kymatio/blob/master/kymatio/scattering3d/scattering3d.py#L180. .
I'll get on correcting the code and making the functions get the appropriate filter passed in.

MuawizChaudhary:
It seems the that the shape of your output from the numpy scattering transform is not correct. I believe that this is due to numpy.concatenate not providing the same behavior as torch.stack. I think the fix will be simple, I just need to change the numpy backend aggregate method, let me commit soon.

edouardoyallon:
@MuawizChaudhary can't fix a bug on the 3D backend of tf, would be great if someone had some times to watch this out! @janden @eickenberg @louity

MuawizChaudhary:
> @MuawizChaudhary can't fix a bug on the 3D backend of tf, would be great if someone had some times to watch this out! @janden @eickenberg @louity

The result of the CDGMM when l=1, j_1=0, and m=1, between the input (U_0_c) and the filter (filter[l][j_1][m]), is zero. If we compare the inputs on the pytorch and tensorflow end, we see that their values agree (message me on slack if you want the output), however the result of the CDGMM on the tensorflow end is an array of 0's, while the result on the pytorch end has an array of nonzero numbers. Some of these nonzero numbers are big enough that it should exclude round off error happening. 

This problem does not appear to happen when len(filter[l][j_1]) =1 or 5. Am I forgetting something basic about tensorflow?

edouardoyallon:
I've understood.. There is a bug in tf... gosh..

If you multiply dotwise a np.complex(which is our case) with a complex placeholder, it seems that the operation is not a complex multiplication... we didn't get such bugs for 1D or 2D because the filters are real in the Fourier domain...

So for 3cdgmm, the correct code is:

```
def cdgmm3d(A, B, inplace=False):
    """
    Pointwise multiplication of complex tensors.

    ----------
    A: complex torch tensor
    B: complex torch tensor of the same size as A
    inplace : boolean, optional
        if set True, all the operations are performed inplace
    Returns
    -------
    output : torch tensor of the same size as A containing the result of the 
             elementwise complex multiplication of A with B 
    """
    if B.ndim != 3:
        raise RuntimeError('The dimension of the second input must be 3.')

    #C = A * B
    import numpy as np
    Cr = tf.cast(tf.real(A)*np.real(B)-tf.imag(A)*np.imag(B),tf.complex128)
    Ci = tf.cast(tf.real(A)*np.imag(B)+tf.imag(A)*np.real(B),tf.complex128)

    return Cr+1.0j*Ci
```

PR #432, opened 2019-10-15
DO NOT MERGE ENH Base files

MuawizChaudhary:
In this pull request we add files that will be used for all other PRs.


CONTRIBUTORS.md
```diff
@@ -13,4 +13,8 @@ Edouard Oyallon
 Dylan Simon
 Louis Thiry
 Sergey Zagoruyko
-Jonas1312
\ No newline at end of file
+Jonas1312
```
MuawizChaudhary: TODO: Get full name.

doc/source/userguide.rst
```diff
@@ -192,70 +208,79 @@ Currently, two backends exist:
...
+This backend can be specified via::
+
+    import torch
+    from kymatio import Scattering2D
+    from kymatio.scattering2d.backend.torch_skcuda_backend import backend
+
+    scattering = Scattering2D(J=2, shape=(32, 32), frontend='torch', backend=backend)
+
+
 Benchmarks
```
MuawizChaudhary: TODO: Rerun these timings.

kymatio/frontend/tensorflow_frontend.py
```diff
@@ -0,0 +1,20 @@
+import tensorflow as tf
+from packaging import version
+
+assert version.parse(tf.__version__)>=version.parse("2.0.0a0"), 'Current TensorFlow version is '+str(tf.__version__)+\
```
MuawizChaudhary: Do we want to be supporting kymatio for only the most recent version of tf? Why not issue a warning instead? 

I've also heard that people prefer the older versions of TF. Might impact the usage of this software package.

kymatio/frontend/torch_frontend.py
```diff
@@ -0,0 +1,31 @@
+from abc import ABCMeta, abstractmethod
+import torch
+import torch.nn as nn
+from packaging import version
+
+assert version.parse(torch.__version__)>=version.parse("1.3"), 'Current PyTorch version is ' + str(torch.__version__) +\
```
MuawizChaudhary: Same

doc/source/userguide.rst
edouardoyallon: They shouldn't have changed (except we could add tf and np backend but i'm not sure how it works)

kymatio/frontend/tensorflow_frontend.py
edouardoyallon: yes, we do, trust me 🤓

kymatio/frontend/torch_frontend.py
edouardoyallon: same - I however have a proposal!

Jenkinsfile
```diff
@@ -17,11 +17,12 @@ pipeline {
       }
       steps {
 	sh 'python3 -c "import torch; torch.cuda.current_device()"'
+	sh 'python3 -c "import tensorflow as tf; tf.test.is_gpu_available()"'
 	sh 'python3 -m venv --system-site-packages --without-pip $HOME'
 	sh '''#!/bin/bash -ex
 	  source $HOME/bin/activate
 	  python3 setup.py develop
-	  KYMATIO_BACKEND=$STAGE_NAME python3 -m pytest --cov=kymatio
```
janden: At this point, the old backends have not been removed, so we need the environment variables for the tests to work.

Jenkinsfile
```diff
@@ -41,7 +42,7 @@ pipeline {
...
```
janden: See above.

kymatio/frontend/tensorflow_frontend.py
```diff
@@ -0,0 +1,20 @@
...
+                                                                ' . Please upgrade TensorFlow to 2.0.0a0 at least.'
+
+from abc import ABCMeta, abstractmethod
+
+class ScatteringTensorflow(tf.Module, metaclass=ABCMeta):
```
janden: `ScatteringTensorflow` or `ScatteringTensorFlow`?

kymatio/frontend/torch_frontend.py
```diff
@@ -0,0 +1,31 @@
...
+       super(ScatteringTorch, self).__init__()
+
+   @abstractmethod
+   def build(self):
+       """ Defines elementary routines. This function should always call and create the filters via
+       self.create_and_register_filters() defined below. For instance, via:
+       self.filters = self.create_and_register_filters() """
+
+   @abstractmethod
+   def create_and_register_filters(self):
```
janden: Rename to `create_filters`.

tools/travis_install.sh
```diff
@@ -12,18 +12,19 @@ if [[ $CONDA == "1" ]]; then
...
     source activate test-environment
-    conda install -c pytorch pytorch-cpu
+    conda install pytorch torchvision cpuonly -c pytorch
+    pip install 'tensorflow>=2.0.0a0'
 else
     pip install --upgrade pytest
     pip install pytest-cov
     if [[ "$TRAVIS_PYTHON_VERSION" == "3.5" ]]; then
-        pip install https://download.pytorch.org/whl/cpu/torch-1.0.0-cp35-cp35m-linux_x86_64.whl
+        pip install torch==1.3.0+cpu torchvision==0.4.1+cpu -f https://download.pytorch.org/whl/torch_stable.html
```
janden: Why are we now requiring 1.3?

doc/source/userguide.rst
edouardoyallon: I've open an issue.

kymatio/frontend/tensorflow_frontend.py
edouardoyallon: 1/ The APIs are changing quite quickly, it's pretty hard to be compatible with both(I didn't want to spend time on that) I think we should only support the latest vrsion of torch or tf, it's easier.
2/ We can't add this as a `requirements.txt` element because it is optional.
3/ a really motivated use could add its own backend for a given tf version. We might be willing to do that later but it is really too much work for now (I spend a significant time just to make sure the same version of tf was working for python 3.5, 3.6 and 3.7)

kymatio/frontend/torch_frontend.py
edouardoyallon: cf proposing later to build a backend for a given version of tf(or even python)

tools/travis_install.sh
edouardoyallon: We're updating the code.

requirements.txt
```diff
@@ -1,4 +1,5 @@
-numpy
-scipy
+numpy>=1.7
+scipy>=1.3
 appdirs
-configparser
\ No newline at end of file
+configparser
+packaging
```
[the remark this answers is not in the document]
MuawizChaudhary: @edouardoyallon

kymatio/frontend/numpy_frontend.py
```diff
@@ -0,0 +1,25 @@
...
+
+   @abstractmethod
+   def scattering(self, x):
+       """ This function should call the functional scattering."""
+
+   def __call__(self, x):
+       return self.scattering(x)
+
+   @abstractmethod
+   def loginfo(self):
```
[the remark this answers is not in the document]
MuawizChaudhary: @edouardoyallon prolly knows better, but I believe we added loginfo functions to each of the *D frontends. So the other frontend's should have this function. I have added them in a new PR.

kymatio/frontend/tensorflow_frontend.py
MuawizChaudhary: @edouardoyallon

kymatio/frontend/tensorflow_frontend.py
MuawizChaudhary: I remember Pytorch 1.2 and 1.3 worked for the changes we made (I'll check how earlier versions of Pytorch work with our update). 

My thinking is, if it isn't broken, let the users use it. If we get to a point where we need to use some new Pytorch/Tensorflow functionality, then we have a backend that works for the past versions, and we can simply copy the old backend and update. 

I suppose that comes with the disadvantage that we may have to make sure that our software works for all versions of Pytorch/Tensorflow that we intend to support. Thoughts @janden ?

kymatio/frontend/torch_frontend.py
MuawizChaudhary: I believe in the last mega PR #415 someone said that we should break this function into create filters and register filters. I also believe someone said something about having NumPy and Tensorflow inherit from the same class as they need to create filters, but do not register them.

kymatio/__init__.py
```diff
@@ -12,8 +13,9 @@
             'HarmonicScattering3D'
             ]
 
-from .scattering1d.scattering1d import Scattering1D
-from .scattering2d.scattering2d import Scattering2D
-from .scattering3d.scattering3d import HarmonicScattering3D
+from .scattering1d import Scattering1D
+from .scattering2d import Scattering2D
+
```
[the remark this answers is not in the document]
MuawizChaudhary: Removed. @edouardoyallon ?

kymatio/frontend/torch_frontend.py
janden: Right. If I remember correctly, I wanted to break it up, @edouardoyallon said that this is just a temporary thing right now, since we hopefully won't have to register filters in the future. If that's the case, I'd rather call this `create_filters` right away instead of this unwieldy name.

tools/travis_install.sh
janden: So will it not run on older versions of PyTorch?

edouardoyallon: No: I'd like to not maintain it!

kymatio/frontend/torch_frontend.py
edouardoyallon: Sounds good to me.

kymatio/frontend/tensorflow_frontend.py
edouardoyallon: TensorFlow I guess... I'm not sure what is the standard way of doing it.

Jenkinsfile
edouardoyallon: Right.

Jenkinsfile
edouardoyallon: Agreed.

doc/source/index.rst
```diff
@@ -33,26 +31,31 @@
 Kymatio
 *******
 
-|pic2|  |pic1| |picDL| |piccodecov|
```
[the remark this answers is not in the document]
edouardoyallon: Refactoring of the graphics.. nothing big.

doc/Makefile
```diff
@@ -4,7 +4,7 @@
 # You can set these variables from the command line.
 SPHINXOPTS    =
 SPHINXBUILD   = sphinx-build
-SPHINXPROJ    = scattering
+SPHINXPROJ    = kymatio
```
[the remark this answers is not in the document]
edouardoyallon: Solved.

kymatio/__init__.py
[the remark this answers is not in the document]
edouardoyallon: I don't know what this does mean.

requirements.txt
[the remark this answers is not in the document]
edouardoyallon: `packaging` allows to use `version` in order to compare versions.

kymatio/frontend/numpy_frontend.py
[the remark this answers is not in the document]
janden: Ok. @edouardoyallon can you explain why it's needed?

kymatio/frontend/numpy_frontend.py
```diff
@@ -0,0 +1,25 @@
+from abc import ABCMeta, abstractmethod
+
+class ScatteringNumpy(object, metaclass=ABCMeta):
```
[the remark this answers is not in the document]
edouardoyallon: can't we do it in a next PR? that doesn't break anything.

The answer to `numpy` should be the same as for `tensorflow`.

kymatio/frontend/numpy_frontend.py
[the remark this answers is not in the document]
edouardoyallon: loginfo is a message that is displayed in the console. It's nice for debugging purposes, and more messages should be actually logged.

kymatio/frontend/tensorflow_frontend.py
janden: Ok, for (1), I agree. It seems that TF1 and TF2 are quite different beasts. Now with torch, it's not clear to me that we can't support earlier versions (at least down to 1.0) without much effort. If it requires a bunch of extra work, sure, let's restrict it to 1.3, but if not, I don't see why we can't let users use an earlier version. We have to bear in mind that users have their constraints too, let's not burden them with more if it's not needed.

For (2), we used to have a bunch of different `requirements.txt` for different purposes (like `requirements_skcuda.txt`). Why not stick to that?

For (3) sure, why not?

kymatio/frontend/torch_frontend.py
janden: Ah yes, sure. These are frontends, though. It should be easier to have those be more compatible (since they're doing less heavy lifting).

kymatio/frontend/tensorflow_frontend.py
edouardoyallon: I think deep learning pipelines change too fastly. The problem is not to use new functionality, yet the syntax is simply different. This was the case with `tensorflow`. Also for debugging purposes, I think it's better to constraint the version we're using. PyTorch got recently updated, why not setting this as a breaking point?

README.md
```diff
@@ -14,7 +14,7 @@ Scattering transforms are translation-invariant signal representations implement
 Use Kymatio if you need a library that:
 * supports 1-D, 2-D, and 3-D wavelets,
 * integrates wavelet scattering in a deep learning architecture, and
-* runs seamlessly on CPU and GPU hardware.
+* runs seamlessly on CPU and GPU hardware, with many deep learning API.
```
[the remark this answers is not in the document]
edouardoyallon: Right.

README.md
```diff
@@ -27,7 +27,7 @@ The resort to PyTorch tensors as inputs to Kymatio allows the programmer to back
 
 Each of these algorithms is written in a high-level imperative paradigm, making it portable to any Python library for array operations as long as it enables complex-valued linear algebra and a fast Fourier transform (FFT).
 
-Currently, there are two available backends, PyTorch (CPU and GPU) and scikit-cuda (GPU only).
+Currently, there are four available frontend-backends, PyTorch (CPU and GPU), PyTorch+SciKit-cuda (GPU only), TensorFlow (CPU and GPU) and NumPy (CPU).
```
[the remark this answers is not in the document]
edouardoyallon: Let's prepare the documentation... I find it safer, as we can already see what's wrong with it. We should merge them in the coming weeks hopefully.

README.md
```diff
@@ -72,40 +67,81 @@ pip install kymatio
 Linux and macOS are the two officially supported operating systems.
 
 
-### GPU acceleration
+### Frontend
 
+#### NumPy
 
-To run Kymatio on a graphics processing unit (GPU), you can either use the PyTorch-style `cuda()` method to move your object to GPU. For extra speed, install the CUDA library and install the `scikit-cuda` dependency by running the following pip command:
+To explicitely call the `numpy` frontend, simply do for instance:
```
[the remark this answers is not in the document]
edouardoyallon: Right.

README.md
```diff
@@ -72,40 +67,81 @@ pip install kymatio
...
 
 ```
-pip install scikit-cuda cupy
+import numpy as np
+from kymatio import Scattering2D
+scattering = Scattering2D(J=2, shape=(32, 32), frontend='numpy')
```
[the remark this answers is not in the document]
edouardoyallon: Right.

kymatio/frontend/tensorflow_frontend.py
edouardoyallon: for (1): right - I'd prefer to set the flag at the latest torch version but why not. I have no idea which version breaks or not, but if one decided to include more version, then each of them should be included in the unit tests.
for (2): it is not required to have cuda, so i don't understand why this should be a requirement

kymatio/frontend/torch_frontend.py
edouardoyallon: no. I had some bugs in the frontend because APIs have changed.

kymatio/frontend/tensorflow_frontend.py
janden: (1) I don't mind setting up the testing for different versions. Better to be safe than sorry. Also, from what I've seen, the PyTorch people do seem quite careful about maintaining backwards compatbility (at least more so than TF), so I don't think it's too big of a risk.

(2) I'm not saying it should be a requirement. I'm saying that we should stick the packages in a file (so that we can install them with `pip install -r`), that way if we add packages for that backend later on, we're just updating that file, not the docs (where it may occur in several places). It also makes updating the tests easier, since they're just using that file.

edouardoyallon: for (2), see with @eickenberg

kymatio/frontend/torch_frontend.py
janden: All right then. In TF or torch?

edouardoyallon: TF - not torch(or not yet) I'm ok to give a short to torch, but TF was really a pain..

README.md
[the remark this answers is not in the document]
janden: Ok. Works for me.

doc/source/index.rst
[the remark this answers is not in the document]
janden: Well codecov seems to be gone now?

doc/source/installation.rst
```diff
@@ -33,9 +30,10 @@ Finally, the package is installed by running::
 Optimizing GPU acceleration
 ===========================
 
-To improve performance with an optimized CUDA implementation, you may install optional CUDA packages by running::
+To improve performance on PyTorch with an optimized CUDA implementation, you may install optional CUDA packages by
+running::
 
-    pip install -r requirements_optional_cuda.txt
+    pip install scikit-cuda cupy
```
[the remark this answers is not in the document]
janden: Again, not _the_ `requirements.txt` but _a_ `requirements.txt` (like `requirements_skcuda.txt`).

kymatio/__init__.py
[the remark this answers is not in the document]
janden: Why `.scattering2d`, not `.scattering2d.scattering2d`?

kymatio/frontend/numpy_frontend.py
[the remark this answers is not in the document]
janden: > can't we do it in a next PR? that doesn't break anything.

It's quite a substantial difference. Also didn't you said the ABCs break our generic `Scattering*d`? I'd rather get it (somewhat) right the first time.

kymatio/frontend/numpy_frontend.py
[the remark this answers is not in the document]
janden: Can't we use `logging` for that?

kymatio/frontend/torch_frontend.py
janden: :+1:

requirements.txt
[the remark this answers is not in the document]
janden: Where do we use that?

doc/source/index.rst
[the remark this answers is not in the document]
edouardoyallon: i think when there are too much wrong commit/PR, it stops working..

doc/source/installation.rst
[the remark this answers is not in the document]
edouardoyallon: @eickenberg

kymatio/__init__.py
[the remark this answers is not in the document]
edouardoyallon: oh - no idea. Feel free to reverse!

requirements.txt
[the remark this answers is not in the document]
edouardoyallon: when it does compare the version in the frontends, eg

`kymatio/frontend/torch_frontend.py` line 4.

kymatio/frontend/numpy_frontend.py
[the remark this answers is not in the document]
edouardoyallon: so this method is called by logging and will be displayed. loginfo is just a method that should be abstracted and will be used by the generic module

doc/source/index.rst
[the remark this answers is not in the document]
janden: Right. Let's figure out how to fix it before putting it back in.

requirements.txt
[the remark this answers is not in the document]
janden: Ah I see. Ok.

What about NumPy and SciPy. Were there issues with earlier versions?

kymatio/frontend/tensorflow_frontend.py
janden: It's normally written TensorFlow, so I suggest we stick to that.

edouardoyallon: Alright.

requirements.txt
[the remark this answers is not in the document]
edouardoyallon: I didn't try them but contrary to tf or torch, those frameworks raise warnings like 2 or 3 versions before to even thinking to delete or modify the standard behavior of a feature..(true story, took 3y for one of my code to get obsolete)...

kymatio/frontend/numpy_frontend.py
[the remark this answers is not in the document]
edouardoyallon: Precisely, `abc` is not well implemented. Thus, instead to break when a function is not implemented, it's imply not even run. The reason is that the constructor is not called: we dynamically change the type without calling again the constructor. Consequently, I'm in favor to maintain it and remove it smoothly in a future PR.

kymatio/__init__.py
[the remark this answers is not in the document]
MuawizChaudhary: What's the final conclusion here?

kymatio/frontend/numpy_frontend.py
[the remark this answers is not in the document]
MuawizChaudhary: Final conclusions? Or still a topic to be discussed?

kymatio/frontend/numpy_frontend.py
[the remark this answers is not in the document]
MuawizChaudhary: Conclusion?

edouardoyallon:
I"ve updated the target branch to `kymatio-v2`.

requirements.txt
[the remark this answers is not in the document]
janden: So they're very stable? If so, shouldn't we lift the version restrictions on them?

README.md
```diff
@@ -72,40 +67,81 @@ pip install kymatio
...
 ```
-os.environ["KYMATIO_BACKEND"] = "skcuda"
+import torch
+from kymatio import Scattering2D
+scattering = Scattering2D(J, shape=(M, N), L=L, frontend='torch')
 ```
 
+#### TensorFlow
+
+After installing the latest version of `tensorflow`, you can call Scattering2d as a `tf.Module` via for instance:
```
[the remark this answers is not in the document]
janden: Yes.

doc/source/userguide.rst
```diff
@@ -192,70 +208,79 @@ Currently, two backends exist:
   improvement is currently small in 1D and 3D, but work is underway to further
   optimize this backend.
 
+This backend can be specified via::
+
+    import torch
+    from kymatio import Scattering2D
+    from kymatio.scattering2d.backend.torch_skcuda_backend import backend
+
+    scattering = Scattering2D(J=2, shape=(32, 32), frontend='torch', backend=backend)
```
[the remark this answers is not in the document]
janden: Yes.

requirements.txt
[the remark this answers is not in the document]
edouardoyallon: So, re-thinking that: what about a simple warning instead of an error, would you be more encline? I'm not really willing to try each possible configurations, but rather to specify a non-outdated, updated and tested code..

kymatio/frontend/tensorflow_frontend.py
MuawizChaudhary: How important is it that that this be written TensorFlow? 

https://github.com/kymatio/kymatio/blob/kymatio-dev-frontend/kymatio/scattering1d/frontend/__init__.py We get errors thrown over here if we rename to TensorFlow, as `frontend  = "tensorflow"`, `frontend.capitalize() = "Tensorflow"`, and if we had a statement which renames `frontend = "tensorflow"` to `frontend = "tensorFlow"`, then `frontend.capitalize = "Tensorflow"`.

One thing we could do is change frontend name to have proper capitalization. So, in tensorflow frontend, `frontend = "TensorFlow"`, and in numpy frontend, `frontend = "NumPy"`.

requirements.txt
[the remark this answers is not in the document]
janden: I think that's good. We should say what versions it's tested for and warn if older versions are used. On the other hand, if we see that it breaks for certain versions, we should put in a minimum version requirement.

PR #440, opened 2019-10-21
DOC Kymatio documentation 0.2.0 [1]

MuawizChaudhary:
As discussed in #432 , it's better to split the pull request into smaller pull request. This pull request changes the documentation in anticipation of 0.2.0. 

Questions: should the readme be included in this pull request or some other separate one? 

Do we want to create a new branch that is just a copy of master, merge all 0.2.0 changes into that one, then merge into master?

Also see #441 .

Recommended to be merged 1st.

PR #433, opened 2019-10-15
DO NOT MERGE Pytorch 2D Back/Frontend

MuawizChaudhary:
Here are the files for the 2d pytorch back and frontend.

Recommended to be merged 3rd/4th/5th.

PR #441, opened 2019-10-21
ENH Kymatio Frontends [2]

MuawizChaudhary:
As discussed in #432 , it's better to split the pull request into smaller pull request. This pull request adds the new frontend classes, which will be used in 0.2.0. 

Also see #440 .

Recommended to merge 2nd.

### DIFF
--- a/kymatio/__init__.py
+++ b/kymatio/__init__.py
@@ -1,8 +1,7 @@
 # Make sure that DeprecationWarning within this package always gets printed
 ### Snippet copied from sklearn.__init__
-import re
 import warnings
-
+import re
 warnings.filterwarnings('always', category=DeprecationWarning,
                         module=r'^{0}.*'.format(re.escape(__name__)))
 ### End Snippet
--- a/kymatio/__init__.py
+++ b/kymatio/__init__.py
@@ -1,7 +1,8 @@
 # Make sure that DeprecationWarning within this package always gets printed
 ### Snippet copied from sklearn.__init__
-import warnings
 import re
+import warnings
+
 warnings.filterwarnings('always', category=DeprecationWarning,
                         module=r'^{0}.*'.format(re.escape(__name__)))
 ### End Snippet
@@ -12,8 +13,9 @@ __all__ = [
             'HarmonicScattering3D'
             ]
 
-from .scattering1d.scattering1d import Scattering1D
-from .scattering2d.scattering2d import Scattering2D
-from .scattering3d.scattering3d import HarmonicScattering3D
+from .scattering1d import Scattering1D
+from .scattering2d import Scattering2D
+
+from .scattering3d import HarmonicScattering3D
 
 from .version import version as __version__
--- a/kymatio/__init__.py
+++ b/kymatio/__init__.py
@@ -18,4 +18,3 @@ from .scattering2d.scattering2d import Scattering2D
 from .scattering3d.scattering3d import HarmonicScattering3D
 
 from .version import version as __version__
-
--- a/kymatio/__init__.py
+++ b/kymatio/__init__.py
@@ -13,9 +13,9 @@ __all__ = [
             'HarmonicScattering3D'
             ]
 
-from .scattering1d import Scattering1D
-from .scattering2d import Scattering2D
-from .scattering3d import HarmonicScattering3D
+from .scattering1d.scattering1d import Scattering1D
+from .scattering2d.scattering2d import Scattering2D
+from .scattering3d.scattering3d import HarmonicScattering3D
 
 from .version import version as __version__
 
--- a/kymatio/__init__.py
+++ b/kymatio/__init__.py
@@ -15,7 +15,7 @@ __all__ = [
 
 from .scattering1d import Scattering1D
 from .scattering2d import Scattering2D
-
 from .scattering3d import HarmonicScattering3D
 
 from .version import version as __version__
+
--- a/kymatio/frontend/base_frontend.py
+++ b/kymatio/frontend/base_frontend.py
@@ -11,6 +11,15 @@ class ScatteringBase():
         self.filters = self.create_filters() """
         raise NotImplementedError
     
+    def register_backend(self, string):
+        """ This function should register the backend to be used"""
+        if not self.backend:
+            backend = __import__(string, globals(), locals(), ['backend'], 0)
+            self.backend = backend.backend
+        elif not self.backend.name.startswith(self.name):
+            raise RuntimeError('This backend is not supported.')
+ 
+
     def create_filters(self):
         """ This function should run a filterbank function that
         will create the filters as numpy array, and then, it should
--- a/kymatio/frontend/base_frontend.py
+++ b/kymatio/frontend/base_frontend.py
@@ -13,7 +13,7 @@ class ScatteringBase():
         raise NotImplementedError
     
     def _instantiate_backend(self, import_string):
-        """ This function should set the backend to be used if not already
+        """ This function should instantiate the backend to be used if not already
         specified"""
         if isinstance(self.backend, str):
             self.backend = importlib.import_module(import_string + self.backend \
--- a/kymatio/frontend/base_frontend.py
+++ b/kymatio/frontend/base_frontend.py
@@ -12,11 +12,11 @@ class ScatteringBase():
         self.filters = self.create_filters() """
         raise NotImplementedError
     
-    def _instantiate_backend(self, string):
+    def _instantiate_backend(self, import_string):
         """ This function should set the backend to be used if not already
         specified"""
         if isinstance(self.backend, str):
-            self.backend = importlib.import_module(string + self.backend \
+            self.backend = importlib.import_module(import_string + self.backend \
                     + "_backend", 'backend')
         elif not self.backend.name.startswith(self.frontend_name):
             raise RuntimeError('This backend is not supported.')
--- a/kymatio/frontend/base_frontend.py
+++ b/kymatio/frontend/base_frontend.py
@@ -18,8 +18,9 @@ class ScatteringBase():
         if isinstance(self.backend, str):
             self.backend = importlib.import_module(string + self.backend \
                     + "_backend", 'backend')
-        elif not self.backend.name.startswith(self.backend_name):
+        elif not self.backend.name.startswith(self.frontend_name):
             raise RuntimeError('This backend is not supported.')
+
  
 
     def create_filters(self):
--- a/kymatio/frontend/base_frontend.py
+++ b/kymatio/frontend/base_frontend.py
@@ -16,10 +16,6 @@ class ScatteringBase():
         save those arrays as module's buffers. """
         raise NotImplementedError
 
-    def scattering(self, x):
-        """ This function should call the functional scattering."""
-        raise NotImplementedError
-    
     def loginfo(self):
         """ Returns the logging message when the frontend is deployed."""
         raise NotImplementedError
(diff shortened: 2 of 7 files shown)


edouardoyallon:
@MuawizChaudhary maybe you can incorporate your frontend here? Also, @janden wants something without the "ABC", yet instead, raise NotImplementedError. The only thing it will change is that:

```
class ScatteringTorch(nn.Module, metaclass=ABCMeta):
```

becomes

```
class ScatteringTorch(nn.Module, ScatteringGeneral):
```

We remove the `import` relative to `abc` and for instance:

```
   @abstractmethod
   def scattering(self, x):
       """ This function should compute the scattering transform."""
```

will now be

```
   def scattering(self, x):
       """ This function should compute the scattering transform."""
      raise NotImplementedError
```

That's all.

edouardoyallon:
Also there is not unittest here, that's *fine*.

edouardoyallon:
Hi,

What's the purpose of "super frontend"? beyond that, I'm ok with merging. @eickenberg @janden ??

MuawizChaudhary:
> Hi,
> 
> What's the purpose of "super frontend"? beyond that, I'm ok with merging. @eickenberg @janden ??

Did it based on Dr. Lostanlen's [suggestion](https://github.com/kymatio/kymatio/pull/447#issuecomment-546919256): 
> `General` is not a good name for the abstraction.
> Consider replacing by `Super` or `Interface` or `Parent` or `Base`
> 
> https://en.wikipedia.org/wiki/Inheritance_(object-oriented_programming)
Think interface would be a better name?

requirements.txt
```diff
@@ -1,4 +1,5 @@
-numpy
-scipy
+numpy>=1.7
```
janden: As mentioned in #440, let's drop these requirements for now.

kymatio/frontend/torch_frontend.py
```diff
@@ -0,0 +1,35 @@
...
+        """ This function should call the functional scattering."""
+        raise NotImplementedError
+
+    def forward(self, x):
+        return self.scattering(x)
+    
+    def loginfo(self):
+        """ Returns the logging message when the frontend is deployed."""
+        raise NotImplementedError
+
```
janden: Remove.

MuawizChaudhary:
> It's not clear to me what the intended structure is here. What's going to inherit from `ScatteringSuper`? Why are certain methods (like `build()`) defined both there and in `ScatteringNumPy`, for example?

Yeah, so my plan was to have a specific ScatteringSuper implementation for each *d front end, as currently the Torch, NumPy, and TensorFlow ends share alot of code (I believe you said this in one of your comments too). Then there would be a ScatteringTorch*d, ScatteringNumPy*d, and ScatteringTensorFlow*d frontend which would subclass the ScatteringSuper*d implementation and the appropriate class in the kymatio base frontend folder. 

Is this fine or would you prefer some other system? Is Super fine or do we want to name it parent or interface?

The build() should not be there. 

You have alot of "remove", but I believe Dr. Oyallon said you wanted to remove the abc structure and have each method raise not implemented errors instead. Correct me if I am mistaken.

kymatio/__init__.py
```diff
@@ -12,8 +13,9 @@
             'HarmonicScattering3D'
             ]
 
-from .scattering1d.scattering1d import Scattering1D
-from .scattering2d.scattering2d import Scattering2D
-from .scattering3d.scattering3d import HarmonicScattering3D
+from .scattering1d import Scattering1D
+from .scattering2d import Scattering2D
+from .scattering3d import HarmonicScattering3D
```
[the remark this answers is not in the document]
MuawizChaudhary: @edouardoyallon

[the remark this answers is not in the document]
edouardoyallon: revert: it allows avoid declaring an init file

edouardoyallon:
> On the subject of removing: @edouardoyallon you were the one that added in the loginfo, do you have any arguments for keeping it in? 

 Actually, with inheritance, the idea is to factorize the code. Thus, the "super" frontend should be inherited by torch, np and tf frontend. Then, redundant functions should be avoided: `loginfo` should be defined in the "super" frontend.

I'm ok to remove the loginfo function from this PR and to add it later. It's pretty useful for debugging and tracking purposes.

kymatio/frontend/torch_frontend.py
```diff
@@ -0,0 +1,35 @@
...
+from packaging import version
+
+assert version.parse(torch.__version__)>=version.parse("1.3"), 'Current PyTorch version is ' + str(torch.__version__) +\
+                                                    ' . Please upgrade PyTorch to 1.3 at least.'
+
+class ScatteringTorch(nn.Module):
+    def __init__(self):
+        super(ScatteringTorch, self).__init__()
+    
+    def build(self):
```
MuawizChaudhary: Remove build, define register_backend.

kymatio/frontend/torch_frontend.py
```diff
@@ -0,0 +1,35 @@
...
+        """ Defines elementary routines. 
+        This function should always call and create the filters via
+        self.create_and_register_filters() defined below. For instance, via:
+        self.filters = self.create_and_register_filters() """
+        raise NotImplementedError
+    
+    def create_filters(self):
```
MuawizChaudhary: Define register_filter.

kymatio/frontend/base_frontend.py
```diff
@@ -0,0 +1,26 @@
+
+class ScatteringBase():
```
MuawizChaudhary: Check all current Scattering*D files and make sure there aren't other functions we need to define.

MuawizChaudhary:
> > On the subject of removing: @edouardoyallon you were the one that added in the loginfo, do you have any arguments for keeping it in?
> 
> Actually, with inheritance, the idea is to factorize the code. Thus, the "super" frontend should be inherited by torch, np and tf frontend. Then, redundant functions should be avoided: `loginfo` should be defined in the "super" frontend.
> 

Perhaps I should slow down a bit on making some of these changes. 

In my current implementation (#433) the ScatteringND\*, where * in {torch, numpy, tensorflow}, inherits from both Scattering* and an implementation of ScatteringSuper. Is there some other better way of doing it?

kymatio/frontend/base_frontend.py
```diff
@@ -0,0 +1,26 @@
...
+    def __init__(self):
+        super(ScatteringBase, self).__init__()
+    
+    def build(self):
+        """ Defines elementary routines.
```
MuawizChaudhary: For the doc strings in the front ends section do we want to follow the [Numpy](https://numpydoc.readthedocs.io/en/latest/format.html) doc strings convention?

MuawizChaudhary:
Thinking about it, every frontend registers the backend, across all the *d. Perhaps we should have those implemented in the corresponding torch, numpy, tensorflow frontend defined here?

I'll commit some stuff related to that soon, if they aren't to y'all's liking, we can simply revert those commits.

edouardoyallon:
> Thinking about it, every frontend registers the backend, across all the *d. Perhaps we should have those implemented in the corresponding torch, numpy, tensorflow frontend defined here?
> 
> I'll commit some stuff related to that soon, if they aren't to y'all's liking, we can simply revert those commits.

Tbh I'm gonna pretty easy-going on this one! A later refactoring would be possible and will be plausible: it shouldn't affect much the future user of this new pkg!

requirements.txt
edouardoyallon: Feel free @MuawizChaudhary to open an issue!

kymatio/frontend/base_frontend.py
```diff
@@ -0,0 +1,26 @@
...
+        save those arrays as module's buffers. """
+        raise NotImplementedError
+
+    def scattering(self, x):
+        """ This function should call the functional scattering."""
+        raise NotImplementedError
+    
+    def loginfo(self):
+        """ Returns the logging message when the frontend is deployed."""
+        raise NotImplementedError
```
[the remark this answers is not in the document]
edouardoyallon: If Joakim said to rmeove, I'm ok with it. I'll add later the log messages.

janden:
> To elucidate with an example on the 1d branch: I would first implement ScatteringSuper1D, which inherits from ScatteringSuper. Then I would implement ScatteringTorch1D by inheriting from ScatteringSuper1D and from ScatteringTorch, and implementing the required methods.

Ok that could work. What would each of these classes implement (that isn't implemented already in one of the parent classes)? I'm just trying to get a sense of what the division of responsibility is here.

> I'll change the name from ScatteringSuper to ScatteringBase, unless you believe strongly in just Scattering.

It might be good to reserve the bare `Scattering` to the user-facing classes (like `Scattering1D`), so `ScatteringBase` might be better here.

> Thinking about it, every frontend registers the backend, across all the *d. Perhaps we should have those implemented in the corresponding torch, numpy, tensorflow frontend defined here?

I don't understand. What does “registering” mean here?

kymatio/frontend/base_frontend.py
janden: We should follow whatever convention is used by the other user-facing functions. I believe it's the same as NumPy, but I'm not 100% certain.

kymatio/frontend/base_frontend.py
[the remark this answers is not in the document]
janden: Yes I'm not sure why we need to reimplement the logging package [that comes standard with Python](https://docs.python.org/3/library/logging.html).

MuawizChaudhary:
> > To elucidate with an example on the 1d branch: I would first implement ScatteringSuper1D, which inherits from ScatteringSuper. Then I would implement ScatteringTorch1D by inheriting from ScatteringSuper1D and from ScatteringTorch, and implementing the required methods.
> 
> Ok that could work. What would each of these classes implement (that isn't implemented already in one of the parent classes)? I'm just trying to get a sense of what the division of responsibility is here.
> 
ScatteringSuper1D would implement the functions in ScatteringSuper. ScatteringTorch1D would implement the functions in ScatteringTorch. Let me know if you need more detail.

> > I'll change the name from ScatteringSuper to ScatteringBase, unless you believe strongly in just Scattering.
> 
> It might be good to reserve the bare `Scattering` to the user-facing classes (like `Scattering1D`), so `ScatteringBase` might be better here.
> 

I believe I've made this change.

janden:
> I believe I've made this change.

Right. I didn't see. Sorry about that.

kymatio/frontend/base_frontend.py
```diff
@@ -0,0 +1,32 @@
...
+    def build(self):
+        """ Defines elementary routines. 
+
+        This function should always call and create the filters via
+        self.create_filters() defined below. For instance, via:
+        self.filters = self.create_filters() """
+        raise NotImplementedError
+    
+    def register_backend(self, string):
+        """ This function should register the backend to be used"""
```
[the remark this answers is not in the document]
MuawizChaudhary: As discussed on the slack, we'll change name from `register_backend` to `set_backend`

kymatio/frontend/base_frontend.py
```diff
@@ -0,0 +1,31 @@
...
+        """ This function should instantiate the backend to be used if not already
+        specified"""
+        if isinstance(self.backend, str):
+            self.backend = importlib.import_module(import_string + self.backend \
+                    + "_backend", 'backend')
+        elif not self.backend.name.startswith(self.frontend_name):
+            raise RuntimeError('This backend is not supported.')
+
+ 
+
```
MuawizChaudhary: Is it me or is this a lot of space?

janden: It is.

janden:
Ok. We're good here, as far as I'm concerned.

@edouardoyallon, @eickenberg?

edouardoyallon:
> Ok. We're good here, as far as I'm concerned.
> 
> @edouardoyallon, @eickenberg?

Please merge.

PR #459, opened 2019-11-28
ENH NumPy 2D Back/Frontend [3b]

MuawizChaudhary:
This is the PR for the NumPy back/frontends. This is up next for review, then TF. 

I tried to make some fixes, however the tests do not run. The current error I am getting is `E       TypeError: __init__() missing 2 required positional arguments: 'J' and 'shape'`, and this is occurring in the numpy_frontend abstract class. I'm not sure why its giving this error, when the torch frontend is initalized in the same way and there is no difference between the two. Perhaps this difference is due to the numpy abstract frontend subclassing off of object?

### DIFF
--- a/kymatio/scattering2d/backend/numpy_backend.py
+++ b/kymatio/scattering2d/backend/numpy_backend.py
@@ -1,0 +1,194 @@
+# Authors: Edouard Oyallon, Sergey Zagoruyko
+
+import numpy as np
+from collections import namedtuple
+
+BACKEND_NAME = 'numpy'
+
+class Pad(object):
+    def __init__(self, pad_size, input_size, pre_pad=False):
+        """
+            Padding which allows to simultaneously pad in a reflection fashion
+            and map to complex.
+
+            Parameters
+            ----------
+            pad_size : list of 4 integers
+                size of padding to apply.
+            input_size : list of 2 integers
+                size of the original signal
+            pre_pad : boolean
+                if set to true, then there is no padding, one simply converts
+                from real to complex.
+
+        """
+        self.pre_pad = pre_pad
+        self.pad_size = pad_size
+
+    def __call__(self, x):
+        if self.pre_pad:
+            return x
+        else:
+            np_pad = ((self.pad_size[0], self.pad_size[1]), (self.pad_size[2], self.pad_size[3]))
+            batch_shape = x.shape[:-2]
+            signal_shape = x.shape[-2:]
+            x = x.reshape((-1, 1) + signal_shape)
+            output = np.pad(x, ((0,0), (0,0), np_pad[0], np_pad[1]), mode='reflect')
+            output = output.reshape(batch_shape + output.shape[-2:])
+            return output
+
+
+def unpad(in_):
+    """
+        Slices the input tensor at indices between 1::-1
+
+        Parameters
+        ----------
+        in_ : tensor_like
+            input tensor
+
+        Returns
+        -------
+        in_[..., 1:-1, 1:-1]
+
+    """
+    return np.expand_dims(in_[..., 1:-1, 1:-1], -3)
+
+class SubsampleFourier(object):
+    """ Subsampling of a 2D image performed in the Fourier domain.
+
+        Subsampling in the spatial domain amounts to periodization
+        in the Fourier domain, hence the formula.
+
+        Parameters
+        ----------
+        x : tensor_like
+            input tensor with at least 5 dimensions, the last being the real
+             and imaginary parts.
+            Ideally, the last dimension should be a power of 2 to avoid errors.
+        k : int
+            integer such that x is subsampled by 2**k along the spatial variables.
+
+        Returns
+        -------
+        res : tensor_like
+            tensor such that its fourier transform is the Fourier
+            transform of a subsampled version of x, i.e. in
+            FFT^{-1}(res)[u1, u2] = FFT^{-1}(x)[u1 * (2**k), u2 * (2**k)]
+
+    """
+    def __call__(self, x, k):
+        batch_shape = x.shape[:-2]
+        signal_shape = x.shape[-2:]
+
+        x = x.reshape((-1,) + signal_shape)
+        y = x.reshape(-1, k, x.shape[1] // k, k, x.shape[2] // k)
+
+        out = np.zeros_like(y, dtype=x.dtype)
+        out = y.mean(axis=(1, 3))
+
+        return out
+
+
+class Modulus(object):
+    """
+        This class implements a modulus transform for complex numbers.
+
+        Usage
+        -----
+        modulus = Modulus()
+        x_mod = modulus(x)
+
+        Parameters
+        ---------
+        x: input complex tensor.
+
+        Returns
+        -------
+        output: a real tensor equal to the modulus of x.
+
+    """
+    def __call__(self, x):
+        norm = np.abs(x)
+        return norm
+
+
+
+
+def fft(x, direction='C2C', inverse=False):
+    """
+        Interface with torch FFT routines for 2D signals.
+
+        Example
+        -------
+        x = torch.randn(128, 32, 32, 2)
+        x_fft = fft(x, inverse=True)
+
+        Parameters
+        ----------
+        input : tensor
+            complex input for the FFT
+        direction : string
+            'C2R' for complex to real, 'C2C' for complex to complex
+        inverse : bool
+            True for computing the inverse FFT.
+            NB : if direction is equal to 'C2R', then an error is raised.
+
+    """
+    if direction == 'C2R':
+        if not inverse:
+            raise RuntimeError('C2R mode can only be done with an inverse FFT.')
+
+    if direction == 'C2R':
+        output = np.real(np.fft.ifft2(x))*x.shape[-1]*x.shape[-2]
+    elif direction == 'C2C':
+        if inverse:
+            output = np.fft.ifft2(x)*x.shape[-1]*x.shape[-2]
+        else:
+            output = np.fft.fft2(x)
+
+    return output
+
+
+
+
+def cdgmm(A, B, inplace=False):
+    """
+        Complex pointwise multiplication between (batched) tensor A and tensor B.
+
+        Parameters
+        ----------
+        A : tensor
+            A is a complex tensor of size (B, C, M, N, 2)
+        B : tensor
+            B is a complex tensor of size (M, N) or real tensor of (M, N)
+        inplace : boolean, optional
+            if set to True, all the operations are performed inplace
+
+        Returns
+        -------
+        C : tensor
+            output tensor of size (B, C, M, N, 2) such that:
+            C[b, c, m, n, :] = A[b, c, m, n, :] * B[m, n, :]
+
+    """
+    if B.ndim != 2:
+        raise RuntimeError('The dimension of the second input must be 2.')
+
+    if inplace:
+        return np.multiply(A, B, out=A)
+    else:
+        return A * B
+
+def empty_like(x, shape):
+    return np.empty(shape, x.dtype)
+
+backend = namedtuple('backend', ['name', 'cdgmm', 'modulus', 'subsample_fourier', 'fft', 'Pad', 'unpad', 'empty_like'])
+backend.name = 'numpy'
+backend.cdgmm = cdgmm
+backend.modulus = Modulus()
+backend.subsample_fourier = SubsampleFourier()
+backend.fft = fft
+backend.Pad = Pad
+backend.unpad = unpad
+backend.empty_like = empty_like
--- a/kymatio/scattering2d/backend/numpy_backend.py
+++ b/kymatio/scattering2d/backend/numpy_backend.py
@@ -79,13 +79,8 @@ class SubsampleFourier(object):
 
     """
     def __call__(self, x, k):
-        batch_shape = x.shape[:-2]
-        signal_shape = x.shape[-2:]
-
-        x = x.reshape((-1,) + signal_shape)
         y = x.reshape(-1, k, x.shape[1] // k, k, x.shape[2] // k)
 
-        out = np.zeros_like(y, dtype=x.dtype)
         out = y.mean(axis=(1, 3))
 
         return out
--- a/kymatio/scattering2d/backend/numpy_backend.py
+++ b/kymatio/scattering2d/backend/numpy_backend.py
@@ -3,7 +3,9 @@
 import numpy as np
 from collections import namedtuple
 
+
 BACKEND_NAME = 'numpy'
+
 
 class Pad(object):
     def __init__(self, pad_size, input_size, pre_pad=False):
@@ -53,6 +55,7 @@ def unpad(in_):
 
     """
     return np.expand_dims(in_[..., 1:-1, 1:-1], -3)
+
 
 class SubsampleFourier(object):
     """ Subsampling of a 2D image performed in the Fourier domain.
@@ -113,8 +116,6 @@ class Modulus(object):
         return norm
 
 
-
-
 def fft(x, direction='C2C', inverse=False):
     """
         Interface with torch FFT routines for 2D signals.
@@ -140,16 +141,14 @@ def fft(x, direction='C2C', inverse=False):
             raise RuntimeError('C2R mode can only be done with an inverse FFT.')
 
     if direction == 'C2R':
-        output = np.real(np.fft.ifft2(x))*x.shape[-1]*x.shape[-2]
+        output = np.real(np.fft.ifft2(x)) * x.shape[-1] * x.shape[-2]
     elif direction == 'C2C':
         if inverse:
-            output = np.fft.ifft2(x)*x.shape[-1]*x.shape[-2]
+            output = np.fft.ifft2(x) * x.shape[-1] * x.shape[-2]
         else:
             output = np.fft.fft2(x)
 
     return output
-
-
 
 
 def cdgmm(A, B, inplace=False):
@@ -180,8 +179,10 @@ def cdgmm(A, B, inplace=False):
     else:
         return A * B
 
+
 def empty_like(x, shape):
     return np.empty(shape, x.dtype)
+
 
 backend = namedtuple('backend', ['name', 'cdgmm', 'modulus', 'subsample_fourier', 'fft', 'Pad', 'unpad', 'empty_like'])
 backend.name = 'numpy'
--- a/kymatio/scattering2d/backend/numpy_backend.py
+++ b/kymatio/scattering2d/backend/numpy_backend.py
@@ -1,4 +1,4 @@
-# Authors: Edouard Oyallon, Sergey Zagoruyko
+# Authors: Edouard Oyallon, Sergey Zagoruyko, Muawiz Chaudhary
 
 import numpy as np
 from collections import namedtuple
--- a/kymatio/scattering2d/backend/numpy_backend.py
+++ b/kymatio/scattering2d/backend/numpy_backend.py
@@ -7,6 +7,14 @@ from collections import namedtuple
 BACKEND_NAME = 'numpy'
 
 
+def _iscomplex(x):
+    return x.dtype == np.complex64 or x.dtype == np.complex128
+
+
+def _isreal(x):
+    return x.dtype == np.float32 or x.dtype == np.float64
+
+
 class Pad(object):
     def __init__(self, pad_size, input_size, pre_pad=False):
         """
@@ -160,8 +168,18 @@ def cdgmm(A, B, inplace=False):
             C[b, c, m, n, :] = A[b, c, m, n, :] * B[m, n, :]
 
     """
+    if not _iscomplex(A):
+        raise TypeError('The first input must be complex.')
+
     if B.ndim != 2:
         raise RuntimeError('The dimension of the second input must be 2.')
+
+    if not _iscomplex(B) and not _isreal(B):
+        raise TypeError('The second input must be complex or real.')
+
+    if A.shape[-2:] != B.shape[-2:]:
+        raise RuntimeError('The inputs are not compatible for '
+                           'multiplication.')
 
     if inplace:
         return np.multiply(A, B, out=A)
--- a/kymatio/scattering2d/backend/numpy_backend.py
+++ b/kymatio/scattering2d/backend/numpy_backend.py
@@ -32,11 +32,7 @@ class Pad(object):
             return x
         else:
             np_pad = ((self.pad_size[0], self.pad_size[1]), (self.pad_size[2], self.pad_size[3]))
-            batch_shape = x.shape[:-2]
-            signal_shape = x.shape[-2:]
-            x = x.reshape((-1, 1) + signal_shape)
-            output = np.pad(x, ((0,0), (0,0), np_pad[0], np_pad[1]), mode='reflect')
-            output = output.reshape(batch_shape + output.shape[-2:])
+            output = np.pad(x, ((0,0), np_pad[0], np_pad[1]), mode='reflect')
             return output
 
 
--- a/kymatio/scattering2d/backend/numpy_backend.py
+++ b/kymatio/scattering2d/backend/numpy_backend.py
@@ -66,18 +66,16 @@ class SubsampleFourier(object):
         Parameters
         ----------
         x : tensor_like
-            input tensor with at least 5 dimensions, the last being the real
-             and imaginary parts.
-            Ideally, the last dimension should be a power of 2 to avoid errors.
+            input tensor with at least three dimensions.
         k : int
-            integer such that x is subsampled by 2**k along the spatial variables.
+            integer such that x is subsampled by k along the spatial variables.
 
         Returns
         -------
         res : tensor_like
             tensor such that its fourier transform is the Fourier
             transform of a subsampled version of x, i.e. in
-            FFT^{-1}(res)[u1, u2] = FFT^{-1}(x)[u1 * (2**k), u2 * (2**k)]
+            F^{-1}(res)[u1, u2] = F^{-1}(x)[u1 * k, u2 * k]
 
     """
     def __call__(self, x, k):
--- a/kymatio/scattering2d/core/scattering2d.py
+++ b/kymatio/scattering2d/core/scattering2d.py
@@ -28,7 +28,6 @@ def scattering2d(x, pad, unpad, backend, J, L, phi, psi, max_order):
     S_0 = unpad(S_0)
 
     output_shape = (x.shape[0], output_size) + S_0.shape[-2:]
-    print(output_shape)
 
     S = empty_like(x, output_shape)
 
--- a/kymatio/scattering2d/frontend/__init__.py
+++ b/kymatio/scattering2d/frontend/__init__.py
@@ -5,12 +5,14 @@ import importlib
 
 class Scattering2D(object):
     def __init__(self, *args, **kwargs):
+        frontend_suffixes = {'torch' : 'Torch', 'numpy' : 'NumPy', 'tensorflow'
+                : 'TensorFlow'}
         if 'frontend' not in kwargs:
             warnings.warn("Torch frontend is currently the default, but NumPy will become the default in the next"
                           " version.", PendingDeprecationWarning)
             frontend = 'torch'
         else:
-            frontend = kwargs['frontend']
+            frontend = kwargs['frontend'].lower()
             kwargs.pop('frontend')
 
         try:
@@ -18,9 +20,10 @@ class Scattering2D(object):
 
             # Create frontend-specific class name by inserting frontend name
             # after `Scattering`.
+            frontend = frontend_suffixes[frontend]
+
             class_name = self.__class__.__name__
-            class_name = (class_name[:-2] + frontend.capitalize()
-                          + class_name[-2:])
+            class_name = (class_name[:-2] + frontend + class_name[-2:])
 
             self.__class__ = getattr(module, class_name)
             self.__init__(*args, **kwargs)
--- a/kymatio/scattering2d/frontend/numpy_frontend.py
+++ b/kymatio/scattering2d/frontend/numpy_frontend.py
@@ -1,0 +1,45 @@
+from ...frontend.numpy_frontend import ScatteringNumPy
+from kymatio.scattering2d.core.scattering2d import scattering2d
+from .base_frontend import ScatteringBase2D
+from ..utils import compute_padding
+from ..filter_bank import filter_bank
+import numpy as np
+
+class ScatteringNumPy2D(ScatteringNumPy, ScatteringBase2D):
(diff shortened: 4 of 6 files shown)


kymatio/scattering2d/backend/numpy_backend.py
```diff
@@ -0,0 +1,178 @@
...
+        self.pad_size = pad_size
+
+    def __call__(self, x):
+        if self.pre_pad:
+            return x
+        else:
+            np_pad = ((self.pad_size[0], self.pad_size[1]), (self.pad_size[2], self.pad_size[3]))
+            batch_shape = x.shape[:-2]
+            signal_shape = x.shape[-2:]
+            x = x.reshape((-1, 1) + signal_shape)
```
MuawizChaudhary: Maybe it might be nice to eventually refactor out to core/scattering2d?

kymatio/scattering2d/backend/numpy_backend.py
```diff
@@ -0,0 +1,178 @@
...
+            output = np.pad(x, ((0,0), (0,0), np_pad[0], np_pad[1]), mode='reflect')
+            output = output.reshape(batch_shape + output.shape[-2:])
```
MuawizChaudhary: Same, but on torch backend its -3 instead. Maybe we should wait for the release of complex tensors?

kymatio/scattering2d/backend/numpy_backend.py
```diff
@@ -0,0 +1,178 @@
...
+            tensor such that its fourier transform is the Fourier
+            transform of a subsampled version of x, i.e. in
+            FFT^{-1}(res)[u1, u2] = FFT^{-1}(x)[u1 * (2**k), u2 * (2**k)]
+    """
+    def __call__(self, x, k):
+        batch_shape = x.shape[:-2]
+        signal_shape = x.shape[-2:]
+        
+        x = x.reshape((-1,) + signal_shape)
+        y = x.reshape(-1, k, x.shape[1] // k, k, x.shape[2] // k)
```
MuawizChaudhary: Torch backend uses -3 instead of -2, and view instead of reshape.

kymatio/scattering2d/backend/numpy_backend.py
```diff
@@ -0,0 +1,178 @@
...
+            if set to True, all the operations are performed inplace
+        Returns
+        -------
+        C : tensor
+            output tensor of size (B, C, M, N, 2) such that:
+            C[b, c, m, n, :] = A[b, c, m, n, :] * B[m, n, :]
+    """
+
+    if B.ndim != 2:
+        raise RuntimeError('The dimension of the second input must be 2.')
```
MuawizChaudhary: We do not have many warnings here.

kymatio/scattering2d/backend/numpy_backend.py
```diff
@@ -0,0 +1,178 @@
...
+
+    if inplace:
+        return np.multiply(A, B, out=A)
+    else:
+        return A * B
+
+def empty_like(x, shape):
+    # Note: This function is not documented, so it might be good to switch to
+    # something more standard.
+    return np.empty(shape, x.dtype)#, shape=shape)
```
MuawizChaudhary: Remove comment.

kymatio/scattering2d/core/scattering2d.py
```diff
@@ -17,20 +17,21 @@ def scattering2d(x, pad, unpad, backend, J, L, phi, psi, max_order):
...
-
+    
     U_0_c = fft(U_r, 'C2C')
 
     # First low pass filter
     U_1_c = cdgmm(U_0_c, phi[0])
     U_1_c = subsample_fourier(U_1_c, k=2 ** J)
 
     S_0 = fft(U_1_c, 'C2R', inverse=True)
+    print(S_0.shape)
```
MuawizChaudhary: Remove.

kymatio/scattering2d/core/scattering2d.py
```diff
@@ -17,20 +17,21 @@ def scattering2d(x, pad, unpad, backend, J, L, phi, psi, max_order):
...
     S_0 = unpad(S_0)
-
+    print(S_0.shape)
```
MuawizChaudhary: Remove.

kymatio/scattering2d/core/scattering2d.py
```diff
@@ -17,20 +17,21 @@ def scattering2d(x, pad, unpad, backend, J, L, phi, psi, max_order):
...
     output_shape = (x.shape[0], output_size) + S_0.shape[-2:]
-    print(output_shape)
 
     S = empty_like(x, output_shape)
+    print(S.shape, output_shape)
```
MuawizChaudhary: Remove.

kymatio/scattering2d/frontend/__init__.py
```diff
@@ -18,9 +18,13 @@ def __init__(self, *args, **kwargs):
...
             # after `Scattering`.
+            frontend = frontend.capitalize()
+            if frontend == 'Numpy':
+                frontend = 'NumPy'
             class_name = self.__class__.__name__
-            class_name = (class_name[:-2] + frontend.capitalize()
-                          + class_name[-2:])
+            class_name = (class_name[:-2] + frontend + class_name[-2:])
+            #print(module, class_name)
+            #raise RuntimeError('\nThe frontend \'' + frontend + '\' could not be correctly imported.')
```
MuawizChaudhary: I need thoughts on this mess of code.

kymatio/scattering2d/frontend/numpy_frontend.py
```diff
@@ -0,0 +1,45 @@
...
+
+        #if not input.is_contiguous():
+        #    raise RuntimeError('Tensor must be contiguous.')
+
+        if (input.shape[-1] != self.N or input.shape[-2] != self.M) and not self.pre_pad:
+            raise RuntimeError('NumPy array must be of spatial size (%i,%i).' % (self.M, self.N))
+
+        if (input.shape[-1] != self.N_padded or input.shape[-2] != self.M_padded) and self.pre_pad:
+            raise RuntimeError('Padded array must be of spatial size (%i,%i).' % (self.M_padded, self.N_padded))
+
```
MuawizChaudhary: I tried to make the above warnings as close as possible to the torch frontend.

kymatio/scattering2d/frontend/numpy_frontend.py
```diff
@@ -0,0 +1,45 @@
...
+        batch_shape = input.shape[:-2]
+        signal_shape = input.shape[-2:]
+
+        input = input.reshape((-1,) + signal_shape)
```
MuawizChaudhary: Perhaps there should be a function in backend where the job of the function is to do this stuff. Honestly, the batch_shape/signal_shape stuff could be a utility function, as the backends uses this code often.

kymatio/scattering2d/frontend/numpy_frontend.py
```diff
@@ -0,0 +1,45 @@
...
+       
+        S = scattering2d(input, self.pad, self.unpad, self.backend, self.J, self.L, self.phi, self.psi, self.max_order)
+        
+        scattering_shape = S.shape[-3:]
+
+        S = S.reshape(batch_shape + scattering_shape)
```
MuawizChaudhary: Same?

kymatio/scattering2d/backend/numpy_backend.py
```diff
@@ -0,0 +1,178 @@
...
+            Padding which allows to simultaneously pad in a reflection fashion
+            and map to complex.
+            Parameters
+            ----------
+            pad_size : list of 4 integers
+                size of padding to apply.
+            input_size : list of 2 integers
+                size of the original signal
+            pre_pad : boolean
+                if set to true, then there is no padding, one simply adds the imaginarty part.
```
lostanlen: " one simply adds the imaginarty part" -> "one simply converts from real to complex"

how useful is it to have this keyword argument though? using the `Pad` class to do anything else than padding seems like a weird design antipattern. If there is really a use case for doing the conversion, consider making a dedicated method. Otherwise, consider removing the functionality.

kymatio/scattering2d/backend/numpy_backend.py
```diff
@@ -0,0 +1,178 @@
...
+        Slices the input tensor at indices between 1::-1
+        Parameters
+        ----------
+        in_ : tensor_like
+            input tensor
+        Returns
+        -------
+        in_[..., 1:-1, 1:-1]
+    """
+    return np.expand_dims(in_[..., 1:-1, 1:-1],-3)
```
lostanlen: missing space after comma

kymatio/scattering2d/backend/numpy_backend.py
```diff
@@ -0,0 +1,178 @@
...
+    """
+        Subsampling of a 2D image performed in the Fourier domain
+        Subsampling in the spatial domain amounts to periodization
+        in the Fourier domain, hence the formula.
+        Parameters
+        ----------
+        x : tensor_like
+            input tensor with at least 5 dimensions, the last being the real
+             and imaginary parts.
+            Ideally, the last dimension should be a power of 2 to avoid errors.
```
lostanlen: huh? this part of the documentation is way too vague. what exactly is the technical spec here?

kymatio/scattering2d/backend/numpy_backend.py
MuawizChaudhary: I think I agree with this assessment, thoughts @edouardoyallon ?

kymatio/scattering2d/backend/numpy_backend.py
MuawizChaudhary: @edouardoyallon

kymatio/scattering2d/backend/numpy_backend.py
janden: Not sure that `reshape` is independent of the frontend. If it is, we can move it.

kymatio/scattering2d/backend/numpy_backend.py
janden: I don't understand. Wait to do what?

kymatio/scattering2d/backend/numpy_backend.py
janden: I don't understand.

kymatio/scattering2d/frontend/__init__.py
janden: It's in the commit message. Since we're now doing `ScatteringTorch2D` instead of `Scattering2DTorch`, some more acrobatics are necessary.

That being said, however, it won't work here, since the `capitalize()` method doesn't turn `numpy` into `NumPy` (or `tf`/`tensorflow` into `TensorFlow`, for that matter). I think a more robust approach would be to have a dict:
```
frontend_suffixes = {'torch': 'Torch', 'numpy': 'NumPy'}
class_name = 'Scattering' + suffixes[frontend] + '2D'
```

kymatio/scattering2d/frontend/numpy_frontend.py
janden: Good idea.

kymatio/scattering2d/backend/numpy_backend.py
janden: That last sentence should be removed (it is in the torch backend).

janden:
There is one more thing: coverage. The tests seem pretty sparse right now. They don't cover many of the errors, any of the tricky input sizes, any of the backend functions, or batch shape agnosticism. This needs to be fixed.

kymatio/scattering2d/backend/numpy_backend.py
MuawizChaudhary: wait until complex tensors are released in Pytorch to refactor this line out to core. 

If this was the torch backend, then we would want `output.shape[-2:]` to be a `output.shape[-3:]`.

kymatio/scattering2d/backend/numpy_backend.py
MuawizChaudhary: I had to put the reshape's here so that the tests would run. Even after adding the reshapes on the numpy frontend, these reshapes here are needed. 

Additionally, the torch frontend has the same reshape operations in the pad call function.

kymatio/scattering2d/backend/numpy_backend.py
MuawizChaudhary: On the torch backend, CDGMM has a lot of warnings it calls. Over here we do not have too many. For example, we could check that the filter must be a 3-tensor, check that the filters are compatible for multiplication or not, check if they're of the same time.  Atleast, thats some of the stuff thats checked for on the torch backend.

kymatio/scattering2d/frontend/numpy_frontend.py
```diff
@@ -0,0 +1,45 @@
...
+from kymatio.scattering2d.core.scattering2d import scattering2d
+from .base_frontend import ScatteringBase2D
+from ..utils import compute_padding
+from ..filter_bank import filter_bank
+import numpy as np
+
+class ScatteringNumPy2D(ScatteringNumPy, ScatteringBase2D):
+    def __init__(self, J, shape, L=8, max_order=2, pre_pad=False, backend='numpy'):
+        ScatteringNumPy.__init__(self)
+        ScatteringBase2D.__init__(**locals())
```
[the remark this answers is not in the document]
MuawizChaudhary: no to **locals()?

kymatio/scattering2d/frontend/numpy_frontend.py
MuawizChaudhary: Just making sure this is fine! Resolve if so.

kymatio/scattering2d/frontend/__init__.py
MuawizChaudhary: Do we want to define this dictionary in some base folder thats not in any of the scattering*d?

kymatio/scattering2d/backend/numpy_backend.py
janden: Which tests would fail? Which function calls, specifically. I thought this was only called by core `scattering2d()`.

kymatio/scattering2d/backend/numpy_backend.py
janden: I think that makes sense. We already have the check that it's a 3-tensor though, since in NumPy, that's just a complex 2-tensor (and we have the check above).

kymatio/scattering2d/frontend/numpy_frontend.py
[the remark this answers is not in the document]
janden: Correct. As we discussed in the torch PR, this can lead to bugs. (For example, is `self` passed as an argument here?)

kymatio/scattering2d/backend/numpy_backend.py
janden: Like I said in another comment, I don't think all of this reshaping needs to be done in the backend. We should simply be able to specify that the backend deals with arrays of shape `(B, N, M)` *always* (for the torch backends, that would be arrays of shape `(B, N, M, 2)`), and it's up to the frontend wrapper to deal with any other batch shapes.

kymatio/scattering2d/frontend/__init__.py
janden: Might make sense down the road, but we can just leave it here for now.

kymatio/scattering2d/backend/numpy_backend.py
```diff
@@ -0,0 +1,188 @@
...
+        ----------
+        in_ : tensor_like
+            input tensor
+
+        Returns
+        -------
+        in_[..., 1:-1, 1:-1]
+
+    """
+    return np.expand_dims(in_[..., 1:-1, 1:-1], -3)
```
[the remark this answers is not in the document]
edouardoyallon: it's related to the `concatenate` function, maybe it's not necessary, it simply depends on the backend internal routines

kymatio/scattering2d/backend/numpy_backend.py
edouardoyallon: Yeah. Absolutely.

kymatio/scattering2d/frontend/numpy_frontend.py
edouardoyallon: yes

kymatio/scattering2d/frontend/numpy_frontend.py
edouardoyallon: yes, cf comment above

kymatio/scattering2d/frontend/numpy_frontend.py
janden: So I think we just leave this for now, but in the future, it might be nice to have two functions: `x, batch_shape = compress_batch(x, dim)`, where `dim` corresponds to how many dimensions to “keep” (here, two), and `x = restore_batch(x, batch_shape)`. This way, we can use the functions across dimensions (and frontends) by changing the `dim` parameter.

kymatio/scattering2d/backend/numpy_backend.py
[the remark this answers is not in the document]
janden: Yeah I think this is a bad idea here. It causes problems in the tests

https://github.com/kymatio/kymatio/blob/caad82702a2856871341afee33e39667b5a79d99/kymatio/scattering2d/tests/test_numpy_scattering2d.py#L47

 (shape grows a singleton dimension after unpadding) and in the core

https://github.com/kymatio/kymatio/blob/caad82702a2856871341afee33e39667b5a79d99/kymatio/scattering2d/core/scattering2d.py#L54

(unpadded arrays have to be squeezed).

I suggest we remove this behavior.

kymatio/scattering2d/frontend/numpy_frontend.py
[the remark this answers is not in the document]
janden: Agreed and agreed.

kymatio/scattering2d/backend/numpy_backend.py
janden: This has now been removed (all batch reshaping is done in frontend).

kymatio/scattering2d/backend/numpy_backend.py
janden: This has now been removed (all batch reshaping is done in frontend).

kymatio/scattering2d/backend/numpy_backend.py
janden: I agree, but this PR is mostly here to refactor the code into the new frontend–backend framework. I suggest we take this (and other problems that cannot be addressed quickly) into an issue and address it later.

kymatio/scattering2d/backend/numpy_backend.py
janden: Again, this should be handled by the frontend.

kymatio/scattering2d/frontend/numpy_frontend.py
janden: Made this into an issue. Resolving.

kymatio/scattering2d/frontend/numpy_frontend.py
janden: See #461. Resolving.

kymatio/scattering2d/backend/numpy_backend.py
janden: Made into #462. Resolving.

kymatio/scattering2d/backend/numpy_backend.py
janden: Rewritten. Resolving.

kymatio/scattering2d/backend/numpy_backend.py
janden: Added.

kymatio/scattering2d/backend/numpy_backend.py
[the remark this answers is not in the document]
janden: Created an issue. Resolving.

janden:
This last test failure is not trivial. It seems that NumPy has some opinions about the default precision of its arrays. I wrote the tests assuming that the default precision is single (which seems to work on my computer), but it seems that the CI tests are using double precision as default. The culprit is likely the definition of the filters, so we may have to dig around in there to sort this out.

edouardoyallon:
I can't reproduce the issue, the tests pass on my machine(and half of it pass on travis). Can someone help with that?

edouardoyallon:
```
        if _base_dtype(A) is not _base_dtype(B):
>           raise TypeError('Inputs must be of the same base dtype.')
E           TypeError: Inputs must be of the same base dtype.
```

This instruction fails. I presume a package is missing when not using anaconda. I'm not familiar enough with python to understand what's going on.

edouardoyallon:
> This last test failure is not trivial. It seems that NumPy has some opinions about the default precision of its arrays. I wrote the tests assuming that the default precision is single (which seems to work on my computer), but it seems that the CI tests are using double precision as default. The culprit is likely the definition of the filters, so we may have to dig around in there to sort this out.

Oh. Ok.

kymatio/scattering2d/filter_bank.py
```diff
@@ -182,7 +182,7 @@ def gabor_2d(M, N, sigma, theta, xi, slant=1.0, offset=0):
             [xx, yy] = np.mgrid[offset + ex * M:offset + M + ex * M, offset + ey * N:offset + N + ey * N]
             arg = -(curv[0, 0] * np.multiply(xx, xx) + (curv[0, 1] + curv[1, 0]) * np.multiply(xx, yy) + curv[
                 1, 1] * np.multiply(yy, yy)) + 1.j * (xx * xi * np.cos(theta) + yy * xi * np.sin(theta))
-            gab += np.exp(arg)
+            gab = gab + np.exp(arg)
```
edouardoyallon: broadcasting isn't allowed with "+=", thus I switched this

kymatio/scattering2d/filter_bank.py
```diff
@@ -171,9 +171,9 @@ def gabor_2d(M, N, sigma, theta, xi, slant=1.0, offset=0):
         morlet_fft : ndarray
             numpy array of size (M, N)
     """
-    gab = np.zeros((M, N), np.complex64)
```
edouardoyallon: I removed any type hard-coding

kymatio/scattering2d/tests/test_numpy_scattering2d.py
```diff
@@ -0,0 +1,313 @@
...
+
+    @pytest.mark.parametrize('backend', backends)
+    def test_Scattering2D(self, backend):
+        test_data_dir = os.path.dirname(__file__)
+        data = None
+        with open(os.path.join(test_data_dir, 'test_data_2d.pt'), 'rb') as f:
+            buffer = io.BytesIO(f.read())
+            data = torch.load(buffer)
+
+        x = data['x'].numpy().astype(np.zeros(1).dtype)
```
edouardoyallon: hint

edouardoyallon:
Okay, so now, torch tests do not pass yet np tests do... gonna fix that!

kymatio/scattering2d/frontend/torch_frontend.py
```diff
@@ -16,7 +16,7 @@ def __init__(self, J, shape, L=8, max_order=2, pre_pad=False, backend='torch'):
         self.register_filters()
 
     def register_single_filter(self, v, n):
-        current_filter = torch.from_numpy(v).unsqueeze(-1)
+        current_filter = torch.from_numpy(v).unsqueeze(-1).type_as(torch.zeros(1))
```
edouardoyallon: funny trick again

edouardoyallon:
Et voilà. Let's move to tf? @MuawizChaudhary @janden

janden:
> Et voilà. Let's move to tf? @MuawizChaudhary @janden

Ok let me look through your changes, clean up, and then I'll merge if everything's good.

kymatio/scattering2d/filter_bank.py
janden: I think it would be better to hard-code, actually. That way it's clear from the docs what the default dtype is.

kymatio/scattering2d/tests/test_numpy_scattering2d.py
janden: Don't like this. Let's fix it once and for all.

kymatio/scattering2d/frontend/torch_frontend.py
janden: Yeah don't like this either.

janden:
So it was a matter of conda NP vs. pip NP? That's quite subtle...

kymatio/scattering2d/frontend/torch_frontend.py
edouardoyallon: Well, it's not that useless actually: imagine if an user needs to work in float64/float32

kymatio/scattering2d/tests/test_numpy_scattering2d.py
edouardoyallon: what if an user needs to change the type?

kymatio/scattering2d/filter_bank.py
edouardoyallon: the default type of np is user dependant na? It can be changed, isn't it?

edouardoyallon:
torch and np don't have the same default type, and depending on how np is installed, the default type can be float32 or float64 (afaiu)

kymatio/scattering2d/backend/numpy_backend.py
```diff
@@ -0,0 +1,188 @@
...
+    else:
+        return A * B
+
+
+def empty_like(x, shape):
+    return np.empty(shape, x.dtype)
+
+def squeeze(x, y, index):
+    y[..., index, :, :] = x.squeeze(-3)
+    return y
```
MuawizChaudhary: This is exactly the same code in the Torch backend squeeze method... would it be so wrong to do `from .torch_backend import squeeze` on the numpy backend? :P

kymatio/scattering2d/core/scattering2d.py
```diff
@@ -28,11 +29,10 @@ def scattering2d(x, pad, unpad, backend, J, L, phi, psi, max_order):
     S_0 = unpad(S_0)
 
     output_shape = (x.shape[0], output_size) + S_0.shape[-2:]
-    print(output_shape)
 
     S = empty_like(x, output_shape)
+    S = squeeze(S_0, S, 0)
```
MuawizChaudhary: This has to be done due to .squeeze() not being a method that works on tf tensors.

kymatio/scattering2d/backend/numpy_backend.py
edouardoyallon: yes

kymatio/scattering2d/core/scattering2d.py
edouardoyallon: makes sens

kymatio/scattering2d/filter_bank.py
janden: It is? How? Still, that's quite a roundabout way of doing it. Here's what I suggest: the default precision is double precision (`np.float64`/`np.complex128`). We then reimplement the same functions that we have in Torch to change the precision: `float()` (to get single precision) and `double()` (to get double precision). This way, the precision is user-defined, but in an easy-to-access way. A slight problem here would be that `scattering.float().double()` won't necessary return the original double-precision filters due to truncation error, but from what I understand, the Torch frontend has the same behavior, so maybe we can just leave it there for now.

kymatio/scattering2d/tests/test_numpy_scattering2d.py
janden: See above.

kymatio/scattering2d/frontend/torch_frontend.py
janden: So now they have to be able to change the default `dtype` in Torch as well? Sounds very difficult.

kymatio/scattering2d/backend/numpy_backend.py
janden: Yeah I don't think we should mix these. Imagine a user only has NumPy installed and not Torch, then they should still be able to run the code.

kymatio/scattering2d/core/scattering2d.py
janden: Wait are we doing TF now?

janden:
Does this `squeeze` business have to be done here? Now all the tests fail.

kymatio/scattering2d/backend/numpy_backend.py
edouardoyallon: yeah!

kymatio/scattering2d/core/scattering2d.py
edouardoyallon: No - could we remove this commit?

kymatio/scattering2d/filter_bank.py
edouardoyallon: I'd be pros the opposite: float32 and complex64 as the default because this is usually what is expected. I agree with your concern. Or maybe simply should we fix to work only with float32?

kymatio/scattering2d/frontend/torch_frontend.py
edouardoyallon: fair enough

kymatio/scattering2d/tests/test_numpy_scattering2d.py
edouardoyallon: agreed

kymatio/scattering2d/filter_bank.py
janden: Ok. How about this? Let's leave it at the NP default (as you have), *but* we relax the CDGMM function so that it doesn't require the same underlying type. We won't have the same interface as the `torch` backends, but we can address that later if we want. (Or not. In NP, you can multiply two arrays of different dtypes and get a sensible result. In Torch, you can't. It might make sense if our frontends mirrored that difference.)

kymatio/scattering2d/core/scattering2d.py
janden: Yes I'll remove.

kymatio/scattering2d/filter_bank.py
janden: Sorry, let me amend that. We leave it at *double precision*, but relax the constraints on the CDGMM.

janden: Gah. I mean *single precision*. Like we had it before. We just don't want it to error if people put in double-precision arrays.

kymatio/scattering2d/core/scattering2d.py
MuawizChaudhary: I reverted the commits, but feel free to drop the original commits from the history.

janden:
Looks like tests are passing. If everyone's ok with this (@edouardoyallon), I will clean up the commit history and merge tomorrow.

PR #464, opened 2019-12-03
MAINT Renormalizing FFT correctly

### LINKED ISSUE #386: Inverse fft in 2D backend

edouardoyallon:
remove inappropriate renormalization

Close #386

### DIFF
--- a/kymatio/scattering2d/frontend/numpy_frontend.py
+++ b/kymatio/scattering2d/frontend/numpy_frontend.py
@@ -1,8 +1,6 @@
 from ...frontend.numpy_frontend import ScatteringNumPy
 from kymatio.scattering2d.core.scattering2d import scattering2d
 from .base_frontend import ScatteringBase2D
-from ..utils import compute_padding
-from ..filter_bank import filter_bank
 import numpy as np
 
 class ScatteringNumPy2D(ScatteringNumPy, ScatteringBase2D):


kymatio/scattering2d/tests/test_torch_scattering2d.py
```diff
@@ -309,6 +309,8 @@ def test_fft(self, backend):
         y[1, 0, :] = x[0, 0, :] + x[0, 1, :] - x[1, 0, :] - x[1, 1, :]
         y[1, 1, :] = x[0, 0, :] - x[0, 1, :] - x[1, 0, :] + x[1, 1, :]
 
+        y = y/2.0 # renormalization
```
MuawizChaudhary: Why divide by 2?

kymatio/scattering2d/filter_bank.py
```diff
@@ -63,7 +61,7 @@ def filter_bank(M, N, J, L=8):
         phi_signal_fourier_res = periodize_filter_fft(phi_signal_fourier, res)
         filters['phi'][res] = phi_signal_fourier_res
         # Normalization to avoid doing it with the FFT.
-        filters['phi'][res] /= M*N // 2 ** (2 * J)
+        filters['phi'][res] *= M*N // 2 ** (2 * J)
```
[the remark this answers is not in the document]
edouardoyallon: this is the "true" renormalization

kymatio/scattering2d/tests/test_torch_scattering2d.py
edouardoyallon: 2^2=4, the FFT is unitary, and now the matrix y=Ax is unitary!

MuawizChaudhary: Never mind, reading the pytorch documentation the result of the FFT is normalized by dividing by the square root of the product of the first two dimensions.

janden: Can we make that clear? E.g. `y /= torch.sqrt(2 ** 2)`. It's a little silly, but more clear, I think.

kymatio/scattering2d/filter_bank.py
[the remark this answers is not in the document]
janden: You still haven't explained why the line above is removed while this one is changed. Also, spaces around asterisk.

kymatio/scattering2d/backend/torch_backend.py
```diff
@@ -223,12 +223,12 @@ def fft(x, direction='C2C', inverse=False):
         raise RuntimeError('Tensors must be contiguous.')
 
     if direction == 'C2R':
-        output = torch.irfft(x, 2, normalized=False, onesided=False) * x.shape[-2] * x.shape[-3]
+        output = torch.irfft(x, 2, normalized=True, onesided=False)
```
janden: The unitary FFT, really? This is quite non-standard. Why not stick to the regular ones?

edouardoyallon: well, the dotwise multiplication is anyway slow - i'm also ok to remove the "normalized" fft

kymatio/scattering2d/filter_bank.py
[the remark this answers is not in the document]
edouardoyallon: a normalization is due because we downsample

edouardoyallon:
It seems this catched a bigger bug: the numpy frontend/backend ran into an error, I'll correct and update!

kymatio/scattering2d/filter_bank.py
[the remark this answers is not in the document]
janden: And the ψ filters are not downsampled?

kymatio/scattering2d/backend/torch_backend.py
janden: Should be faster then with `normalized=False` since that's only one elementwise multiplication (in the IFFT).

kymatio/scattering2d/tests/test_torch_scattering2d.py
edouardoyallon: This is now fixed(plz resolve if OK)

Issue #463, opened 2019-12-03
Remove `expand_dims`/`unsqueeze` in `unpad`


janden:
It causes problems in the tests

https://github.com/kymatio/kymatio/blob/caad82702a2856871341afee33e39667b5a79d99/kymatio/scattering2d/tests/test_numpy_scattering2d.py#L47

 (shape grows a singleton dimension after unpadding) and in the core

https://github.com/kymatio/kymatio/blob/caad82702a2856871341afee33e39667b5a79d99/kymatio/scattering2d/core/scattering2d.py#L54

(unpadded arrays have to be squeezed).

I suggest we remove this behavior.

_Originally posted by @janden in https://github.com/kymatio/kymatio/pull/459_

This also applies to the torch frontend, where `expand_dims` is replaced by `unsqueeze`:

https://github.com/kymatio/kymatio/blob/c746cce7a2658b04089f197d06878f7faaa89f5b/kymatio/scattering2d/backend/torch_backend.py#L107

Issue #419, opened 2019-08-17
ENH removing torch from the tests


edouardoyallon:
Most of the "data" to compare with the MATLAB versions has been saved via torch. One should aim to remove this line, for instance

https://github.com/kymatio/kymatio/blob/kymatio-dev-frontend/kymatio/scattering2d/tests/test_tensorflow_scattering2d.py#L5

which will affect

https://github.com/kymatio/kymatio/blob/kymatio-dev-frontend/kymatio/scattering2d/tests/test_tensorflow_scattering2d.py#L32

PR #465, opened 2019-12-04
ENH TensorFlow 2D Back/Frontend [3c]

MuawizChaudhary:
This is the PR for the 2d Tensorflow back/front ends.  This is up for review after NumPy is done. 

Unfortunately, squeeze is not an attribute on tensorflow tensors, so the biggest barrier currently is removing the usage of squeeze from core/scattering2d.

### DIFF
--- a/kymatio/scattering2d/backend/tensorflow_backend.py
+++ b/kymatio/scattering2d/backend/tensorflow_backend.py
@@ -1,0 +1,177 @@
+# Authors: Edouard Oyallon, Sergey Zagoruyko
+
+import tensorflow as tf
+import numpy as np
+from collections import namedtuple
+
+BACKEND_NAME = 'tensorflow'
+
+def _iscomplex(x):
+    return x.dtype == np.complex64 or x.dtype == np.complex128
+
+def _isreal(x):
+    return x.dtype == np.float32 or x.dtype == np.float64
+
+class Pad(object):
+    def __init__(self, pad_size, input_size, pre_pad=False):
+        """
+            Padding which allows to simultaneously pad in a reflection fashion
+            and map to complex.
+            Parameters
+            ----------
+            pad_size : list of 4 integers
+                size of padding to apply.
+            input_size : list of 2 integers
+                size of the original signal
+            pre_pad : boolean
+                if set to true, then there is no padding, one simply adds the imaginarty part.
+        """
+        self.pre_pad = pre_pad
+        self.pad_size = pad_size
+
+    def __call__(self, x):
+        if self.pre_pad:
+            return x
+        else:
+            paddings = [[0, 0]] * len(x.shape[:-2])
+            paddings += [[self.pad_size[0], self.pad_size[1]], [self.pad_size[2], self.pad_size[3]]]
+            return tf.cast(tf.pad(x, paddings, mode="REFLECT"), tf.complex64)
+
+def unpad(in_):
+    """
+        Slices the input tensor at indices between 1::-1
+        Parameters
+        ----------
+        in_ : tensor_like
+            input tensor
+        Returns
+        -------
+        in_[..., 1:-1, 1:-1]
+    """
+    return in_[..., 1:-1, 1:-1]
+
+class SubsampleFourier(object):
+    """ Subsampling of a 2D image performed in the Fourier domain.
+
+        Subsampling in the spatial domain amounts to periodization
+        in the Fourier domain, hence the formula.
+
+        Parameters
+        ----------
+        x : tensor_like
+            input tensor with at least three dimensions.
+        k : int
+            integer such that x is subsampled by k along the spatial variables.
+
+        Returns
+        -------
+        res : tensor_like
+            tensor such that its fourier transform is the Fourier
+            transform of a subsampled version of x, i.e. in
+            F^{-1}(res)[u1, u2] = F^{-1}(x)[u1 * k, u2 * k]
+
+    """
+    def __call__(self, x, k):
+        y = tf.reshape(x, (-1, k, x.shape[1] // k, k, x.shape[2] // k))
+
+        out = tf.reduce_mean(y, axis=(1, 3))
+        return out
+
+
+class Modulus(object):
+    """
+        This class implements a modulus transform for complex numbers.
+        Usage
+        -----
+        modulus = Modulus()
+        x_mod = modulus(x)
+        Parameters
+        ---------
+        x: input complex tensor.
+        Returns
+        -------
+        output: a real tensor equal to the modulus of x.
+    """
+    def __call__(self, x):
+        norm = tf.abs(x)
+        return tf.cast(norm, tf.complex64)
+
+
+def fft(x, direction='C2C', inverse=False):
+    """
+        Interface with torch FFT routines for 2D signals.
+        Example
+        -------
+        x = torch.randn(128, 32, 32, 2)
+        x_fft = fft(x, inverse=True)
+        Parameters
+        ----------
+        input : tensor
+            complex input for the FFT
+        direction : string
+            'C2R' for complex to real, 'C2C' for complex to complex
+        inverse : bool
+            True for computing the inverse FFT.
+            NB : if direction is equal to 'C2R', then an error is raised.
+    """
+    if direction == 'C2R':
+        if not inverse:
+            raise RuntimeError('C2R mode can only be done with an inverse FFT.')
+
+    if direction == 'C2R':
+        output = tf.math.real(tf.signal.ifft2d(x, name='irfft2d'))
+    elif direction == 'C2C':
+        if inverse:
+            output = tf.signal.ifft2d(x, name='ifft2d')
+        else:
+            output = tf.signal.fft2d(x, name='fft2d')
+
+    return output
+
+
+def cdgmm(A, B, inplace=False):
+    """
+        Complex pointwise multiplication between (batched) tensor A and tensor B.
+        Parameters
+        ----------
+        A : tensor
+            A is a complex tensor of size (B, C, M, N, 2)
+        B : tensor
+            B is a complex tensor of size (M, N) or real tensor of (M, N)
+        inplace : boolean, optional
+            if set to True, all the operations are performed inplace
+        Returns
+        -------
+        C : tensor
+            output tensor of size (B, C, M, N, 2) such that:
+            C[b, c, m, n, :] = A[b, c, m, n, :] * B[m, n, :]
+    """
+    if B.ndim != 2:
+        raise RuntimeError('The dimension of the second input must be 2.')
+
+    if not _iscomplex(A):
+        raise TypeError('The first input must be complex.')
+
+    if not _iscomplex(B) and not _isreal(B):
+        raise TypeError('The second input must be complex or real.')
+
+    if A.shape[-2:] != B.shape[-2:]:
+        raise RuntimeError('The inputs are not compatible for '
+                           'multiplication.')
+
+    return A * B
+
+def concatenate(arrays):
+    return tf.stack(arrays, axis=-3)
+
+
+backend = namedtuple('backend', ['name', 'cdgmm', 'modulus', 'subsample_fourier', 'fft', 'Pad', 'unpad', 'concatenate'])
+
+backend.name = 'tensorflow'
+backend.cdgmm = cdgmm
+backend.modulus = Modulus()
+backend.subsample_fourier = SubsampleFourier()
+backend.fft = fft
+backend.Pad = Pad
+backend.unpad = unpad
+backend.concatenate = concatenate
--- a/kymatio/scattering2d/backend/tensorflow_backend.py
+++ b/kymatio/scattering2d/backend/tensorflow_backend.py
@@ -1,4 +1,4 @@
-# Authors: Edouard Oyallon, Sergey Zagoruyko
+# Authors: Edouard Oyallon, Sergey Zagoruyko, Muawiz Chaudhary
 
 import tensorflow as tf
 import numpy as np
--- a/kymatio/scattering2d/frontend/tensorflow_frontend.py
+++ b/kymatio/scattering2d/frontend/tensorflow_frontend.py
@@ -1,0 +1,47 @@
+import tensorflow as tf
+from ...frontend.tensorflow_frontend import ScatteringTensorFlow
+from kymatio.scattering2d.core.scattering2d import scattering2d
+from .base_frontend import ScatteringBase2D
+
+class ScatteringTensorFlow2D(ScatteringTensorFlow, ScatteringBase2D):
+    def __init__(self, J, shape, L=8, max_order=2, pre_pad=False, backend='tensorflow', name='Scattering2D'):
+        ScatteringTensorFlow.__init__(self, name)
+        ScatteringBase2D.__init__(self, J, shape, L, max_order, pre_pad, backend)
+        ScatteringBase2D._instantiate_backend(self, 'kymatio.scattering2d.backend.')
+        ScatteringBase2D.build(self)
+        ScatteringBase2D.create_filters(self)
+
+    def scattering(self, input):
+        with tf.name_scope('scattering') as scope:
+            try:
+                input = tf.convert_to_tensor(input)
+            except ValueError:
+                raise TypeError('The input should be convertible to a '
+                                'TensorFlow Tensor.')
+
+            if len(input.shape) < 2:
+                raise RuntimeError('Input tensor should have at least two '
+                                   'dimensions.')
+
+            if (input.shape[-1] != self.N or input.shape[-2] != self.M) and not self.pre_pad:
+                raise RuntimeError('Tensor must be of spatial size (%i,%i).' % (self.M, self.N))
+
+            if (input.shape[-1] != self.N_padded or input.shape[-2] != self.M_padded) and self.pre_pad:
+                raise RuntimeError('Padded tensor must be of spatial size (%i,%i).' % (self.M_padded, self.N_padded))
+
+            batch_shape = tuple(input.shape[:-2])
+            signal_shape = tuple(input.shape[-2:])
+
+            input = tf.reshape(input, (-1,) + signal_shape)
+
+            S = scattering2d(input, self.pad, self.unpad, self.backend, self.J, self.L, self.phi, self.psi, self.max_order)
+            scattering_shape = tuple(S.shape[-3:])
+            S = tf.reshape(S, batch_shape + scattering_shape)
+            return S
+
+    @tf.Module.with_name_scope
+    def __call__(self, x):
+        return self.scattering(x)
+
+
+__all__ = ['ScatteringTensorFlow2D']
--- a/kymatio/scattering2d/tests/test_tensorflow_scattering2d.py
+++ b/kymatio/scattering2d/tests/test_tensorflow_scattering2d.py
@@ -1,0 +1,314 @@
+import os
+import io
+import tensorflow as tf
+from kymatio.scattering2d import Scattering2D
+import torch
+import numpy as np
+import pytest
+from collections import namedtuple
+
+backends = []
+
+from kymatio.scattering2d.backend.tensorflow_backend import backend
+backends.append(backend)
+
+class TestPad:
+    @pytest.mark.parametrize('backend', backends)
+    def test_Pad(self, backend):
+        pad = backend.Pad((2, 2, 2, 2), (4, 4), pre_pad=False)
+
+        x = np.random.randn(4, 4) + 1J * np.random.randn(4, 4)
+        x = x[np.newaxis, ...]
+
+        z = pad(x)
+
+        assert z.shape == (1, 8, 8)
+        assert np.isclose(z[0, 2, 2], x[0, 0, 0])
+        assert np.isclose(z[0, 1, 0], x[0, 1, 2])
+        assert np.isclose(z[0, 1, 1], x[0, 1, 1])
+        assert np.isclose(z[0, 1, 2], x[0, 1, 0])
+        assert np.isclose(z[0, 1, 3], x[0, 1, 1])
+
+        pad = backend.Pad((2, 2, 2, 2), (4, 4), pre_pad=True)
+
+        x = np.random.randn(8, 8) + 1J * np.random.randn(8, 8)
+
+        z = pad(x)
+
+        assert np.allclose(x, z)
+
+    @pytest.mark.parametrize('backend', backends)
+    def test_unpad(self, backend):
+        x = np.random.randn(4, 4) + 1J * np.random.randn(4, 4)
+
+        y = backend.unpad(x)
+
+        assert y.shape == (2, 2)
+        assert np.isclose(y[0, 0], x[1, 1])
+        assert np.isclose(y[0, 1], x[1, 2])
+
+class TestModulus:
+    @pytest.mark.parametrize('backend', backends)
+    def test_Modulus(self, backend):
+        modulus = backend.modulus
+
+        x = np.random.rand(100, 10, 4) + 1J * np.random.rand(100, 10, 4)
+
+        y = modulus(x)
+        u = np.squeeze(np.sqrt(np.real(x) ** 2 + np.imag(x) ** 2))
+        v = y
+        assert np.allclose(u, v)
+
+class TestSubsampleFourier:
+    @pytest.mark.parametrize('backend', backends)
+    def test_SubsampleFourier(self, backend):
+        subsample_fourier = backend.subsample_fourier
+
+        x = (np.random.rand(100, 128, 128)
+             + 1J * np.random.rand(100, 128, 128))
+
+        y = np.zeros((100, 8, 8), dtype=np.complex128)
+
+        from itertools import product
+        for i, j in product(range(8), range(8)):
+            for m, n in product(range(16), range(16)):
+                y[..., i, j] += x[..., i + m * 8, j + n * 8]
+
+        y /= 16 ** 2
+
+        z = subsample_fourier(x, k=16)
+        assert np.allclose(y, z)
+
+class TestCDGMM:
+    @pytest.fixture(params=(False, True))
+    def data(self, request):
+        real_filter = request.param
+        x = (np.random.randn(100, 128, 128)
+             + 1J * np.random.randn(100, 128, 128))
+        filt = (np.random.randn(128, 128)
+                + 1J * np.random.randn(128, 128))
+        y = (np.random.randn(100, 128, 128)
+             + 1J * np.random.randn(100, 128, 128))
+
+        if real_filter:
+            filt = np.real(filt)
+
+        y = x * filt
+
+        return x, filt, y
+
+    @pytest.mark.parametrize('backend', backends)
+    @pytest.mark.parametrize('inplace', (False, True))
+    def test_cdgmm_forward(self, data, backend, inplace):
+        x, filt, y = data
+
+        z = backend.cdgmm(x, filt, inplace=inplace)
+
+        assert np.allclose(y, z)
+
+    @pytest.mark.parametrize('backend', backends)
+    def test_cdgmm_exceptions(self, backend):
+        with pytest.raises(TypeError) as record:
+            backend.cdgmm(np.empty((3, 4, 5)).astype(np.float64),
+                          np.empty((4, 5)).astype(np.complex128))
+        assert 'first input must be complex' in record.value.args[0]
+
+        with pytest.raises(RuntimeError) as record:
+            backend.cdgmm(np.empty((3, 4, 5)).astype(np.complex128),
+                          np.empty((3, 4, 5)).astype(np.complex128))
+        assert 'second input must be 2' in record.value.args[0]
+
+        with pytest.raises(TypeError) as record:
+            backend.cdgmm(np.empty((3, 4, 5)).astype(np.complex128),
+                          np.empty((4, 5)).astype(np.int64))
+        assert 'second input must be complex or real' in record.value.args[0]
+
+        with pytest.raises(RuntimeError) as record:
+            backend.cdgmm(np.empty((3, 4, 5)).astype(np.complex128),
+                          np.empty((4, 6)).astype(np.complex128))
+        assert 'not compatible for multiplication' in record.value.args[0]
+
+
+
+class TestFFT:
+    @pytest.mark.parametrize('backend', backends)
+    def test_fft(self, backend):
+        x = np.random.randn(2, 2) + 1J * np.random.randn(2, 2)
+        x = x.astype('complex64')
+        y = np.array([[x[0, 0] + x[0, 1] + x[1, 0] + x[1, 1],
+                       x[0, 0] - x[0, 1] + x[1, 0] - x[1, 1]],
+                      [x[0, 0] + x[0, 1] - x[1, 0] - x[1, 1],
+                       x[0, 0] - x[0, 1] - x[1, 0] + x[1, 1]]])
+
+        z = backend.fft(x, direction='C2C')
+
+        assert np.allclose(y, z)
+
+        z = backend.fft(x, direction='C2C', inverse=True)
+
+        z = z * 4
+
+        assert np.allclose(y, z)
+
+        z = backend.fft(x, direction='C2R', inverse=True)
+
+        z = z * 4
+
+        assert not np.iscomplexobj(z)
+        assert np.allclose(np.real(y), z)
+
+
+    @pytest.mark.parametrize('backend', backends)
+    def test_fft_exceptions(self, backend):
+        with pytest.raises(RuntimeError) as record:
+            backend.fft(np.empty((2, 2)), direction='C2R',
+                        inverse=False)
+        assert 'done with an inverse' in record.value.args[0]
+
+
+class TestBackendUtils:
+    @pytest.mark.parametrize('backend', backends)
+    def test_concatenate(self, backend):
+        x = np.random.randn(3, 6, 6) + 1J * np.random.randn(3, 6, 6)
+        y = np.random.randn(3, 6, 6) + 1J * np.random.randn(3, 6, 6)
+        z = np.random.randn(3, 6, 6) + 1J * np.random.randn(3, 6, 6)
+
+        w = backend.concatenate((x, y, z))
+
+        assert w.shape == (x.shape[0],) + (3,) + (x.shape[-2:])
+        assert np.allclose(w[:, 0, ...], x)
+        assert np.allclose(w[:, 1, ...], y)
+        assert np.allclose(w[:, 2, ...], z)
+
+
+class TestScattering2DTensorFlow:
+    def reorder_coefficients_from_interleaved(self, J, L):
+        # helper function to obtain positions of order0, order1, order2 from interleaved
+        order0, order1, order2 = [], [], []
+        n_order0, n_order1, n_order2 = 1, J * L, L ** 2 * J * (J - 1) // 2
+        n = 0
+        order0.append(n)
+        for j1 in range(J):
+            for l1 in range(L):
+                n += 1
+                order1.append(n)
+                for j2 in range(j1 + 1, J):
+                    for l2 in range(L):
+                        n += 1
+                        order2.append(n)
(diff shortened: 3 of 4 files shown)


kymatio/scattering2d/frontend/tensorflow_frontend.py
```diff
@@ -0,0 +1,38 @@
...
+
+class ScatteringTensorFlow2D(ScatteringTensorFlow, ScatteringBase2D):
+    def __init__(self, J, shape, L=8, max_order=2, pre_pad=False, backend='tensorflow', name='Scattering2D'):
+        ScatteringTensorFlow.__init__(self, name)
+        ScatteringBase2D.__init__(self, J, shape, L, max_order, pre_pad, backend)
+        ScatteringBase2D._instantiate_backend(self, 'kymatio.scattering2d.backend.')
+        ScatteringBase2D.build(self)
+        ScatteringBase2D.create_filters(self)
+
+    def scattering(self, input):
```
MuawizChaudhary: We should add some warning checks here like the torch/numpy frontends.

kymatio/scattering2d/tests/test_tensorflow_scattering2d.py
```diff
@@ -0,0 +1,74 @@
+import os
+import io
+import tensorflow as tf
+from kymatio.scattering2d import Scattering2D
+import torch
+import numpy as np
+import pytest
+from collections import namedtuple
+
+class TestScattering2DTensorflow:
```
MuawizChaudhary: Add more tests

kymatio/scattering2d/backend/tensorflow_backend.py
```diff
@@ -0,0 +1,165 @@
...
+    print(tf.shape(A))
+    return A * B
+
+def empty_like(x, shape):
+    return tf.zeros(shape, dtype=x.dtype)
+
+def squeeze(x, y, index):
+    y_list = tf.unstack(y, axis = -3)
+    y_list[index] = tf.squeeze(x, axis = -3)
+    return tf.stack(y_list, axis = -3)
```
MuawizChaudhary: this seems like SUCH a large amount of memory being moved around if `y` was large...

edouardoyallon: I don't know how tf works.

kymatio/scattering2d/tests/test_tensorflow_scattering2d.py
edouardoyallon: I agree - I'm not a good candidate to this task - I don't understand much of tf!

kymatio/scattering2d/frontend/tensorflow_frontend.py
edouardoyallon: I'm not sure it does apply here to a similar extent: tf has no "lowlevel" tensor contrary to torch!

edouardoyallon:
Can we move the commit to add squeeze here, and then make sure the tests run and then .. what shoulw we test?

MuawizChaudhary:
It seems like cherry-picking a commit for a file that does not exist in the current branch does not do anything...

kymatio/scattering2d/backend/tensorflow_backend.py
edouardoyallon: I think none of us if a tf user, would it be OK to wait for tf user to improve our code?(maybe @janden wants to dedicate some times to optimize this..)

kymatio/scattering2d/frontend/__init__.py
```diff
@@ -5,22 +5,25 @@
...
             module = importlib.__import__(frontend + '_frontend', globals(), locals(), [], 1)
 
             # Create frontend-specific class name by inserting frontend name
             # after `Scattering`.
+            frontend = frontend_suffixes[frontend]
+
             class_name = self.__class__.__name__
-            class_name = (class_name[:-2] + frontend.capitalize()
-                          + class_name[-2:])
+            class_name = (class_name[:-2] + frontend + class_name[-2:])
```
edouardoyallon: why is it modified here?

kymatio/scattering2d/backend/tensorflow_backend.py
MuawizChaudhary: Might be a good idea. 

I'll try to spend some time tomorrow to see whats the best way of doing this stuff. (right now I should sleep, 2am almost lol )

kymatio/scattering2d/frontend/__init__.py
MuawizChaudhary: We use a dictionary to get the frontend name so that we can import the correct class.

edouardoyallon: is it realted to tf and couldn't it be done elsewhere?

MuawizChaudhary: Not sure. It should have been modified in the numpy 2d pr.

edouardoyallon:
The numpy backend `squeeze` function was not implemented

kymatio/scattering2d/frontend/__init__.py
edouardoyallon: so let's remove it, and let's add it in a separated PR

edouardoyallon:
In this PR, we must also modify travis and jenkins and decide which version of `tf`we waant to use.

janden:
> Unfortunately, squeeze is not an attribute on tensorflow tensors, so the biggest barrier currently is removing the usage of squeeze from core/scattering2d.

This whole business will be fixed by addressing #463, which removes the need for `squeeze` in core. Will get started on a PR.

janden:
Ok please see #467.

kymatio/scattering2d/frontend/__init__.py
janden: Since NP wasn't merged at the time this PR was started, it appears as a diff here. What we have to do is rebase the branch on top of the updated `kymatio-v2` and then these wrinkles should shake out. I suggest we wait until #467 is merged, however, since that will also affect the work here.

kymatio/scattering2d/frontend/tensorflow_frontend.py
janden: > tf has no "lowlevel" tensor contrary to torch!

Could you expand on this a bit?

kymatio/scattering2d/tests/test_tensorflow_scattering2d.py
janden: Well, we should have at least the same types of tests that we do for torch. Putting code out that we don't test because we don't know how to test it sounds like a bad idea.

kymatio/scattering2d/backend/tensorflow_backend.py
janden: Yeah I'm fine leaving this somewhat unoptimized for now. Important thing is to have something that works and has an interface that we believe will be stable for the next few versions.

MuawizChaudhary: As discussed in #468 , we will be bringing back concatenate.

kymatio/scattering2d/backend/tensorflow_backend.py
```diff
@@ -0,0 +1,176 @@
+# Authors: Edouard Oyallon, Sergey Zagoruyko
+
+import tensorflow as tf
+import numpy as np
+from collections import namedtuple
+
+BACKEND_NAME = 'tensorflow'
+
+def _iscomplex(x):
+    return x.dtype == np.complex64 or x.dtype == np.complex128
```
MuawizChaudhary: When Tensorflow tensors are 'evaluated', they are automatically converted into numpy arrays. So this is fine.

kymatio/scattering2d/backend/tensorflow_backend.py
```diff
@@ -0,0 +1,176 @@
...
+
+def _isreal(x):
+    return x.dtype == np.float32 or x.dtype == np.float64
```
MuawizChaudhary: Same.

kymatio/scattering2d/tests/test_tensorflow_scattering2d.py
```diff
@@ -0,0 +1,366 @@
...
+import numpy as np
+import pytest
+from collections import namedtuple
+
+backends = []
+
+from kymatio.scattering2d.backend.tensorflow_backend import backend
+backends.append(backend)
+
+class TestPad:
```
MuawizChaudhary: Alot of the tests in this file are exactly the same as the ones on the Numpy end. The only changes are tensorflow is written instead of numpy, and there's one small difference I'll point out in a comment.

kymatio/scattering2d/tests/test_tensorflow_scattering2d.py
```diff
@@ -0,0 +1,366 @@
...
+                          np.empty((4, 6)).astype(np.complex128))
+        assert 'not compatible for multiplication' in record.value.args[0]
+
+
+
+class TestFFT:
+    @pytest.mark.parametrize('backend', backends)
+    def test_fft(self, backend):
+        x = np.random.randn(2, 2) + 1J * np.random.randn(2, 2)
+        x = x.astype('complex64')
```
MuawizChaudhary: Here is the difference! we have to cast to `complex64` in order for the test here to pass. On the Numpy, we don't need to cast to `complex64`. 

This is the only difference, else most things are the same. We should probably look into unifying both set of tests.

kymatio/scattering2d/tests/test_tensorflow_scattering2d.py
MuawizChaudhary: Resolve?

kymatio/scattering2d/backend/tensorflow_backend.py
```diff
@@ -0,0 +1,176 @@
...
+                if set to true, then there is no padding, one simply adds the imaginarty part.
+        """
+        self.pre_pad = pre_pad
+        self.pad_size = pad_size
+
+    def __call__(self, x):
+        if self.pre_pad:
+            return x
+        else:
+            paddings = [[0, 0]]*len(x.shape[:-2])
```
MuawizChaudhary: Wouldn't it make sense to be explicit and just write out ` [[0, 0], [0, 0]]`?

kymatio/scattering2d/backend/tensorflow_backend.py
MuawizChaudhary: Also, I tried `tf.complex64` and `tf.complex128` and it just error ed out. I'll try again tomorrow to make sure, but I am very sure it didn't work.

kymatio/scattering2d/backend/tensorflow_backend.py
```diff
@@ -0,0 +1,176 @@
...
+    def __init__(self, pad_size, input_size, pre_pad=False):
+        """
+            Padding which allows to simultaneously pad in a reflection fashion
+            and map to complex.
+            Parameters
+            ----------
+            pad_size : list of 4 integers
+                size of padding to apply.
+            input_size : list of 2 integers
+                size of the original signal
```
MuawizChaudhary: Should we remove?

kymatio/scattering2d/backend/tensorflow_backend.py
```diff
@@ -0,0 +1,176 @@
...
+        self.pre_pad = pre_pad
+        self.pad_size = pad_size
+
+    def __call__(self, x):
+        if self.pre_pad:
+            return x
+        else:
+            paddings = [[0, 0]]*len(x.shape[:-2])
+            paddings += [[self.pad_size[0],self.pad_size[1]],[self.pad_size[2],self.pad_size[3]]]
+            return tf.cast(tf.pad(x, paddings, mode="REFLECT"), tf.complex64)
```
MuawizChaudhary: Do we want to be casting to `complex64`?  Or `complex128`?

kymatio/scattering2d/backend/tensorflow_backend.py
```diff
@@ -0,0 +1,176 @@
...
+        inverse : bool
+            True for computing the inverse FFT.
+            NB : if direction is equal to 'C2R', then an error is raised.
+    """
+    if direction == 'C2R':
+        if not inverse:
+            raise RuntimeError('C2R mode can only be done with an inverse FFT.')
+
+    if direction == 'C2R':
+        output = tf.math.real(tf.signal.ifft2d(x, name='irfft2d'))*tf.cast(x.shape[-1]*x.shape[-2],tf.float32)
```
MuawizChaudhary: Will this normalization be needed after #470 ?

kymatio/scattering2d/backend/tensorflow_backend.py
```diff
@@ -0,0 +1,176 @@
...
+    
+    if not _iscomplex(A):
+        raise TypeError('The first input must be complex.')
+    
+    if not _iscomplex(B) and not _isreal(B):
+        raise TypeError('The second input must be complex or real.')
+
+    if A.shape[-2:] != B.shape[-2:]:
+        raise RuntimeError('The inputs are not compatible for '
+                           'multiplication.')
```
MuawizChaudhary: I like the wording on this.

kymatio/scattering2d/tests/test_tensorflow_scattering2d.py
```diff
@@ -0,0 +1,366 @@
...
+
+        assert np.allclose(Sg, S)
+
+    def test_inputs(self):
+        fake_backend = namedtuple('backend', ['name',])
+        fake_backend.name = 'fake'
+
+        with pytest.raises(RuntimeError) as ve:
+            scattering = Scattering2D(2, shape=(10, 10), frontend='tensorflow', backend=fake_backend)
+        assert 'not supported' in ve.value.args[0]
```
MuawizChaudhary: This errors out even through the code is basically the same in the numpy test file.

kymatio/scattering2d/backend/tensorflow_backend.py
edouardoyallon: So I think this function is not worth here because indeed, tensorflow will throw quickly an error.

kymatio/scattering2d/backend/tensorflow_backend.py
edouardoyallon: same as above

kymatio/scattering2d/backend/tensorflow_backend.py
edouardoyallon: @janden view?

kymatio/scattering2d/backend/tensorflow_backend.py
edouardoyallon: I'm ok to keep it this wya.

kymatio/scattering2d/backend/tensorflow_backend.py
edouardoyallon: @janden 64, isn't it?

kymatio/scattering2d/backend/tensorflow_backend.py
edouardoyallon: I don't think so. But for now, it is necessary. I'll update the second paper merged such that it'llwork.

kymatio/scattering2d/tests/test_tensorflow_scattering2d.py
edouardoyallon: Fine

kymatio/scattering2d/tests/test_tensorflow_scattering2d.py
edouardoyallon: isn't it a sort of cheat? I'm not sure how tf works

kymatio/scattering2d/tests/test_tensorflow_scattering2d.py
edouardoyallon: OK - so we need to add that in frontend

kymatio/scattering2d/tests/test_tensorflow_scattering2d.py
edouardoyallon: ok

kymatio/scattering2d/tests/test_tensorflow_scattering2d.py
MuawizChaudhary: Explain what you mean by cheat? 

The casting is needed in order for the tests to pass.

kymatio/scattering2d/tests/test_tensorflow_scattering2d.py
janden: Should be `ImportError`, not `RuntimeError`.

kymatio/scattering2d/tests/test_tensorflow_scattering2d.py
janden: Why was this done exactly? Why should we expect feeding NumPy arrays into the TF frontend to work?

edouardoyallon: there are two modes with tf, but I'm not much familiar with it

kymatio/scattering2d/tests/test_tensorflow_scattering2d.py
edouardoyallon: I don't know which tests to add

kymatio/scattering2d/backend/tensorflow_backend.py
janden: It's mostly for compatibility. Once we implement the skcuda `cdgmm`, we need these safeguards to avoid segfaults.

kymatio/scattering2d/backend/tensorflow_backend.py
janden: Why?

kymatio/scattering2d/backend/tensorflow_backend.py
janden: I'd prefer `2 * [[0, 0]]` or `2 * [2 * [0]]`.

kymatio/scattering2d/backend/tensorflow_backend.py
janden: Yes.

kymatio/scattering2d/backend/tensorflow_backend.py
```diff
@@ -0,0 +1,176 @@
...
+            input tensor
+        Returns
+        -------
+        in_[..., 1:-1, 1:-1]
+    """
+    return in_[..., 1:-1, 1:-1]
+
+class SubsampleFourier(object):
+    """
+        Subsampling of a 2D image performed in the Fourier domain
```
janden: Fix this docstring so it corresponds to the NP and Torch versions (e.g., no `2 ** k`, no powers of two to avoid errors, no FFT where we should have F, etc.).

kymatio/scattering2d/backend/tensorflow_backend.py
janden: Agreed.

kymatio/scattering2d/backend/tensorflow_backend.py
janden: So... do you want to change it?

kymatio/scattering2d/backend/tensorflow_backend.py
```diff
@@ -0,0 +1,176 @@
...
+        raise TypeError('The second input must be complex or real.')
+
+    if A.shape[-2:] != B.shape[-2:]:
+        raise RuntimeError('The inputs are not compatible for '
+                           'multiplication.')
+
+    return A * B
+
+def concatenate(arrays):
+    return tf.stack(arrays, axis = -3)
```
janden: No spaces around `=`.

kymatio/scattering2d/frontend/tensorflow_frontend.py
```diff
@@ -0,0 +1,38 @@
+__all__ = ['Scattering2DTensorflow']
```
janden: Move to end of module.

kymatio/scattering2d/frontend/tensorflow_frontend.py
```diff
@@ -0,0 +1,38 @@
...
+    def __init__(self, J, shape, L=8, max_order=2, pre_pad=False, backend='tensorflow', name='Scattering2D'):
+        ScatteringTensorFlow.__init__(self, name)
+        ScatteringBase2D.__init__(self, J, shape, L, max_order, pre_pad, backend)
+        ScatteringBase2D._instantiate_backend(self, 'kymatio.scattering2d.backend.')
+        ScatteringBase2D.build(self)
+        ScatteringBase2D.create_filters(self)
+
+    def scattering(self, input):
+
+
```
janden: What's with all these blank lines?

kymatio/scattering2d/frontend/tensorflow_frontend.py
```diff
@@ -0,0 +1,38 @@
...
+            signal_shape = input.shape[-2:]
+
+            input = tf.reshape(input, (-1,) + signal_shape)
+
+            S = scattering2d(input, self.pad, self.unpad, self.backend, self.J, self.L, self.phi, self.psi, self.max_order)
+            scattering_shape = tuple(S.shape[-3:])
+            S = tf.reshape(S, batch_shape + scattering_shape)
+            return S
+
+
```
janden: Again, blank lines.

kymatio/scattering2d/frontend/tensorflow_frontend.py
```diff
@@ -0,0 +1,38 @@
...
+
+
+    @tf.Module.with_name_scope
+    def __call__(self, x):
+        return self.scattering(x)
+
```
janden: Remove blank line.

kymatio/scattering2d/tests/test_tensorflow_scattering2d.py
janden: > I'm not much familiar with it

Is anyone around here? I'm serious. Do we know what we're doing?

kymatio/scattering2d/tests/test_tensorflow_scattering2d.py
janden: Cheat? No I don't think so. Both Torch and TF only allow multiplication (and addition, etc.) of arrays that have the same dtype. NP does not, so we get away with certain things. If you look at the corresponding Torch test, it also has it in single precision (although it's not explicit since it's the default precision for random numbers in Torch).

kymatio/scattering2d/tests/test_tensorflow_scattering2d.py
```diff
@@ -0,0 +1,366 @@
...
+        assert not np.iscomplexobj(z)
+        assert np.allclose(np.real(y), z)
+
+
+    @pytest.mark.parametrize('backend', backends)
+    def test_fft_exceptions(self, backend):
+        with pytest.raises(RuntimeError) as record:
+            backend.fft(np.empty((2, 2)), direction='C2R',
+                        inverse=False)
+        assert 'done with an inverse' in record.value.args[0]
```
janden: Two blank lines after this.

kymatio/scattering2d/tests/test_tensorflow_scattering2d.py
janden: Wait, why is this class repeated?

kymatio/scattering2d/tests/test_tensorflow_scattering2d.py
```diff
@@ -0,0 +1,366 @@
...
+
+        w = backend.concatenate((x, y, z))
+
+        assert w.shape == (x.shape[0],) + (3,) + (x.shape[-2:])
+        assert np.allclose(w[:, 0, ...], x)
+        assert np.allclose(w[:, 1, ...], y)
+        assert np.allclose(w[:, 2, ...], z)
+
+
+class TestScattering2DTensorflow:
```
janden: `Tensorflow` → `TensorFlow`.

kymatio/scattering2d/frontend/tensorflow_frontend.py
edouardoyallon: So it seems there are two modes, one which is similar to `torch`, namely the `eager mode` and one for which you need to declare some `tf.placeholder`. I never really used `tf`, thus I'm not sure what is the most standard way to develop. We'd need some tf users' feedbacks to improve this part of the software

kymatio/scattering2d/frontend/tensorflow_frontend.py
edouardoyallon: remove

kymatio/scattering2d/frontend/tensorflow_frontend.py
edouardoyallon: to fix

kymatio/scattering2d/frontend/tensorflow_frontend.py
edouardoyallon: remove

kymatio/scattering2d/frontend/tensorflow_frontend.py
edouardoyallon: remove

kymatio/scattering2d/tests/test_tensorflow_scattering2d.py
edouardoyallon: agreed

kymatio/scattering2d/tests/test_tensorflow_scattering2d.py
edouardoyallon: agreed

kymatio/scattering2d/backend/tensorflow_backend.py
edouardoyallon: ok.

kymatio/scattering2d/backend/tensorflow_backend.py
edouardoyallon: @MuawizChaudhary ?

kymatio/scattering2d/backend/tensorflow_backend.py
edouardoyallon: ok.

kymatio/scattering2d/backend/tensorflow_backend.py
edouardoyallon: 64 everywhere.

kymatio/scattering2d/backend/tensorflow_backend.py
edouardoyallon: let me know if this is ok

kymatio/scattering2d/backend/tensorflow_backend.py
edouardoyallon: should be fine.

kymatio/scattering2d/backend/tensorflow_backend.py
edouardoyallon: @MuawizChaudhary you want or you don't want

kymatio/scattering2d/frontend/tensorflow_frontend.py
edouardoyallon: thoughts?

kymatio/scattering2d/tests/test_tensorflow_scattering2d.py
edouardoyallon: I find it a bit weird to try numpy tensors here but I'd like to have a fresh opinion from a tf user. I don't know any (👀)

kymatio/scattering2d/tests/test_tensorflow_scattering2d.py
edouardoyallon: ok.

kymatio/scattering2d/tests/test_tensorflow_scattering2d.py
edouardoyallon: fixed

edouardoyallon:
The duplication of the tf test classes overloaded the one which was not passing.

kymatio/scattering2d/backend/tensorflow_backend.py
```diff
@@ -33,7 +33,7 @@ def __call__(self, x):
         if self.pre_pad:
             return x
         else:
-            paddings = [[0, 0]]*len(x.shape[:-2])
+            paddings = 2 * [[0, 0]]
```
edouardoyallon: Actually, this was a bad idea: this is simply adding the necessary 0 for other dimensions. Gonna revert.

edouardoyallon:
Types tests are missing, beyond few questions to solve, this is almost ready to merge.

edouardoyallon:
Travis:

```
kymatio/scattering2d/tests/test_tensorflow_scattering2d.py::TestScattering2DTensorFlow::test_Scattering2D[backend]
  /home/travis/build/kymatio/kymatio/kymatio/frontend/tensorflow_frontend.py:5: SyntaxWarning: assertion is always true, perhaps remove parentheses?
    assert (version.parse(tf.__version__) >= version.parse('2.0.0a0'),
kymatio/scattering2d/tests/test_tensorflow_scattering2d.py::TestScattering2DTensorFlow::test_Scattering2D[backend]
  /home/travis/miniconda/envs/test-environment/lib/python3.7/site-packages/tensorflow_core/python/training/tracking/data_structures.py:720: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated since Python 3.3,and in 3.9 it will stop working
    if not isinstance(wrapped_dict, collections.Mapping):
```

lostanlen:
@edouardoyallon 

```
kymatio/scattering2d/tests/test_tensorflow_scattering2d.py::TestScattering2DTensorFlow::test_Scattering2D[backend]
  /home/travis/build/kymatio/kymatio/kymatio/frontend/tensorflow_frontend.py:5: SyntaxWarning: assertion is always true, perhaps remove parentheses?
    assert (version.parse(tf.__version__) >= version.parse('2.0.0a0'),
```

seems like we are asserting on a tuple because of the parentheses and comma.
But i can't find this assert in the PR. There is no trace of a `version.parse`

edouardoyallon:
It's there: https://github.com/kymatio/kymatio/blob/kymatio-v2/kymatio/frontend/tensorflow_frontend.py#L5-L7 but that should be done accordingly to our choice for travis(I guess)

kymatio/scattering2d/tests/test_tensorflow_scattering2d.py
janden: Ok. Will see if I can trick someone into looking at it. I think it's fine to merge, but before we go to `master`, someone should have looked at it and made sure it's ok.

kymatio/scattering2d/backend/tensorflow_backend.py
janden: :+1:

kymatio/scattering2d/backend/tensorflow_backend.py
janden: Ah ok. With batch reshaping happening in the frontend, though, it should be a problem to assume everything has four dimensions (batch, channel, x, y).

janden:
Tests pass. Will rebase tomorrow and merge unless anyone has problems with it.

kymatio/scattering2d/backend/tensorflow_backend.py
MuawizChaudhary: Not used.

kymatio/scattering2d/backend/tensorflow_backend.py
MuawizChaudhary: No, this is good. Just a positive comment :)

kymatio/scattering2d/frontend/tensorflow_frontend.py
MuawizChaudhary: I can ask one of my research group mates, however they haven't used TF much in a while. Everyone in my lab uses pytorch.

kymatio/scattering2d/tests/test_tensorflow_scattering2d.py
MuawizChaudhary: I'm familiar with TF at a basic level, it was used for the labs/programming assignments for my universities machine learning classes, however I've used Pytorch so much more in general.

I'd like to think that if we got this down on the tf front/backend (we're getting the same numbers as the torch end) we did some stuff right!

edouardoyallon:
Does that sound good if one removes:

https://github.com/kymatio/kymatio/blob/kymatio-v2/kymatio/frontend/tensorflow_frontend.py#L5-L7

? (then let's merge)

janden:
> Does that sound good if one removes:
> 
> https://github.com/kymatio/kymatio/blob/kymatio-v2/kymatio/frontend/tensorflow_frontend.py#L5-L7
> 
> ? (then let's merge)

Why? It's just a warning.

edouardoyallon:
> > Does that sound good if one removes:
> > https://github.com/kymatio/kymatio/blob/kymatio-v2/kymatio/frontend/tensorflow_frontend.py#L5-L7
> > ? (then let's merge)
> 
> Why? It's just a warning.

Right now, it's an assert. And we obtain a arning from travis that this "assert" is trivial. I think it would be good indeed it actually launches a warning.

janden:
> Right now, it's an assert. And we obtain a arning from travis that this "assert" is trivial. I think it would be good indeed it actually launches a warning.

I see. Yes, let's make it an if statement and then a warning, that makes more sense.

kymatio/scattering2d/backend/tensorflow_backend.py
edouardoyallon: right. Let's keep it, as it is used by other backends. Are you guys ok with this?

kymatio/scattering2d/frontend/tensorflow_frontend.py
edouardoyallon: would be great. It could be done later

janden: Ok.

kymatio/scattering2d/backend/tensorflow_backend.py
janden: Ok.

janden:
> I see. Yes, let's make it an if statement and then a warning, that makes more sense.

On second thought, why don't we just list it as an explicit requirement? If we check the TF version here, why not NP, Torch, etc.? I suggest we remove this.

edouardoyallon:
> > I see. Yes, let's make it an if statement and then a warning, that makes more sense.
> 
> On second thought, why don't we just list it as an explicit requirement? If we check the TF version here, why not NP, Torch, etc.? I suggest we remove this.

<o/

janden:
On the other hand, the testing issues I was having myself were due to running TF 1.14 instead of TF 2.0, so...

edouardoyallon:
> On the other hand, the testing issues I was having myself were due to running TF 1.14 instead of TF 2.0, so...

A torch 0.1 user would encounter the same issues. I think a warning doesn't kill.

edouardoyallon:
I propose as follow: let's remove it in this PR, and let's open an issue.

janden:
> I think a warning doesn't kill.

Yes but it does prevent us from getting 100% coverage (assuming we don't do a separate test case for environments without the proper version of TF installed).

janden:
> I propose as follow: let's remove it in this PR, and let's open an issue.

:+1:

edouardoyallon:
meeeeeeeeeeeeeeeeeeeerge 🎉

janden:
> meeeeeeeeeeeeeeeeeeeerge tada

Yes just need some time to resolve conflicts and clean up the commits. Should be done later today.

kymatio/scattering2d/tests/test_tensorflow_scattering2d.py
```diff
@@ -0,0 +1,314 @@
+import os
+import io
+import tensorflow as tf
+from kymatio.scattering2d import Scattering2D
+import torch
```
lostanlen: i'm pretty sure that `import torch` is unnecessary here, amirite? :D

edouardoyallon: it is, because of torch.load() - cf #419

lostanlen: oh, ok, nvm; thanks for clarifying!

PR #466, opened 2019-12-04
DO NOT MERGE Torch squeeze 

MuawizChaudhary:
This PR refactors squeeze from the core file into a corresponding torch backend function. The new squeeze function takes in as input the big array of scattering coefficents, the current set of scattering coefficents,  and the current order, and outputs the big array with the current set of scattering coefficents added to a specific position.

### DIFF
--- a/kymatio/scattering2d/backend/torch_backend.py
+++ b/kymatio/scattering2d/backend/torch_backend.py
@@ -315,7 +315,11 @@ def empty_like(x, shape):
     # something more standard.
     return x.new(*shape)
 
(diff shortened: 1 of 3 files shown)


edouardoyallon:
Given that this PR is purely motivated by the tf PR, i'd have preferred to see all the corresponding commits in the TF pr. Now, we have to wait to merge torch, numpy before being able to work on it and we're not sure no further modifications will be needed.

edouardoyallon:
this should be added in the tf branch for sure

PR #468, opened 2019-12-06
REF Reintroduce lists of coefficients in 2D core

janden:
This will simplify the TF frontend, since TF doesn't allow item assignment in `Tensor`s. As a result, we must construct list of `Tensor`s beforehand and concatenate them at the end. This implementation is slightly less efficient in NP and Torch, but seems to be the most straightforward way of implementing it in TF.

### DIFF
--- a/kymatio/scattering2d/backend/torch_backend.py
+++ b/kymatio/scattering2d/backend/torch_backend.py
@@ -321,7 +321,12 @@ def empty_like(x, shape):
     # something more standard.
     return x.new(*shape)
 
-backend = namedtuple('backend', ['name', 'cdgmm', 'modulus', 'subsample_fourier', 'fft', 'Pad', 'unpad', 'empty_like'])
+
+def concatenate(arrays):
+    return torch.stack(arrays, axis=-3)
+
+
+backend = namedtuple('backend', ['name', 'cdgmm', 'modulus', 'subsample_fourier', 'fft', 'Pad', 'unpad', 'empty_like', 'concatenate'])
 backend.name = 'torch'
 backend.cdgmm = cdgmm
 backend.modulus = Modulus()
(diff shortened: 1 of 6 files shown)


kymatio/scattering2d/core/scattering2d.py
```diff
@@ -73,10 +68,15 @@ def scattering2d(x, pad, unpad, backend, J, L, phi, psi, max_order):
...
             S_2_r = unpad(S_2_r)
 
-            S[..., n_order2, :, :] = S_2_r
-            n_order2 += 1
+            out_S_2.append(S_2_r)
+
+    out_S = []
+    out_S.extend(out_S_0)
+    out_S.extend(out_S_1)
+    out_S.extend(out_S_2)
```
MuawizChaudhary: This passes the tests yes?

kymatio/scattering2d/backend/torch_backend.py
```diff
@@ -321,7 +321,12 @@ def empty_like(x, shape):
     # something more standard.
     return x.new(*shape)
 
-backend = namedtuple('backend', ['name', 'cdgmm', 'modulus', 'subsample_fourier', 'fft', 'Pad', 'unpad', 'empty_like'])
+
+def concatenate(arrays):
+    return torch.stack(arrays, axis=-3)
```
MuawizChaudhary: This seems so much simpler than what we were doing beforehand!

kymatio/scattering2d/core/scattering2d.py
janden: It does locally, but it's always good to check on Jenkins/Travis.

kymatio/scattering2d/backend/torch_backend.py
janden: Indeed it is.

janden:
> Mayhaps it'd be a good idea to have a corresponding test in the torch test suite, but I'm sure its fine.

Done.

edouardoyallon:
Merge when tests pass

PR #470, opened 2019-12-06
MAINT fft

### LINKED ISSUE #386: Inverse fft in 2D backend

edouardoyallon:
Close #386

### DIFF
--- a/kymatio/scattering2d/tests/test_numpy_scattering2d.py
+++ b/kymatio/scattering2d/tests/test_numpy_scattering2d.py
@@ -149,9 +149,13 @@ class TestFFT:
 
         z = backend.fft(x, direction='C2C', inverse=True)
 
+        z = z * 4
+
         assert np.allclose(y, z)
 
         z = backend.fft(x, direction='C2R', inverse=True)
+
+        z = z * 4
 
         assert not np.iscomplexobj(z)
         assert np.allclose(np.real(y), z)
--- a/kymatio/scattering2d/backend/numpy_backend.py
+++ b/kymatio/scattering2d/backend/numpy_backend.py
@@ -138,10 +138,10 @@ def fft(x, direction='C2C', inverse=False):
             raise RuntimeError('C2R mode can only be done with an inverse FFT.')
 
     if direction == 'C2R':
-        output = np.real(np.fft.ifft2(x)) * x.shape[-1] * x.shape[-2]
+        output = np.real(np.fft.ifft2(x))
     elif direction == 'C2C':
         if inverse:
-            output = np.fft.ifft2(x) * x.shape[-1] * x.shape[-2]
+            output = np.fft.ifft2(x)
         else:
             output = np.fft.fft2(x)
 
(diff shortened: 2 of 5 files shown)


edouardoyallon:
> Before we normalized, now we do not. How come?

So - this will be a more standard fft, even if it doesn't satisfy fft o ifft = Id. I think it's more standard this way, than before.

@janden view?

kymatio/scattering2d/tests/test_numpy_scattering2d.py
```diff
@@ -149,10 +149,14 @@ def test_fft(self, backend):
 
         z = backend.fft(x, direction='C2C', inverse=True)
 
+        z = z * 4
```
lostanlen: i would prefer if you didn't update the z between the definition and the test

lostanlen:
`ifft(fft(x))` is an absolute must

edouardoyallon:
> > So - this will be a more standard fft, even if it doesn't satisfy fft o ifft = Id. I think it's more standard this way, than before.
> 
> It does satisfy `ifft(fft(x)) = x`, as far as I can tell. Why wouldn't it? It's just not unitary.

?? No, I had to renormalize by "4", cf test

kymatio/scattering2d/tests/test_numpy_scattering2d.py
edouardoyallon: @janden see

janden:
> ?? No, I had to renormalize by "4", cf test

When you test the `ifft`, yes. In that test, `y` is the DFT of `x`. Since this is a 2×2 DFT, it also happens to be `4*ifft(x)`, so we simplify the test by comparing to that. Might be less confusing to simply take `ifft(y)` and compare it to `x`.

edouardoyallon:
> > ?? No, I had to renormalize by "4", cf test
> 
> When you test the `ifft`, yes. In that test, `y` is the DFT of `x`. Since this is a 2×2 DFT, it also happens to be `4*ifft(x)`, so we simplify the test by comparing to that. Might be less confusing to simply take `ifft(y)` and compare it to `x`.

damn, I read too quickly - then in this case, there is an inconsistency in the normalization of the filters.. I'm ok to leave as it is.

janden:
> I'm ok to leave as it is.

You mean close the PR? Why? Can't we just properly normalize the filters?

edouardoyallon:
> > I'm ok to leave as it is.
> 
> You mean close the PR? Why? Can't we just properly normalize the filters?

Sorry, it's late here, I'm alwaysa  bit confused by the normalization linked to the downsmapling, it's actually all good, sorry for the mess

PR #482, opened 2020-01-16
Fix and homogenize subsampling description of SubsampleFourier docstr…

anakin-datawalker:
…ings within scattering2d backend modules (change 2**k by k)

Close #482


edouardoyallon:
Look good.

PR #483, opened 2020-01-17
Multi GPU kymatio

anakin-datawalker:
Small changes for multi GPU. Tested on cifar.py function.

Seems to work 90% of the time for 2 GPUs, roughly 20-30% for 4 GPUs. Otherwise segmentation fault.

### DIFF
--- a/examples/2d/cifar.py
+++ b/examples/2d/cifar.py
@@ -139,7 +139,7 @@ def main():
     model = Scattering2dCNN(K, args.classifier)
 
     if use_cuda:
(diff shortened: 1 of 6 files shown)


lostanlen:
thanks! how come we observe such a big decrease in test coverage?

anakin-datawalker:
It seems to be coming from files not affected by the pull request so no idea why such change has occurred...

edouardoyallon:
@janden is the GPU server from the FI down?

edouardoyallon:
for working on this one, we need a travis script running in a multi gpu setting, @janden thoughts?

edouardoyallon:
given my msg in the issue, this can't work!

PR #499, opened 2020-01-23
1D np frontend

edouardoyallon:


edouardoyallon:
@janden if you feel to finish it, i think it's almost done

PR #526, opened 2020-01-26
unify benchmarks

### LINKED ISSUE #273: DOC 3D benchmarks
Currently none are listed in the user's guide. We should probably include something here.

### LINKED ISSUE #359: Normalize benchmarks
Right now, the different `compute_speed.py` scripts have very different running times. We should normalize these so that they take about the same time. Of course, this has to be balanced with ensuring large enough batch sizes to ensure maximum efficiency.

lostanlen:
Closes #359

### DIFF
--- a/benchmarks/benchmarks/common.py
+++ b/benchmarks/benchmarks/common.py
@@ -1,2 +1,9 @@
 import torch
+
+
+# Make sure we have the same random numbers for each run.
 torch.manual_seed(0)
+
+# This constant corresponds to the total number of samples per batch. It is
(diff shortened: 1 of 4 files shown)


benchmarks/benchmarks/common.py
```diff
@@ -1,2 +1,6 @@
 import torch
 torch.manual_seed(0)
+
+# This constant corresponds to the total number of samples per batch. It is
+# shared across all benchmarks of scattering: 1D, 2D, and 3D.
+SCATTERING_BENCHMARK_SIZE = 2**21
```
janden: Spaces around `**`.

PR #535, opened 2020-01-27
ENH Add sklearn tranformer and example

### LINKED ISSUE #277: ENH scikit-learn-compatible transformer
Add a mixin or wrapper that can use the scattering transform in the context of sklearn:

- Instantiate transformer with image shape
- images are provided as `X`, probably raveled, `X.shape = n_samples, n_features`
- output is also raveled, so that it can go to the next part of a pipeline

eugenium:
Closes #277 .

### DIFF
--- a/examples/2d/plot_sklearn.py
+++ b/examples/2d/plot_sklearn.py
@@ -1,0 +1,40 @@
+"""
+Scikit-learn transformer example
+================================
+Here we demonstrate a simple application of scattering as a transformer
+"""
+
+
+from kymatio.sklearn import Scattering2D
+from sklearn.pipeline import Pipeline
+from sklearn.model_selection import train_test_split
+from sklearn import datasets
+from sklearn.linear_model import LogisticRegression
+from sklearn.metrics import accuracy_score
+
+
+# Create a scattering object (can be Scattering1D or Scattering3D as well)
+S = Scattering2D(shape=(8, 8), J=1)
+
+# Use the toy digits dataset (8x8 digits)
+digits = datasets.load_digits()
+
+# Split data into train and test subsets
+images = digits.images
+images = images.reshape((images.shape[0], -1))
+x_train, x_test, y_train, y_test = train_test_split(images, digits.target, test_size=0.5, shuffle=False)
+
+# Create the scikit-learn transformer
+
+# Create a classifier: a support vector classifier
+classifier = LogisticRegression()
+estimators = [('scatter', S), ('clf', classifier)]
+pipeline = Pipeline(estimators)
+
+# We learn the digits on the first half of the digits
+pipeline.fit(x_train, y_train)
+
+# Now predict the value of the digit on the second half:
+y_pred = pipeline.predict(x_test)
+
+print(accuracy_score(y_test, y_pred))
--- a/examples/2d/plot_sklearn.py
+++ b/examples/2d/plot_sklearn.py
@@ -4,37 +4,74 @@ Scikit-learn transformer example
 Here we demonstrate a simple application of scattering as a transformer
 """
 
+###############################################################################
+# Preliminaries
+# -------------
+#
+# Import the relevant classes and functions from sciki-learn.
 
-from kymatio.sklearn import Scattering2D
 from sklearn.pipeline import Pipeline
 from sklearn.model_selection import train_test_split
 from sklearn import datasets
 from sklearn.linear_model import LogisticRegression
 from sklearn.metrics import accuracy_score
 
+###############################################################################
+# Import the scikit-learn `Scattering2D` frontend.
 
-# Create a scattering object (can be Scattering1D or Scattering3D as well)
-S = Scattering2D(shape=(8, 8), J=1)
+from kymatio.sklearn import Scattering2D
 
-# Use the toy digits dataset (8x8 digits)
+###############################################################################
+# Preparing the data
+# ------------------
+#
+# First, we load the dataset. In this case, it's the UCI ML digits dataset
+# included with scikit-learn, consisting of 8×8 images of handwritten digits
+# from one to ten.
+
 digits = datasets.load_digits()
 
-# Split data into train and test subsets
+###############################################################################
+# We then extract the images, reshape them to an array of size `(n_features,
+# n_samples)` needed for processing in a scikit-learn pipeline.
+
 images = digits.images
 images = images.reshape((images.shape[0], -1))
-x_train, x_test, y_train, y_test = train_test_split(images, digits.target, test_size=0.5, shuffle=False)
 
-# Create the scikit-learn transformer
+###############################################################################
+# We then split the images (and their labels) into a train and a test set.
 
-# Create a classifier: a support vector classifier
+x_train, x_test, y_train, y_test = train_test_split(images, digits.target,
+                                                    test_size=0.5, shuffle=False)
+
+###############################################################################
+# Training and testing the model
+# ------------------------------
+# Create a `Scattering2D` object, which implements a scikit-learn
+# `Transformer`. We set the input shape to match that of the the images (8×8)
+# and the averaging scale is set to `J = 1`, which means that the local
+# invariance is `2 ** 1 = 1`.
+
+S = Scattering2D(shape=(8, 8), J=1)
+
+###############################################################################
+# We then plug this into a scikit-learn pipeline which takes the scattering
+# features and provides them to a `LogisticRegression` classifier.
+
 classifier = LogisticRegression()
 estimators = [('scatter', S), ('clf', classifier)]
 pipeline = Pipeline(estimators)
 
-# We learn the digits on the first half of the digits
+###############################################################################
+# Given the pipeline, we train it on `(x_train, y_train)` using
+# `pipelien.fit`.
+
 pipeline.fit(x_train, y_train)
 
-# Now predict the value of the digit on the second half:
+###############################################################################
+# Finally, we calculate the predicted labels on the test data and output the
+# classification accuracy.
+
 y_pred = pipeline.predict(x_test)
 
-print(accuracy_score(y_test, y_pred))
+print('Accuracy:', accuracy_score(y_test, y_pred))
--- a/kymatio/sklearn.py
+++ b/kymatio/sklearn.py
@@ -1,0 +1,45 @@
+from sklearn.base import BaseEstimator, TransformerMixin
+
+from kymatio.numpy import Scattering1D as ScatteringNumPy1D
+from kymatio.numpy import Scattering2D as ScatteringNumPy2D
+from kymatio.numpy import HarmonicScattering3D as HarmonicScatteringNumPy3D
+
+
+class ScatteringTransformerMixin(BaseEstimator, TransformerMixin):
+    def fit(self, x=None, y=None):
+        # No fitting necessary.
+        return self
+
+    def predict(self, x):
+        n_samples = x.shape[0]
+
+        x = x.reshape((-1,) + self.shape)
+
+        Sx = self.scattering(x)
+
+        Sx = Sx.reshape(n_samples, -1)
+
+        return Sx
+
+    transform = predict
+
+
+class ScatteringTransformer1D(ScatteringNumPy1D, ScatteringTransformerMixin):
+    pass
+
+
+class ScatteringTransformer2D(ScatteringNumPy2D, ScatteringTransformerMixin):
+    pass
+
+
+class HarmonicScatteringTransformer3D(HarmonicScatteringNumPy3D,
+                                      ScatteringTransformerMixin):
+    pass
+
+
+Scattering1D = ScatteringTransformer1D
+Scattering2D = ScatteringTransformer2D
+HarmonicScattering3D = HarmonicScatteringTransformer3D
+
+
+__all__ = ['Scattering1D', 'Scattering2D', 'HarmonicScattering3D']
--- a/kymatio/tests/test_sklearn.py
+++ b/kymatio/tests/test_sklearn.py
@@ -1,0 +1,29 @@
+import os
+import io
+import numpy as np
+
+from kymatio.sklearn import Scattering2D as ScatteringTransformer2D
+from kymatio.numpy import Scattering2D as ScatteringNumPy2D
+
+def test_sklearn_transformer():
+    test_data_dir = os.path.join(os.path.dirname(__file__), "..",
+                                 "scattering2d", "tests")
+
+    with open(os.path.join(test_data_dir, 'test_data_2d.npz'), 'rb') as f:
+        buf = io.BytesIO(f.read())
+        data = np.load(buf)
+
+    x = data['x']
+    J = data['J']
+
+    S = ScatteringNumPy2D(J, x.shape[2:])
+    Sx = S.scattering(x)
+
+    x_raveled = x.reshape(x.shape[0], -1)
+    Sx_raveled = Sx.reshape(x.shape[0], -1)
+
+    st = ScatteringTransformer2D(J, x.shape[2:])
+
+    t = st.transform(x_raveled)
+
+    assert np.allclose(Sx_raveled, t)


kymatio/utils.py
```diff
@@ -0,0 +1,57 @@
...
+        ======
+        Y: ndarray, shape (n_samples, n_scattering_features)
+            The scattering coefficients, raveled in C order
+        """
+
+        self.scattering = S
+        self.signal_shape = signal_shape
+        self.frontend = frontend
+
+        assert frontend in ['numpy','torch']
```
MuawizChaudhary: Why not Tensorflow?

kymatio/utils.py
```diff
@@ -0,0 +1,57 @@
...
+        transformed = self.scattering(X_reshaped)
+
+        if self.frontend is 'torch':
+            output = transformed.detach().cpu().numpy().reshape(X.shape[0], -1)
+        else:
+            output = transformed.reshape(X.shape[0], -1)
+
+        return output
+
+    transform = predict
```
MuawizChaudhary: What is this line doing here?

eugenium: the tranformer needs to define a transform function

kymatio/utils.py
eugenium: fixed

kymatio/utils.py
```diff
@@ -0,0 +1,57 @@
...
+        ======
+        Y: ndarray, shape (n_samples, n_scattering_features)
+            The scattering coefficients, raveled in C order
+        """
+
+        self.scattering = S
+        self.signal_shape = signal_shape
+        self.frontend = frontend
+
+        assert frontend in ['numpy','torch','tensorflow']
```
MuawizChaudhary: I believe you need to use the tensor-flow reshape op, rather than `.reshape`

eugenium: we decided to remove tf compatability for this transformer

edouardoyallon:
should we reaname this `utils.py` to something like `sklearn.py` ?

examples/2d/plot_sklearn.py
```diff
@@ -0,0 +1,43 @@
+"""
+ScikitLearn Transformer
```
lostanlen: scikit-learn

examples/2d/plot_sklearn.py
```diff
@@ -0,0 +1,43 @@
...
+from sklearn.model_selection import train_test_split
+from sklearn import datasets
+from sklearn.linear_model import LogisticRegression
+
+# Use numpy or torch for the scattering
+frontend = 'numpy'
+
+
+# Create a scattering object (can be Scattering1D or Scattering3D as well)
+S = Scattering2D(shape = (8, 8), J = 1, frontend = frontend)
```
lostanlen: this example is not very realistic. consider a CIFAR-sized image with J>1

lostanlen: also, i don't think that you need space around the equal sign for kwargs

examples/2d/plot_sklearn.py
```diff
@@ -0,0 +1,43 @@
...
+
+# Use the toy digits dataset (8x8 digits)
+digits = datasets.load_digits()
+
+# Split data into train and test subsets
+X_train, X_test, y_train, y_test = train_test_split(
+    digits.images.reshape((len(digits.images), -1)), digits.target, test_size=0.5, shuffle=False)
+
+# Create the scikitlearn transformer
+st = ScatteringTransformer(S, digits.images[0].shape,frontend)
```
lostanlen: space after comma

examples/2d/plot_sklearn.py
```diff
@@ -0,0 +1,43 @@
...
+
+# Create a classifier: a support vector classifier
+classifier = LogisticRegression()
+estimators = [('scatter', st), ('clf', classifier)]
+pipe = Pipeline(estimators)
```
lostanlen: pipe -> pipeline

pipe makes is sound like we're talking about a UNIX pipeline

kymatio/tests/test_utils.py
```diff
@@ -0,0 +1,39 @@
...
+
+from kymatio.utils import ScatteringTransformer
+from kymatio.scattering2d import Scattering2D
+import io
+import numpy as np
+from numpy.testing import assert_array_almost_equal
+from sklearn.utils.estimator_checks import check_estimator
+
+def test_sklearn_transformer():
+    test_data_dir = os.path.join(os.path.dirname(__file__),
```
lostanlen: weird indentation

kymatio/utils.py
```diff
@@ -0,0 +1,53 @@
...
+
+        signal_shape: tuple of ints
+            The shape of one sample. Scikit-learn convention is to work with
+            2D matrices only, of shape (n_samples, n_features). Data is
+            delivered in this way and has to be reshaped before transform
+
+        Output
+        ======
+        Y: ndarray, shape (n_samples, n_scattering_features)
+            The scattering coefficients, raveled in C order
```
lostanlen: C order -> row-major order (C-like)

examples/2d/plot_sklearn.py
eugenium: we have lots of CIFAR and MNIST examples, but these take a long time to run, I think for this one it make sense if its one of the clean and fast running examples

kymatio/utils.py
```diff
@@ -0,0 +1,53 @@
+from sklearn.base import BaseEstimator, TransformerMixin
```
janden: Maybe makes more sense to name this module `sklearn` instead of `utils`?

kymatio/utils.py
```diff
@@ -0,0 +1,53 @@
...
+        """Creates an object that is compatible with the scikit-learn API
+        and implements the .transform method.
+        
+        Parameters
+        ==========
+
+        S: an instance of Scattering1D, Scattering2D or Scattering3D
+            This instance is called by the transformer
+
+        signal_shape: tuple of ints
```
janden: Can't this be acquired from `S`?

kymatio/utils.py
```diff
@@ -0,0 +1,53 @@
...
+
+class ScatteringTransformer(BaseEstimator, TransformerMixin):
+    def __init__(self, S, signal_shape, frontend='numpy'):
+        """Creates an object that is compatible with the scikit-learn API
+        and implements the .transform method.
+        
+        Parameters
+        ==========
+
+        S: an instance of Scattering1D, Scattering2D or Scattering3D
```
janden: Why can't this be fixed by the class itself? In other words, we'd have `Scattering1D`, `Scattering2D`, etc., just like we have for Keras?

kymatio/utils.py
```diff
@@ -0,0 +1,53 @@
+from sklearn.base import BaseEstimator, TransformerMixin
+import torch
```
janden: Will fail on NP-only installs.

kymatio/utils.py
```diff
@@ -0,0 +1,53 @@
...
+
+        Output
+        ======
+        Y: ndarray, shape (n_samples, n_scattering_features)
+            The scattering coefficients, raveled in C order
+        """
+
+        self.scattering = S
+        self.signal_shape = signal_shape
+        self.frontend = frontend
```
janden: Again, shouldn't this be inferred from `S`?

examples/2d/plot_sklearn.py
```diff
@@ -0,0 +1,43 @@
...
+
+# We learn the digits on the first half of the digits
+pipe.fit(X_train, y_train)
+
+# Now predict the value of the digit on the second half:
+y_pred = pipe.predict(X_test)
+from sklearn.metrics import accuracy_score
+print(accuracy_score(y_test, y_pred))
```
janden: Newline.

examples/2d/plot_sklearn.py
janden: I agree. It's better to have smaller examples in general.

examples/2d/plot_sklearn.py
```diff
@@ -0,0 +1,43 @@
+"""
```
janden: Needs to be formatted for the gallery like the other examples.

examples/2d/plot_sklearn.py
```diff
@@ -0,0 +1,43 @@
...
+classifier = LogisticRegression()
+estimators = [('scatter', st), ('clf', classifier)]
+pipe = Pipeline(estimators)
+
+# We learn the digits on the first half of the digits
+pipe.fit(X_train, y_train)
+
+# Now predict the value of the digit on the second half:
+y_pred = pipe.predict(X_test)
+from sklearn.metrics import accuracy_score
```
janden: Move imports to top.

examples/2d/plot_sklearn.py
edouardoyallon: :+1:

kymatio/utils.py
edouardoyallon: Yes. Definintely. It should be a sklearn file, and we should do from kymatio.sklearn import ...

kymatio/utils.py
edouardoyallon: I agree!

kymatio/utils.py
```diff
@@ -0,0 +1,59 @@
...
+        return self
+
+    def predict(self, X):
+        if self.frontend == 'torch':
+            import torch
+
+        X_reshaped = X.reshape((-1,) + self.signal_shape)
+
+        if self.frontend is 'torch':
+            X_reshaped = torch.from_numpy(X_reshaped)
```
edouardoyallon: if we're "frontend" independent, then, one should transform into np array any "type" of data? Maybe it's safer to be purely numpy compatible first? @janden

kymatio/utils.py
janden: Yes, and (as you said) it would be good to allow this to be instantiated using the `ScatteringEntry` class. However, this is not possible right now since we don't follow the same conventions (wrapping the `Scattering` object, not having a separate class for each dimension).

kymatio/utils.py
janden: Yes I don't see the why we should have this frontend stuff in here. As far as I can tell, `sklearn` is used with NP. Sure, there are extensions like `skorch` that let you do `sklearn` on `torch`, but this is a separate package.

janden:
> Small modifications to do but this should be definitively incorporated in the pkg

I disagree. We need to decide whether to keep the same (dimension independent) interface, passing the `Scattering` object to the constructor, or making it more compatible with the other frontends, defining separate classes for each dimension and accepting the parameters directly (like in the Keras frontend). The multiple-frontend support also confuses me still. Why do we need this?

edouardoyallon:
> > Small modifications to do but this should be definitively incorporated in the pkg
> 
> I disagree. We need to decide whether to keep the same (dimension independent) interface, passing the `Scattering` object to the constructor, or making it more compatible with the other frontends, defining separate classes for each dimension and accepting the parameters directly (like in the Keras frontend). The multiple-frontend support also confuses me still. Why do we need this?

Given your messages, I thought you wanted `torch` to be compatible with `sklearn`. I think it'd be great and enough to be only `np` compatible.

kymatio/utils.py
eugenium: Having 3 separate transformers might make sense. I do prefer that over havign the use create their own scattering objects, just thought it would look messy

Now about the np-only framework, Maybe I am a bit naive but I think for many cases cpu scatteirng is slow to the point of not being useful, so if we dont expose the gpu acceleration to a sklearn user we might never really get any sklearn users?

kymatio/utils.py
eugenium: your right. But if we are switching to the transformer creating the scattering object this is moot

examples/2d/plot_sklearn.py
eugenium: It is, isnt it? e.g. https://github.com/kymatio/kymatio/blob/master/examples/2d/long_mnist_classify.py

kymatio/utils.py
eugenium: @edouardoyallon  the sklearn user never interacts with torch here except as a dependency, the transformer is np in np out. The default is also numpy so the torch and gpu acceleration are completely optional and only if the user really wants to use them.

If you can name me some real examples that arent just demos where scattering is useful on cpu esp in the context of sklearn it might be helpful to understand what we want from this.

eugenium:
> > Small modifications to do but this should be definitively incorporated in the pkg
> 
> I disagree. We need to decide whether to keep the same (dimension independent) interface, passing the `Scattering` object to the constructor, or making it more compatible with the other frontends, defining separate classes for each dimension and accepting the parameters directly (like in the Keras frontend). The multiple-frontend support also confuses me still. Why do we need this?

I would vote for your proposal that makes a separate transformer for each 1d,2d,3d.

kymatio/utils.py
janden: > gpu acceleration are completely optional and only if the user really wants to use them.

Yeah but how much acceleration do you get if you're moving it onto the GPU, transforming, and then off? Seems like you would spend a lot of time copying to and from the GPU.

> If you can name me some real examples that arent just demos where scattering is useful on cpu esp in the context of sklearn it might be helpful to understand what we want from this.

So all the audio experiments (classification, reconstruction, etc.) were done on CPU in a reasonable amount of time. For classification, it was hooked up to an SVM, which I think could nicely integrate into `sklearn`. Once this is merged, I could code up an example or two on some audio classification problems.

examples/2d/plot_sklearn.py
janden: Don't you need the long line of `#`s to mark off the text portions (instead of having them appear as just comments in the code)? I don't quite remember.

janden:
> Given your messages, I thought you wanted torch to be compatible with sklearn.

Sorry. I wasn't very clear. I was proposing to get rid of the torch stuff. Again, I'm not convinced it would be that much faster if we're transferring onto and off the GPU.

> I would vote for your proposal that makes a separate transformer for each 1d,2d,3d.

Good.

I can get started on making the changes (getting rid of the frontend, creating a separate interface for 1D/2D/3D) later today.

kymatio/utils.py
eugenium: > > gpu acceleration are completely optional and only if the user really wants to use them.
> 
> Yeah but how much acceleration do you get if you're moving it onto the GPU, transforming, and then off? Seems like you would spend a lot of time copying to and from the GPU.

In cases of reasonable sized natural images (e.g. 224x224x3) and larger J it would still give a huge speed up, 1 or 2 orders of magnitude even with data transfer. But ok its true 1d version might be reasonably fast for useful input sizes. 
> 
> > If you can name me some real examples that arent just demos where scattering is useful on cpu esp in the context of sklearn it might be helpful to understand what we want from this.
> 
> So all the audio experiments (classification, reconstruction, etc.) were done on CPU in a reasonable amount of time. For classification, it was hooked up to an SVM, which I think could nicely integrate into `sklearn`. Once this is merged, I could code up an example or two on some audio classification problems.

I'm  still have a worry that we should take care to not expose the casual user (e.g. someone trying to stick it in their sklearn pipeline) to something much slower than what they are used to, or what they get with another transformer. Perhaps we can print warnings or some such when the sizes are relatively large.

examples/2d/plot_sklearn.py
eugenium: aah ok I see. I thought were referring to the top part of the file.

kymatio/utils.py
janden: > larger J it would still give a huge speed up, 1 or 2 orders of magnitude even with data transfer.

Ok that's good to know. In that case, it's something that would be useful more generally for the NP frontend. How about implementing torch (or TF) *backend* for NP that would enable these speedups.

> Perhaps we can print warnings or some such when the sizes are relatively large.

Eh I'm not too keen on this. We can put something in the docs on the NP frontend, but I don't think it's a good idea to warn for large sizes. It's not like calling the NP FFT with large inputs gives you a warning. If you work with large arrays, things are going to take longer (whether you're on GPU or CPU).

janden:
So I made a few changes:
* Only NP frontends now (we will address this later by making the NP frontend compatible with TF and Torch backends).
* `Transformer`/`Estimator` is now coded as a mixin.
* The sklearn class now *inherits* from `Scattering2D` together with the mixin, so there's no wrapping anymore.
* I added code for 1D and harmonic 3D (although I haven't tested this, yet).

edouardoyallon:
should we merge?

examples/2d/plot_sklearn.py
edouardoyallon: I'd like to see the result (on the website) but looks good

kymatio/utils.py
janden: Made an issue about GPU backends for NP. We can discuss adding warnings for large sizes in the issues.

PR #547, opened 2020-01-28
TST Add Python 3.8 to Travis test suite

janden:


janden:
Appparently, neither Torch or TF support 3.8, so this would be a NP-only test. Not worth it at this time.

PR #551, opened 2020-01-28
[DO NOT MERGE] API Refactor code to take advantage of rfft 

### LINKED ISSUE #462: Separate padding from casting to complex
How useful is it to have this keyword argument [pre_pad] though? using the `Pad` class to do anything else than padding seems like a weird design antipattern. If there is really a use case for doing the conversion, consider making a dedicated method. Otherwise, consider removing the functionality.

_Originally posted by @lostanlen in https://github.com/kymatio/kymatio/pull/459_

Padding and moving to complex are very different things. However, both involve significant memory movement (if we're working with large arrays), so a case can be made for combining them. Yet another approach would be to avoid casting to complex in the first place. Indeed, this simply concatenates zeros and slows down further computation (since we're just multiplying by zero there), which may be avoided with a little extra work.

MuawizChaudhary:
This in progress pr refactors the padding and modulus classes to not cast to complex. Instead, the complex casting is taken care of by a framework specific rfft function. This PR also refactors the algorithm to take into account the change in real or complex tensors.  Closes #462.


kymatio/backend/base_backend.py
```diff
@@ -51,6 +56,8 @@ def fft_forward(self, x, direction='C2C', inverse=False):
                 output = self.ifft(x)
             else:
                 output = self.fft(x)
+        elif direction == 'R2C':
```
edouardoyallon: It would make sens, here, to add a test for checking the input is indeed real.

kymatio/backend/torch_backend.py
```diff
@@ -179,6 +179,7 @@ def cdgmm(A, B, inplace=False):
 
     type_checks(A)
 
+    print(A.shape, B.shape)
```
edouardoyallon: ??

kymatio/scattering1d/backend/torch_backend.py
```diff
@@ -132,19 +100,19 @@ def unpad(x, i0, i1):
...
+    lambda x: torch.rfft(x, 1, normalized=False, onesided=False),
     lambda x: torch.ifft(x, 1, normalized=False),
     lambda x: torch.irfft(x, 1, normalized=False, onesided=False),
     type_checks)
 
 
 backend = namedtuple('backend', ['name', 'modulus_complex', 'subsample_fourier', 'real', 'unpad', 'fft', 'concatenate'])
 backend.name = 'torch'
-backend.modulus_complex = Modulus()
+backend.modulus = Modulus()
```
edouardoyallon: 👍

kymatio/scattering1d/core/scattering1d.py
```diff
@@ -77,7 +77,9 @@ def scattering1d(x, pad, unpad, backend, J, psi1, psi2, phi, pad_left=0, pad_rig
     U_0 = pad(x, pad_left=pad_left, pad_right=pad_right)
 
     # compute the Fourier transform
-    U_0_hat = fft(U_0, 'C2C')
+    print(U_0.shape)
```
edouardoyallon: ??

MuawizChaudhary:
I'm very suspicious of the skcuda, the tests all pass, even through there are functions defined in the skcuda backend that are not defined anymore. Can someone chime in and tell me why this is the case?

kymatio/scattering1d/backend/numpy_backend.py
```diff
@@ -35,7 +35,7 @@ def subsample_fourier(x, k):
     return res
 
 
-def pad_1d(x, pad_left, pad_right, mode='constant', value=0.):
+def pad(x, pad_left, pad_right, mode='reflect', value=0.):
```
MuawizChaudhary: should the defaults for `pad_left`, `pad_right` be `=0`?

kymatio/scattering1d/backend/torch_backend.py
```diff
@@ -40,7 +41,7 @@ def subsample_fourier(x, k):
     res = x.view(x.shape[:-2] + (k, N // k, 2)).mean(dim=-3)
     return res
 
-def pad_1d(x, pad_left, pad_right, mode='constant', value=0.):
+def pad(x, pad_left, pad_right, mode='reflect', value=0.):
```
MuawizChaudhary: Why doesn't torch have its own general backed pad primitive?

MuawizChaudhary: Also add defaults for 'pad_left', 'pad_right'.

kymatio/backend/torch_backend.py
```diff
@@ -182,7 +183,7 @@ def cdgmm(A, B, inplace=False):
     if A.shape[-len(B.shape):-1] != B.shape[:-1]:
         raise RuntimeError('The filters are not compatible for multiplication.')
 
-    if A.dtype is not B.dtype:
+    if A.dtype is not B.dtype or type(A) is not type(B):
```
MuawizChaudhary: I had to add this so [this](https://github.com/kymatio/kymatio/pull/551/files#diff-3ee77cd2d2c53e9935b29a827d4f7a0fR123) test would pass. Should we keep one or the other, or both?

kymatio/scattering2d/backend/tensorflow_backend.py
```diff
@@ -84,6 +84,7 @@ def __call__(self, x, k):
 backend.modulus = Modulus()
 backend.subsample_fourier = SubsampleFourier()
 backend.fft = FFT(lambda x: tf.signal.fft2d(x, name='ifft2d'),
```
MuawizChaudhary: Wrong name.

kymatio/scattering3d/backend/numpy_backend.py
```diff
@@ -7,29 +7,13 @@
 
 BACKEND_NAME = 'numpy'
 
+from ...backend.numpy_backend import modulus, cdgmm
+from ...backend.base_backend import FFT
 
 def _iscomplex(x):
     return x.dtype == np.complex64 or x.dtype == np.complex128
```
MuawizChaudhary: Remove.

kymatio/scattering3d/tests/test_torch_scattering3d.py
```diff
@@ -85,54 +95,36 @@ def test_cdgmm3d(device, backend, inplace):
...
 
         with pytest.raises(RuntimeError) as record:
             x = torch.randn((3, 3, 3, 2), device=device)
             y = torch.randn((4, 4, 4, 2), device=device)
             backend.cdgmm3d(x, y)
         assert "not compatible" in record.value.args[0]
 
-        x = torch.randn((2, 3, 3, 3, 2), device=device)
-        y = torch.randn((3, 3, 3, 2), device=device)
-        backend.cdgmm3d(x, y)
```
MuawizChaudhary: Why was this here in the first place like this?

kymatio/scattering3d/tests/test_torch_scattering3d.py
```diff
@@ -85,54 +95,36 @@ def test_cdgmm3d(device, backend, inplace):
...
-        assert "should be complex" in record.value.args[0]
-
-        # This one is a little tricky. We can't have the number of dimensions be
-        # greater than 4 since that triggers the "not compatible" error.
-        with pytest.raises(RuntimeError) as record:
-            x = torch.randn((3, 3, 2), device=device)
-            y = torch.randn((3, 3, 2), device=device)
-            backend.cdgmm3d(x, y)
-        assert "must be simply a complex" in record.value.args[0]
-
```
MuawizChaudhary: Are we fine with these removals?

kymatio/scattering2d/backend/numpy_backend.py
```diff
@@ -96,6 +96,7 @@ def concatenate(arrays):
 backend.modulus = modulus
 backend.subsample_fourier = SubsampleFourier()
 backend.fft = FFT(lambda x:np.fft.fft2(x),
+                  lambda x:np.fft.fft2(x),
                   lambda x:np.fft.ifft2(x),
                   lambda x:np.real(np.fft.ifft2(x)),
                   lambda x:None)
```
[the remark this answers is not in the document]
lostanlen: yes

kymatio/backend/base_backend.py
janden: Yes.

kymatio/backend/torch_backend.py
```diff
@@ -125,19 +124,21 @@ class Modulus():
...
-        type_checks(x)
-
-        norm = torch.zeros_like(x)
-        norm[..., 0] = modulus(x)
+        type_checks_complex(x)
+        norm = modulus(x)
         return norm
 
-def type_checks(x):
+def type_checks_complex(x):
```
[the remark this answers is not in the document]
janden: Yes, but it should call `type_checks` at the end instead of duplicating the code.

kymatio/backend/torch_backend.py
janden: How was it passing before? We should be checking the `dtype`, not the type (since that's influenced by other things).

kymatio/scattering1d/backend/numpy_backend.py
janden: If they were zero before, let's leave them at zero now.

kymatio/scattering1d/backend/numpy_backend.py
```diff
@@ -111,25 +87,22 @@ def unpad(x, i0, i1):
...
-backend.real = real
 backend.unpad = unpad
 backend.pad = pad
-backend.pad_1d = pad_1d
 backend.cdgmm = cdgmm
-backend.fft = FFT(lambda x:scipy.fftpack.fft(x),
-                  lambda x:scipy.fftpack.ifft(x),
-                  lambda x:np.real(scipy.fftpack.ifft(x)),
+backend.fft = FFT(lambda x:scipy.fft.fft(x),
+                  lambda x:scipy.fft.fft(x),
```
janden: So SciPy autodetects the input type and calls the real FFT if necessary?

kymatio/scattering1d/backend/tensorflow_backend.py
```diff
@@ -113,16 +89,15 @@ def unpad(x, i0, i1):
...
+backend.modulus = Modulus()
 backend.subsample_fourier = subsample_fourier
-backend.real = real
 backend.unpad = unpad
 backend.cdgmm = cdgmm
 backend.pad = pad
-backend.pad_1d = pad_1d
-backend.fft = FFT(lambda x: tf.signal.fft(x, name='ifft1d'),
+backend.fft = FFT(lambda x: tf.signal.fft(x, name='fft1d'),
+                  lambda x: tf.signal.fft(tf.cast(x, tf.complex64), name='rfft1d'),
```
janden: So there's no speedup in TF for real FFTs?

kymatio/scattering1d/backend/torch_backend.py
janden: Ok.

janden: Yes. Set to zero as before.

kymatio/scattering2d/tests/test_torch_scattering2d.py
```diff
@@ -236,10 +233,6 @@ def test_cdgmm_exceptions(self, backend):
             backend.cdgmm(torch.empty(3, 4, 5, 2), torch.empty(4, 3, 2))
         assert 'not compatible' in exc.value.args[0]
 
-        with pytest.raises(TypeError) as exc:
-            backend.cdgmm(torch.empty(3, 4, 5, 1), torch.empty(4, 5, 1))
-        assert 'input should be complex' in exc.value.args[0]
-
```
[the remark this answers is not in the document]
janden: We shouldn't remove tests if they correspond to actual code that's executed in the library. The question is whether `cdgmm` should be testing for complex inputs. The implementation may depend on it being complex (hence the `c` in `cdgmm`), so I suggest we keep the test for now and discuss this in an issue somewhere.

kymatio/scattering2d/tests/test_torch_scattering2d.py
```diff
@@ -327,6 +320,15 @@ def test_fft(self, backend):
         assert z.shape == x.shape[:-1]
         assert torch.allclose(y[..., 0], z)
 
+        x = x[..., 0].contiguous()
+        y[..., 1] = 0.0
+
+        z = backend.fft(x, 'R2C')
+        z = z
```
janden: What?

kymatio/scattering2d/tests/test_torch_scattering2d.py
```diff
@@ -327,6 +320,15 @@ def test_fft(self, backend):
         assert z.shape == x.shape[:-1]
         assert torch.allclose(y[..., 0], z)
 
+        x = x[..., 0].contiguous()
```
janden: Maybe call it `x_r`?

kymatio/scattering2d/tests/test_torch_scattering2d.py
```diff
@@ -327,6 +320,15 @@ def test_fft(self, backend):
...
+        y[..., 1] = 0.0
```
janden: Just `0` is fine.

kymatio/scattering2d/tests/test_torch_scattering2d.py
```diff
@@ -327,6 +320,17 @@ def test_fft(self, backend):
...
+
+        z = backend.fft(x, 'R2C')
+        z = z
+
+        assert torch.allclose(y, z)
```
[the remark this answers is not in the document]
janden: Looks fine to me. What's messy about it?

kymatio/scattering2d/tests/test_torch_scattering2d.py
```diff
@@ -336,11 +340,6 @@ def test_fft_exceptions(self, backend_device):
                         inverse=False)
         assert 'done with an inverse' in record.value.args[0]
 
-        x = torch.rand(4, 4, 1)
-        x = x.to(device)
-        with pytest.raises(TypeError) as record:
-            backend.fft(x)
-        assert 'complex' in record.value.args[0]
```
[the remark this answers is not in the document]
janden: Well the test is checking for the `C2C` case, correct? In that case, it should be an error to give it real inputs.

kymatio/scattering3d/backend/torch_backend.py
```diff
@@ -59,10 +27,10 @@ def modulus_rotation(x, module=None):
             which is covariant to 3D translations and rotations.
     """
     if module is None:
-        module = torch.zeros_like(x)
+        module = torch.zeros_like(x)[..., 0]
```
janden: Explicitly set the shape instead of slicing.

kymatio/scattering3d/tests/test_torch_scattering3d.py
```diff
@@ -54,10 +54,20 @@ def test_FFT3d_central_freq_batch(device, backend):
 @pytest.mark.parametrize("device", devices)
 @pytest.mark.parametrize("backend", backends)
 def test_fft3d_error(backend, device):
-    x = torch.zeros(8, 1)
-    with pytest.raises(TypeError) as record:
-        backend.fft(x)
-    assert "should be complex" in record.value.args[0]
```
janden: Again, this needs to be checked in the default (`C2C`) case.

kymatio/scattering3d/tests/test_torch_scattering3d.py
janden: I don't understand the question. Why shouldn't it have been here?

kymatio/backend/torch_backend.py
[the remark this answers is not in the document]
MuawizChaudhary: Good idea.

kymatio/scattering3d/tests/test_torch_scattering3d.py
MuawizChaudhary: Its not testing anything, nor was any of it used in a test.

kymatio/scattering1d/backend/tensorflow_backend.py
MuawizChaudhary: There is a rfft function, however we don't get the correct size from the output of the tf.rfft function.

kymatio/scattering3d/tests/test_torch_scattering3d.py
janden: Looks like it's testing that broadcasting works properly? Just because there's no assert doesn't mean that it's not testing a certain code path (more like a smoke test than a standard unit test).

kymatio/scattering1d/backend/tensorflow_backend.py
janden: What size does it give?

kymatio/backend/base_backend.py
MuawizChaudhary: In one backend (tensorflow), we cast to complex before we send it into the fft. How do we reconcile that?

kymatio/backend/torch_backend.py
MuawizChaudhary: The 3d test has a different type check, and I don't believe it has a check for dtype, just type.

MuawizChaudhary: I suppose I should change the test/code so that we just make sure its the dtype thats checked.

kymatio/scattering1d/backend/numpy_backend.py
MuawizChaudhary: Yes.  https://docs.scipy.org/doc/scipy/reference/generated/scipy.fft.fft.html#scipy.fft.fft

> If the data type of x is real, a “real FFT” algorithm is automatically used, which roughly halves the computation time. To increase efficiency a little further, use rfft, which does the same calculation, but only outputs half of the symmetrical spectrum.

kymatio/scattering1d/backend/tensorflow_backend.py
MuawizChaudhary: Doesn't appear to be, the size output by the rfft in tf is 513.

MuawizChaudhary:
Actually, the skcuda backend needs to be tested. Can someone tell me why the skcuda is not triggering any of the tests?

kymatio/backend/base_backend.py
janden: > How do we reconcile that?
Reconcile what with what? The to this class (`FFT`) specifies that the input should be real, so we should check to make sure that it is. What it then does internally (such as for TF) is unrelated.

kymatio/backend/torch_backend.py
janden: :+1:

kymatio/scattering1d/backend/numpy_backend.py
janden: :ok:

kymatio/scattering1d/backend/tensorflow_backend.py
janden: As opposed to? 1024? So it's a one-sided transform?

kymatio/backend/base_backend.py
MuawizChaudhary: The 'rfft' for tensorflow isn't the `tf.signal.rfft` but its `tf.sginal.fft`, and we cast the input to complex.

janden: Ok. So what's the problem?

MuawizChaudhary: I was thinking that when it checks if the input tf tensor is real, that its gonna error out, b/c we cast to complex, but I'm guessing we could just make the tf not check for real?

kymatio/scattering1d/backend/tensorflow_backend.py
MuawizChaudhary: Yeah. is it possible to attach/repeat something in order to get the full size?

kymatio/backend/base_backend.py
janden: I still don't understand. In our wrapper, we check that the input is indeed real if `R2C` is specified, correct? Then, we may cast it to complex after that and pass it to `tf.signal.fft` and return the output. At what point are we providing a complex array where a real array is expected?

MuawizChaudhary: Ok you are correct, I wasn't understanding the design, I get it now. 

Yeah, lets add the check for real.

janden: :+1:

PR #552, opened 2020-01-28
API Attempt at new meta

janden:
This is a proposal for dealing with metadata in the 2D code. The idea is to not have a separate `meta` function (or `filters`), but to output the metatdata *along* with the scattering coefficients, if desired. Here, this is achieved using the `out_type` attribute. By default, it is set to `'array'`, which returns the array/tensor we normally get. If it's set to `list`, we get a list of dictionaries. The indexing of the list corresponds to that of the array output (the second dimension if the batch index is one-dimensional). Each element has a key `'coef'`, which contains the actual scattering coefficients, but also `'j'` and `'theta'`. (Probably makes sense to have an `'order'` key, among others, but this is just a proof of concept.)

As an example, we have (in the NP frontend)
```
x = np.random.randn(8, 32, 32)
S = Scattering2D(J=3, shape=(32, 32))

# Get out normal array output. S.out_type is 'array' by default.
Sx_array = S.scattering(x)

# Get our list of dictionaries output
S.out_type = 'list'
Sx_list = S.scattering(x)

# Coefficients are in the 'coef' element.
Sx_array[:, 123, :, :] == Sx_list[123]['coef']

# Get metadata on that scattering coefficient.
j1, j2 = Sx_list[123]['j']
theta1, theta2 = Sx_list[123]['theta']
```
One of the advantages of this approach is that the indexing remains the same, so it's easy to correlate the two outputs (and we can effectively get a `meta` structure for an `'array'` output by setting `out_type = 'list'` and calling `scattering` with an empty input). Another is the flexibility we will have in adding new elements to the dictionary (containing, for example, pointers to the actual filters used). It should also extend well to other modalities (harmonic 3D scattering, joint time–frequency scattering, etc.). Finally, it's not a very *hard* feature, since we can deprecate/disable it in the future if we don't like it by defining new `out_type`s to replace it.

If you think this looks good, I will proceed to make the same additions to the 1D and 3D (deprecating the old `vectorize=False` option in 1D along with the `meta` method).

### DIFF
--- a/kymatio/scattering2d/core/scattering2d.py
+++ b/kymatio/scattering2d/core/scattering2d.py
@@ -8,8 +8,8 @@ def scattering2d(x, pad, unpad, backend, J, L, phi, psi, max_order, vectorize=Tr
     cdgmm = backend.cdgmm
     concatenate = backend.concatenate
 
-    # S is simply a dictionary if we do not perform the averaging...
-    S = {}
+    # Define lists for output.
+    out_S_0, out_S_1, out_S_2 = [], [], []
 
     U_r = pad(x)
 
@@ -22,7 +22,7 @@ def scattering2d(x, pad, unpad, backend, J, L, phi, psi, max_order, vectorize=Tr
     S_0 = fft(U_1_c, 'C2R', inverse=True)
     S_0 = unpad(S_0)
 
-    S[()] = S_0
+    out_S_0.append(S_0)
 
     for n1 in range(len(psi)):
         j1 = psi[n1]['j']
@@ -40,8 +40,7 @@ def scattering2d(x, pad, unpad, backend, J, L, phi, psi, max_order, vectorize=Tr
         S_1_r = fft(S_1_c, 'C2R', inverse=True)
         S_1_r = unpad(S_1_r)
 
-
-        S[(n1,)] = S_1_r
+        out_S_1.append(S_1_r)
 
         if max_order < 2:
             continue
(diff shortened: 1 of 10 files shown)


lostanlen:
This is great. One important use case of this sort of pattern is to compute an argmax on the scattering coefficients.
"Here's a signal x; At every scale j1, what is the j2 of maximum S2(x)[j1,j2]?"

Doing this currently is super hard. Yet this is at the crux of making the scattering transform interpretable (i.e., more interpretable than a trained CNN). It has also shown a lot of insight in audio processing papers, for analysis spectrotemporal modulations.

I'm really excited about this!

kymatio/scattering2d/core/scattering2d.py
```diff
@@ -22,10 +23,14 @@ def scattering2d(x, pad, unpad, backend, J, L, phi, psi, max_order, vectorize=Tr
     S_0 = fft(U_1_c, 'C2R', inverse=True)
     S_0 = unpad(S_0)
 
-    S[()] = S_0
+    out_S_0.append({'coef': S_0,
+                    'j': (),
```
edouardoyallon: 'j':J?

janden:
> "Here's a signal x; At every time t and scale j1, what is the j2 of maximum S2(x)[t,j1,j2]?"

Not sure how to do this cleanly here, but it should be easier than with a separate meta structure.

janden:
Ok. Merging.

PR #553, opened 2020-01-28
ENH keras integration

epnev:
Added a simple 2D keras based demo.
Was able to run on tensorflow 2.1 (but possibly we can also do 2.0)


kymatio/keras.py
```diff
@@ -0,0 +1,26 @@
...
+    def __init__(self,  J, shape, L=8, max_order=2, pre_pad=False):
+        self.J = J
+        self.shape = shape
+        self.L = L
+        self.max_order = max_order
+        self.pre_pad = pre_pad
+        super(Scattering2D, self).__init__()
+
+    def call(self, x):
+        S = scattering2D(J=self.J, shape=self.shape, L=self.L,
```
eickenberg: if I read this correctly, it will create the filters at every call. maybe put that into a build function?

epnev: Good point - will test to see if it makes a difference

janden:
So I was able to get the reshaping to work again. The trick was to use `tf.shape`, which returns a Tensor (containing the shape at execution time).  These then have to be concatenated using `tf.concat` (not `+`, since that would add their values) when reshaping.

epnev:
With  https://github.com/kymatio/kymatio/pull/553/commits/a5195c0cd6d97211e34f742565ccf643568e5347 you don't need to specify a shape when constructing the scattering layers in keras as it reads them directly from the input. However, the input does need to have a specified size. I believe this happens because the code asserts that the length satisfies certain conditions prior to constructing the object and requires an actual number to do so. This requirement can probably be lifted, but that would introduce many changes to the code.

janden:
@edouardoyallon @eickenberg Do we want to include the Keras “frontend” in the multiplexer/entry point (i.e., `Scattering2D(J=J, frontend='keras')`? I don't see why not.

edouardoyallon:
> @edouardoyallon @eickenberg Do we want to include the Keras “frontend” in the multiplexer/entry point (i.e., `Scattering2D(J=J, frontend='keras')`? I don't see why not.

Well, sure.

janden:
> Well, sure.

Ok. Messing around with the entry code in another PR right now so will do that later.

Merging this now.

PR #559, opened 2020-02-07
Fix doc 1 d

### LINKED ISSUE #411: Inconsistent documentation for Scattering1D input and output shapes
Hi,

The shape of the input and output tensors of Scattering1D is inconsistent/unclear in the documentation.

For example in the 1D section of the [user guide](https://www.kymat.io/userguide.html):
> If the input x is a Tensor of size (B, 1, T), the output of the 1D scattering transform is of size (B, P, T/2^J)  

But the output shape is actually (B, 1, P, T/2^J) with such an input tensor.

Similarly, in the [code reference for Scattering1d](https://www.kymat.io/codereference.html) the example provided shows an input tensor of shape (1, 1, T), which I assume corresponds to (batch size, channels, time samples):

```# Set the parameters of the scattering transform.
J = 6
T = 2**13
Q = 8

# Generate a sample signal.
x = torch.randn(1, 1, T)

# Define a Scattering1D object.
S = Scattering1D(J, T, Q)

# Calculate the scattering transform.
Sx = S.forward(x)
```
Which corresponds to an input shape of (1, 1, 8192) and an output size of (1, 1, 126, 128)

Even though the documentation for the forward function a few paragraph down says:

> Given an input Tensor of size (B, T0), where B is the batch size and T0 is the length of the individual signals, this function computes its scattering transform. If the vectorize flag is set to True, the output is in the form of a Tensor or size (B, C, T1), where T1 is the signal length after subsampling to the scale 2**J (with the appropriate oversampling factor to reduce aliasing), and C is the number of scattering coefficients.

Thank you.

edouardoyallon:
Closes #411 

This fix an issue concerning the documentaiton of intput/output shape for 1D which was reported in #411

### DIFF
--- a/doc/source/userguide.rst
+++ b/doc/source/userguide.rst
@@ -92,7 +92,7 @@ Output size
 1-D
 ---
 
-If the input :math:`x` is a Tensor of size :math:`(B, 1, T)`, the output of the
+If the input :math:`x` is a Tensor of size :math:`(B, T)`, the output of the
 1D scattering transform is of size :math:`(B, P, T/2^J)`, where :math:`P` is
 the number of scattering coefficients and :math:`2^J` is the maximum scale of the
 transform. The value of :math:`P` depends on the maximum order of the scattering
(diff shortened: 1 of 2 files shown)


edouardoyallon:
TODO in a different.

PR #564, opened 2020-02-18
TST Move tests into separate subdirectory of root

janden:
Instead of having tests nestled in with the actual code, move them out
to the root.


janden:
Unfortunately, this will cause our coverage to drop since the tests themselves will no longer be included.

Issue #542, opened 2020-01-27
Error in regularized inverse scattering mnist torch example


MuawizChaudhary:
In the planned update, one of the examples, `regularized_inverse_scattering_MNIST_torch.py`, errors out. The error I get is 

> RuntimeError: output with shape [1, 28, 28] doesn't match the broadcast shape [3, 28, 28]

This happens on line 94.

eugenium:
Change line https://github.com/kymatio/kymatio/blob/0.2.X/examples/2d/regularized_inverse_scattering_MNIST_torch.py#L86 to  transforms.Normalize((0.5,), (0.5,))

edouardoyallon:
this needs to be check before the release!

edouardoyallon:
Still not fixed

PR #576, opened 2020-02-19
BUG in regularized inverse 2D

edouardoyallon:
Addresses #542 and correct a bug to use this on a laptop.

### DIFF
--- a/examples/2d/regularized_inverse_scattering_MNIST_torch.py
+++ b/examples/2d/regularized_inverse_scattering_MNIST_torch.py
@@ -33,6 +33,8 @@ from kymatio.caching import get_cache_dir
 from kymatio.datasets import get_dataset_dir
 
 
+device = "cuda" if torch.cuda.is_available() else "cpu"
+
 class Generator(nn.Module):
     def __init__(self, num_input_channels, num_hidden_channels, num_output_channels=1, filter_size=3):
         super(Generator, self).__init__()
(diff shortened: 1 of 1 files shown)


examples/2d/regularized_inverse_scattering_MNIST_torch.py
```diff
@@ -84,7 +86,7 @@ def forward(self, input_tensor):
 
     transforms_to_apply = transforms.Compose([
         transforms.ToTensor(),
-        transforms.Normalize((0.5, 0.5, 0.5), (0.5, 0.5, 0.5))  # Pixel values should be in [-1,1]
+        transforms.Normalize((0.5,), (0.5,))  # Pixel values should be in [-1,1]
```
janden: Don't see how this comment makes any sense. It normalizes the input to have mean 0.5 and standard deviation 0.5, no?

janden:
I should also say that this runs fine on the cluster here using the CPU.

examples/2d/regularized_inverse_scattering_MNIST_torch.py
edouardoyallon: Is it better?

edouardoyallon: (I've simply removed this comment and explained it was simply a standard practice and for compatbility issues)

Issue #565, opened 2020-02-18
Document new out_type argument for 1D and 2D


janden:
This was added, but has not yet been documented.

PR #580, opened 2020-02-22
ENH odcumentation of out_type

edouardoyallon:
Solves #565


janden:
This needs to be part of `param_vectorize` since it's not available in frontends that don't allow for `list` output (like sklearn and Keras). Will fix.

PR #601, opened 2020-03-10
TST Install TF in Circle CI config

### LINKED ISSUE #600: Circle CI doesn't have TF
As a result, the docs for TF and Keras are not generated (see discussion in #595). Issue seems to be that TF is not installed. The Circle CI log has this
```
WARNING: autodoc: failed to import module 'tensorflow' from module 'kymatio'; the following exception was raised:
No module named 'tensorflow'
WARNING: autodoc: failed to import module 'keras' from module 'kymatio'; the following exception was raised:
No module named 'tensorflow'
```
Can probably be solved by adding `pip install tensorflow` after https://github.com/kymatio/kymatio/blob/9df4b37e49e78e174cdacf436ee8e78315fcbd19/.circleci/config.yml#L49

janden:
Hopefully fixes #600. Would probably have to merge into dev to see if it works.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -47,6 +47,7 @@ jobs:
             pip install -r requirements_optional.txt
             pip install torch
             pip install torchvision
+            pip install tensorflow
 
       - save_cache:
           paths:


edouardoyallon:
@eickenberg should we try it? I remember u removed it because then  circl ci was crashing.

eickenberg:
When I was trying to get CircleCI to build the docs, there was one point (after managing it so that it doesn't get killed due to timeout), when there were mysterious segfaults during the tensorflow examples. They persisted when I stopped execution of the tensorflow examples and only went away when I stopped importing tensorflow.

Let's try again and see what happens

PR #603, opened 2020-03-11
test circle ci

edouardoyallon:
Is it because there is no gpu?


edouardoyallon:
why isn't circle ci running?

PR #613, opened 2020-06-04
FIX replace incorrect algorithm image

MuawizChaudhary:
The current image on the [Kymatio User Guide](https://www.kymat.io/userguide.html#practical-implementation) is wrong. The multiplication with first order wavelets occurs first, followed by multiplication with second order wavelets in the scattering transform. 

The image provided in the PR is the fixed image that was submitted and uploaded to the website a few months ago. I do not know how the correct image was overwritten. I believe it must have happened during the Sprint.


janden:
I'm not sure how this happened. Looking at the history it seems that this has been the image since November 2018. Maybe the website got manually overwritten at some point and then reverted with the auto-update?

eickenberg:
Good question on how this might have happened. I am convinced we had this fixed.

Maybe one of those big rebases went wrong? I don't see how, but I can't think of any other reason right now

MuawizChaudhary:
> Can you make sure the text for the labels (“Lowpass”, “1st wavelet”, and so on) are left-aligned? Right now the ones for the wavelets are offset to the left. Also would be good to fix the vertical spacing so it's uniform.

I don't have access to this original diagram. I could modify it with paint tho.

janden:
> I don't have access to this original diagram.

Oh. Didn't you modify the existing one?

> I could modify it with paint tho.

Ok. Let's do that.

Issue #445, opened 2019-10-24
ENH [low priority] option to drop low-pass filters


eickenberg:
I am finding myself in a situation where I need access to the non-lowpass-filtered wavelet moduli.

I realize there is a potential space issue, but it might be not only me wanting access to these.

Going to just convolve the filters myself for the moment, but I might make a PR in the future

lostanlen:
This will cause a huge increase in memory footprint, and might break compatibility if whatever downstream application assumes that all coefficients have the same resolution. Unless you want to oversample?

eickenberg:
Oversampling might be OK. Though I'd probably also want to support the
appropriate subsampling.

I am going to write an example that does this and show that I have to
basically copy out the whole loop that does the scattering and do a minor
modification. Given that example, if more people want this as a primary
functionality, we can then consider implementing it later.

Since this codebase is primarily useful for research purposes, in general I
think it would be nice to be able to tinker with the inner states of the
system easily. I get the memory footprint issue, but I think that only
becomes a big deal at layer 2

lostanlen:
> Since this codebase is primarily useful for research purposes, in general I
think it would be nice to be able to tinker with the inner states of the
system easily. 

Yes, this is what prompted me to make `scattering.m` after `scatnet` v0.2. It is easier to inspect the inner state of the transform in `scattering.m`. But the memory footprint is terrible. It's a tradeoff.

janden:
In terms of sampling, it could simply be a matter of disallowing `vectorize` when we don't have averaging. In fact, I think this is already done in 1D.

MuawizChaudhary:
Does this dictionary stuff we discussed allow this?

lostanlen:
@MuawizChaudhary yes, this is the long-term goal of the dictionary output

PR #618, opened 2020-06-19
API Conv prim

MuawizChaudhary:
This PR adds the convolution primitive, which generalizes the scattering convolution operation. This PR  will eventually resolve issue #305. Perhaps we can resolve issue #445 ?


kymatio/scattering1d/conv_primitive.py
```diff
@@ -0,0 +1,27 @@
+import torch.nn as nn
+import torch.nn.functional as F
+
+class Conv1dFFTPrimative(nn.Module):
```
eickenberg, line 4: primitive

kymatio/scattering1d/conv_primitive.py
```diff
@@ -0,0 +1,27 @@
...
+    def __init__(self, backend):
+        super(Conv1dFFTPrimative, self).__init__()
+        self.name = 'torch'
+        self.backend = backend
+
+    def preprocess_signal(self, x, pad_left=None, pad_right=None):
```
eickenberg, line 10: I would go with `padding` and then a tuple of paddings or a list of tuples of paddings.

The reason I say this is that this fourier primitive can then eventually become N-D and can be used for all dimensionalities. Or we can inherit the different dimensionalities from it and have the inherited class fix the dimension

kymatio/scattering1d/conv_primitive.py
```diff
@@ -0,0 +1,27 @@
...
+        if pad_left is not None:
```
eickenberg, line 11: I could fuzz attack this by putting `pad_left=5, pad_right=None` and probably cause an error.
And with `pad_left=None, pad_right=5` an inaction.

kymatio/scattering1d/conv_primitive.py
```diff
@@ -0,0 +1,27 @@
+import torch.nn as nn
```
eickenberg, line 1: Maybe call it `convolution.py` for now or `convolution_1d.py` maybe. Do we duplicate filenames already? I think so. In that case use the first.

kymatio/scattering1d/core/scattering1d.py
```diff
@@ -1,11 +1,11 @@
 # Authors: Mathieu Andreux, Joakim Anden, Edouard Oyallon
 # Scientific Ancestry: Joakim Anden, Mathieu Andreux, Vincent Lostanlen
-
+from ..conv_primitive import Conv1dFFTPrimative, Conv1dSpatialPrimative
 
 def scattering1d(x, pad, unpad, backend, J, psi1, psi2, phi, pad_left=0,
         pad_right=0, ind_start=None, ind_end=None, oversampling=0,
         max_order=2, average=True, size_scattering=(0, 0, 0),
-        vectorize=False, out_type='array'):
+        vectorize=False, out_type='array', conv_prim='fft'):
```
eickenberg, line 8: `convolution_type='fft'`

kymatio/scattering1d/conv_primitive.py
eickenberg: Actually I would just call it `Conv1dFFT`

kymatio/scattering1d/frontend/torch_frontend.py
```diff
@@ -19,6 +19,7 @@ def __init__(self, J, shape, Q=1, max_order=2, average=True,
         ScatteringBase1D._instantiate_backend(self, 'kymatio.scattering1d.backend.')
         ScatteringBase1D.build(self)
         ScatteringBase1D.create_filters(self)
+        self.conv_prim = 'fft'
```
eickenberg, line 22: `self.convolution` or `self.conv`

kymatio/scattering1d/frontend/torch_frontend.py
```diff
@@ -27,27 +28,41 @@ def register_filters(self):
         saves those arrays as module's buffers."""
         n = 0
         # prepare for pytorch
+        # TODO: have option for different shaped filter depending on
```
eickenberg, line 31: not sure i understand this comment

kymatio/scattering1d/frontend/torch_frontend.py
```diff
@@ -27,27 +28,41 @@ def register_filters(self):
...
+        # convolution primitive?
         for k in self.phi_f.keys():
             if type(k) != str:
                 # view(-1, 1).repeat(1, 2) because real numbers!
-                self.phi_f[k] = torch.from_numpy(
-                    self.phi_f[k]).float().view(-1, 1)
+                if self.conv_prim == 'spatial':
```
eickenberg, line 36: Can you explain this?

eickenberg: what is `spatial` -- i didn't see it elsewhere

eickenberg:
Since this is rebased on another PR, it is a bit hard to review like this. Let's merge the rfft PR first, then rebase this on top. Then things will become clearer

janden:
Same thing here. The history includes too many unrelated commits. Either rebase or I will review when the relevant PRs are merged.

kymatio/scattering1d/conv_primitive.py
MuawizChaudhary: You are smart.

kymatio/scattering1d/frontend/torch_frontend.py
MuawizChaudhary: `spatial` is where we use the torch `Conv1d` module.

MuawizChaudhary: I had it in, but then removed it. I guess I should remove this code then?

eickenberg: OH. I guess we need to find a better name for that

MuawizChaudhary:
What will tests for the convolution primitive look like?

kymatio/scattering1d/convolution.py
```diff
@@ -0,0 +1,27 @@
+import torch.nn as nn
+import torch.nn.functional as F
+
```
janden, line 3: Another blank line.

kymatio/scattering1d/convolution.py
```diff
@@ -0,0 +1,27 @@
...
+        self.name = 'torch'
+        self.backend = backend
+
+    def preprocess_signal(self, x, padding=()):
+        if len(padding) != 0:
+            x = self.backend.pad(x, pad_left=padding[0], pad_right=padding[1])
+        x_hat = self.backend.rfft(x)
+        return x_hat
+
+    def convolution(self, x_hat, conv_filter, sampling_factor, domain='', ind_start=None, ind_end=None):
```
janden, line 16: `sampling` → `subsampling`

kymatio/scattering1d/convolution.py
```diff
@@ -0,0 +1,27 @@
+import torch.nn as nn
+import torch.nn.functional as F
+
+class Conv1dFFT(nn.Module):
+    def __init__(self, backend):
+        super(Conv1dFFT, self).__init__()
+        self.name = 'torch'
+        self.backend = backend
+
+    def preprocess_signal(self, x, padding=()):
```
janden, line 10: Why not have `padding=None`?

kymatio/scattering1d/convolution.py
```diff
@@ -0,0 +1,27 @@
...
+
+    def convolution(self, x_hat, conv_filter, sampling_factor, domain='', ind_start=None, ind_end=None):
+        y_c = self.backend.cdgmm(x_hat, conv_filter)
+        y_hat = self.backend.subsample_fourier(y_c, 2**sampling_factor)
+        if domain == 'real':
+            y_r = self.backend.irfft(y_hat)
+        else:
+            y_r = self.backend.ifft(y_hat)
+
+        if ind_start is not None:
```
janden, line 24: I'm not a huge fan of this. Ideally, we should *always* pad and unpad. It seems weird to only do it in certain circumstances. We should discuss.

kymatio/scattering1d/convolution.py
```diff
@@ -0,0 +1,27 @@
...
+
+    def preprocess_signal(self, x, padding=()):
+        if len(padding) != 0:
+            x = self.backend.pad(x, pad_left=padding[0], pad_right=padding[1])
+        x_hat = self.backend.rfft(x)
+        return x_hat
+
+    def convolution(self, x_hat, conv_filter, sampling_factor, domain='', ind_start=None, ind_end=None):
+        y_c = self.backend.cdgmm(x_hat, conv_filter)
+        y_hat = self.backend.subsample_fourier(y_c, 2**sampling_factor)
```
janden, line 18: Spaces around `**`.

kymatio/scattering1d/convolution.py
```diff
@@ -0,0 +1,27 @@
...
+        if domain == 'real':
```
janden, line 19: Ideally, this should be derived from `x` and `conv_filter`.

kymatio/scattering1d/core/scattering1d.py
```diff
@@ -1,11 +1,11 @@
 # Authors: Mathieu Andreux, Joakim Anden, Edouard Oyallon
 # Scientific Ancestry: Joakim Anden, Mathieu Andreux, Vincent Lostanlen
-
+from ..convolution import Conv1dFFT
 
 def scattering1d(x, pad, unpad, backend, J, psi1, psi2, phi, pad_left=0,
         pad_right=0, ind_start=None, ind_end=None, oversampling=0,
         max_order=2, average=True, size_scattering=(0, 0, 0),
-        vectorize=False, out_type='array'):
+        vectorize=False, out_type='array', convolution_type='fft'):
```
janden, line 8: `conv_type` is shorter and aligns better with `Conv1dFFT`.

kymatio/scattering1d/convolution.py
```diff
@@ -0,0 +1,27 @@
+import torch.nn as nn
+import torch.nn.functional as F
+
+class Conv1dFFT(nn.Module):
```
janden, line 4: We capitalize `D` elsewhere, right? On the other hand, we may write FFT as `Fft` to make it easier to read.

kymatio/scattering1d/core/scattering1d.py
```diff
@@ -64,6 +64,7 @@ def scattering1d(x, pad, unpad, backend, J, psi1, psi2, phi, pad_left=0,
     irfft = backend.irfft
     cdgmm = backend.cdgmm
     concatenate = backend.concatenate
+    conv = Conv1dFFT(backend)
```
janden, line 67: Should check the `conv_type` argument, right? If only to error if it's not `'fft'`.

kymatio/scattering1d/frontend/torch_frontend.py
```diff
@@ -27,27 +28,41 @@ def register_filters(self):
...
         n = 0
         # prepare for pytorch
+        # TODO: have option for different shaped filter depending on
+        # convolution primitive?
         for k in self.phi_f.keys():
             if type(k) != str:
                 # view(-1, 1).repeat(1, 2) because real numbers!
-                self.phi_f[k] = torch.from_numpy(
-                    self.phi_f[k]).float().view(-1, 1)
+                if self.convolution == 'spatial':
```
janden, line 36: What's going on here?

kymatio/scattering1d/convolution.py
MuawizChaudhary: explain?

kymatio/scattering1d/frontend/torch_frontend.py
MuawizChaudhary: This is code I added to make things work for the spatial convolution.  

I should remove it seeing as we are not adding a spatial convolution.

kymatio/scattering1d/convolution.py
janden: If `x` is real and `conv_filter` is Hermitian symmetric (that is, its IFT is real), the result of the convolution is real. Checking `x` is easy (we can do that in the `preprocess_signal` step). Checking the symmetry `conv_filter` is expensive, though.

The way to do it is to preprocess both the signal *and* the filter. The filter preprocessing would take place during the initialization of the scattering object (right after they're calculated). In the case of the FFT convolution, the only “preprocessing” would be to check the symmetry and store that as a flag. For other convolutions (like spatial or overlap–add), we would do other types of processing. The spatial convolution, for example, would compute the IFT and cut the result down so that only the non-zero elements are retained.

Longer-term, once we include padding into the convolution, this filter preprocessing would also include calculating the amount of padding that is needed. Short-term, we should try to include a preprocessing step for the filter as well (if only just to check the symmetry of the filter for now).

kymatio/scattering1d/frontend/torch_frontend.py
janden: Ok. Yeah remove it for now.

kymatio/scattering1d/core/scattering1d.py
MuawizChaudhary: The `Conv1DFFT` should be called at the top of the frontend files, methinks. kinda like how the core is called here.

kymatio/scattering1d/convolution.py
MuawizChaudhary: I believe that `x` should always be real. I could add that, however the domain specified is asking "should we be taking the real part of this signal after computing the ifft." One way we could abstract this out would be having a operation that takes the real part of the ifft, so we call that in core whenever we want to get the real part of the tensor. This would be an operation in the general torch/numpy/tf backends. 

this is also what we were doing before R2C too.

janden: > I believe that x should always be real.

Right now, yes, but it's not necessarily always true. In the joint time–frequency transform, for example, you have to convolve complex signals sometimes.

> One way we could abstract this out would be having a operation that takes the real part of the ifft, so we call that in core whenever we want to get the real part of the tensor.

Right, but as you point out, that would basically undo a lot of the R2C stuff, which I think is useful. It would also prevent us from taking advantage of an efficient `irfft` if one ever materializes.

As I said, what we want to do is have another component that preprocesses the filters according to the convolution type we're doing. For the FFT, this just involves determining if the filter is real and computing the Fourier transform, storing both.

kymatio/scattering1d/core/scattering1d.py
janden: Yes the convolution object needs to be tied to the frontend, not to the core. This way it can be involved in the filter preprocessing as discussed above. It'll then be passed into the core and used there, just like the backend.

kymatio/scattering1d/convolution.py
```diff
@@ -0,0 +1,22 @@
+import torch.nn as nn
+import torch.nn.functional as F
+
+
+class Conv1DFFT(nn.Module):
```
eickenberg, line 5: if this is framework-bound then it should figure in the name

PR #573, opened 2020-02-19
API Refactor code to take advantage of rfft

janden:
This in progress pr refactors the padding and modulus classes to not cast to complex. Instead, the complex casting is taken care of by a framework specific rfft function. This PR also refactors the algorithm to take into account the change in real or complex tensors.  Closes #462.

PR #620, opened 2020-06-27
MAINT Inplace argument removal

MuawizChaudhary:
This PR removes the inplace argument. 

We should squash and merge the R2C pr soon! #573

### DIFF
--- a/kymatio/backend/numpy_backend.py
+++ b/kymatio/backend/numpy_backend.py
@@ -41,7 +41,7 @@ def _is_real(x):
     return (x.dtype == np.float32) or (x.dtype == np.float64)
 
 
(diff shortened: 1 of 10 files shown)


MuawizChaudhary:
Commit 9aa9904 is the most relevant one to review.

eickenberg:
This looks fine to me.

Before we merge, I just want to check with @edouardoyallon : We have not found a single place that uses `inplace=True` so to simplify the code have decided that we can remove that option. Do you have any opinions about this or arguments to keep it that we have overlooked?

MuawizChaudhary:
When I discussed with him he said he was ok with removal.

PR #621, opened 2020-06-27
Jax 2d

MuawizChaudhary:
This PR includes the jax implementation of the 2d scattering transform.

### DIFF
--- a/kymatio/backend/jax_backend.py
+++ b/kymatio/backend/jax_backend.py
@@ -1,0 +1,80 @@
+import jax.numpy as np
+
+def input_checks(x):
+    if x is None:
+        raise TypeError('The input should be not empty.')
+
+
+def modulus(x):
+    """
+        This function implements a modulus transform for complex numbers.
+
+        Usage
+        -----
+        x_mod = modulus(x)
+
+        Parameters
+        ---------
+        x: input complex tensor.
+
+        Returns
+        -------
+        output: a real tensor equal to the modulus of x.
+
+    """
+    return np.abs(x)
+
+
+def complex_check(x):
+    if not _is_complex(x):
+        raise TypeError('The input should be complex.')
+
+
+def real_check(x):
+    if not _is_real(x):
+        raise TypeError('The input should be real.')
+
+
+def _is_complex(x):
+    return (x.dtype == np.complex64) or (x.dtype == np.complex128)
+
+
+def _is_real(x):
+    return (x.dtype == np.float32) or (x.dtype == np.float64)
+
+
+def cdgmm(A, B, inplace=False):
+    """
+        Complex pointwise multiplication between (batched) tensor A and tensor B.
+
+        Parameters
+        ----------
+        A : tensor
+            A is a complex tensor of size (B, C, M, N, 2)
+        B : tensor
+            B is a complex tensor of size (M, N) or real tensor of (M, N)
+        inplace : boolean, optional
+            if set to True, all the operations are performed inplace
+
+        Returns
+        -------
+        C : tensor
+            output tensor of size (B, C, M, N, 2) such that:
+            C[b, c, m, n, :] = A[b, c, m, n, :] * B[m, n, :]
+
+    """
+
+    if not _is_complex(A):
+        raise TypeError('The first input must be complex.')
+
+    if A.shape[-len(B.shape):] != B.shape[:]:
+        raise RuntimeError('The inputs are not compatible for '
+                           'multiplication.')
+
+    if not _is_complex(B) and not _is_real(B):
+        raise TypeError('The second input must be complex or real.')
+
+    if inplace:
+        return np.multiply(A, B, out=A)
+    else:
+        return A * B
--- a/kymatio/backend/jax_backend.py
+++ b/kymatio/backend/jax_backend.py
@@ -43,7 +43,7 @@ def _is_real(x):
     return (x.dtype == np.float32) or (x.dtype == np.float64)
 
 
-def cdgmm(A, B, inplace=False):
+def cdgmm(A, B):
     """
         Complex pointwise multiplication between (batched) tensor A and tensor B.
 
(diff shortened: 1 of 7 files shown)


kymatio/backend/jax_backend.py
```diff
@@ -0,0 +1,80 @@
...
+        raise TypeError('The first input must be complex.')
+
+    if A.shape[-len(B.shape):] != B.shape[:]:
+        raise RuntimeError('The inputs are not compatible for '
+                           'multiplication.')
+
+    if not _is_complex(B) and not _is_real(B):
+        raise TypeError('The second input must be complex or real.')
+
+    if inplace:
```
eickenberg, line 77: maybe don't add inplace back if we are trying to get rid of it in the package

MuawizChaudhary: Yeah, I can remove it now or after the inplace pr is merged.

kymatio/scattering2d/backend/jax_backend.py
```diff
@@ -0,0 +1,119 @@
+# Authors: Edouard Oyallon, Sergey Zagoruyko, Muawiz Chaudhary
```
eickenberg, line 1: this author list is wrong. we should probably remove all those lines from the full repo, because they do not give any information beyond the authors and contributors list we already have. so i'd remove this here and then do a separate PR to remove it everywhere else

tests/scattering2d/test_jax_scattering2d.py
```diff
@@ -0,0 +1,285 @@
...
+backends.append(backend)
+
+
+class TestPad:
+    @pytest.mark.parametrize('backend', backends)
+    def test_Pad(self, backend):
+        pad = backend.Pad((2, 2, 2, 2), (4, 4), pre_pad=False)
+
+        x = np.random.randn(4, 4) + 1J * np.random.randn(4, 4)
+        x = x[np.newaxis, ...]
```
eickenberg, line 22: why not just add the axis above when creating `x`

kymatio/scattering2d/backend/jax_backend.py
```diff
@@ -0,0 +1,119 @@
...
+    def __init__(self, pad_size, input_size, pre_pad=False):
+        """
+            Padding which allows to simultaneously pad in a reflection fashion
+            and map to complex.
+
+            Parameters
+            ----------
+            pad_size : list of 4 integers
+                size of padding to apply.
+            input_size : list of 2 integers
```
eickenberg, line 23: what do we need this arg for? It seems to get completely lost. Do other backends need it?

tests/scattering2d/test_jax_scattering2d.py
```diff
@@ -0,0 +1,285 @@
...
+        assert z[0, 1, 2] == x[0, 1, 0]
+        assert z[0, 1, 3] == x[0, 1, 1]
+
+        pad = backend.Pad((2, 2, 2, 2), (4, 4), pre_pad=True)
+
+        x = np.random.randn(8, 8) + 1J * np.random.randn(8, 8)
+
+        z = pad(x)
+
+        assert np.allclose(x, z)
```
eickenberg, line 39: this doesn't seem like it can pass

tests/scattering2d/test_jax_scattering2d.py
```diff
@@ -0,0 +1,285 @@
...
+
+class TestModulus:
+    @pytest.mark.parametrize('backend', backends)
+    def test_Modulus(self, backend):
+        modulus = backend.modulus
+
+        x = np.random.rand(100, 10, 4) + 1J * np.random.rand(100, 10, 4)
+
+        y = modulus(x)
+        u = np.squeeze(np.sqrt(np.real(x) ** 2 + np.imag(x) ** 2))
```
eickenberg, line 60: what is the squeeze for?

I would opt for `x.real, x.imag`

tests/scattering2d/test_jax_scattering2d.py
```diff
@@ -0,0 +1,285 @@
...
+        v = y
```
eickenberg, line 61: that seems unnecessary

kymatio/scattering2d/backend/jax_backend.py
MuawizChaudhary: agree

kymatio/scattering2d/backend/jax_backend.py
MuawizChaudhary: I believe torch needs it. https://github.com/kymatio/kymatio/blob/dev/kymatio/scattering2d/backend/torch_backend.py#L76

tests/scattering2d/test_jax_scattering2d.py
MuawizChaudhary: Good question.

tests/scattering2d/test_jax_scattering2d.py
MuawizChaudhary: it passes

tests/scattering2d/test_jax_scattering2d.py
MuawizChaudhary: the squeeze is unnecessary

kymatio/scattering2d/backend/jax_backend.py
eickenberg: I don't actually understand why it can't get that out of the batch. it's not like it's precomputing anything

tests/scattering2d/test_jax_scattering2d.py
eickenberg: how? `z=pad(x)` can't even be of the same shape as `x`

eickenberg: ah, it's `pre_pad=False` -- does this mean it's already padded or not? Not, right? So then they should be different shapes. I guess we wanted to get rid of that `pre_pad` thing anyway

PR #625, opened 2020-06-30
API Gabor Scattering

MuawizChaudhary:
This adds the Gabor Scattering to Kymatio. Just for Torch.

### DIFF
--- a/kymatio/scattering3d/backend/torch_backend.py
+++ b/kymatio/scattering3d/backend/torch_backend.py
@@ -1,6 +1,6 @@
 import torch
 import warnings
-
+from torch.nn.functional import pad
 BACKEND_NAME = 'torch'
 from collections import namedtuple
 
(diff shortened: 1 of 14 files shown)


kymatio/scattering3d/backend/torch_backend.py
```diff
@@ -58,11 +58,129 @@ def compute_integrals(input_array, integral_powers):
...
+        and map to complex. 
+
+        Parameters
+        ----------
+        pad_size : list of 4 integers
+            Size of padding to apply [top, bottom, left, right].
+        input_size : list of 2 integers
+            Size of the original signal [height, width].
+        pre_pad : boolean, optional
+            If set to true, then there is no padding one simply adds the
```
eickenberg, line 73: this is not great semantically. do we have this everywhere?

kymatio/scattering3d/backend/torch_backend.py
```diff
@@ -58,11 +58,129 @@ def compute_integrals(input_array, integral_powers):
...
+    def __call__(self, x, k):
+        complex_check(x)
+        contiguous_check(x)
+
+        y = x.view(-1,
+                    k, x.shape[1] // k,
+                    k, x.shape[2] // k,
+                    k, x.shape[3] // k,
+                    2)
+        out = y.mean(5, keepdim=False).mean(3, keepdim=False).mean(1, keepdim=False)
```
eickenberg, line 170: i guess it is still not possible to take the mean over multiple axes yet?

MuawizChaudhary: actually, it appears to be.

kymatio/scattering3d/backend/torch_backend.py
MuawizChaudhary: lets make an issue.

eickenberg: What is going to be the merge order of these PRs? Will this be merged before or after the one removing `pre_pad`

kymatio/scattering3d/backend/torch_backend.py
```diff
@@ -58,11 +58,129 @@ def compute_integrals(input_array, integral_powers):
...
+
+        """
+        if not self.pre_pad:
+            # Pytorch expects its padding as [left, right, top, bottom]
+            pad_size_tmp = self.pad_size_tmp
+            x = pad(x, (pad_size_tmp[2], pad_size_tmp[3], 
+                pad_size_tmp[0], pad_size_tmp[1], pad_size_tmp[4],
+                pad_size_tmp[5]), mode='constant')
+
+            # Note: PyTorch is not effective to pad signals of size N-1 with N
```
eickenberg, line 128: We are padding with zeros for 3D, at least according to `mode=constant`. So this stuff should probably go along with the stuff above

kymatio/scattering3d/filter_bank.py
```diff
@@ -187,3 +188,219 @@ def solid_harmonic_3d(M, N, O, sigma, l, fourier=True):
...
+    start_y = int(N * 2 ** (-res - 1))
+    len_z = int(P * (1 - 2 ** (-res)))
+    start_z = int(P * 2 ** (-res - 1))
+    mask[start_x:start_x + len_x,:, :] = 0
+    mask[:, start_y:start_y + len_y, :] = 0
+    mask[..., start_z:start_z + len_z] = 0
+
+    x = np.multiply(x,mask)
+
+    for k in range(int(M / 2 ** res)):
```
eickenberg, line 265: OK, so this needs fixing. I believe that a simple

```python
def periodize(arr, res):
    ds = 2 ** res
    *b, d, h, w = arr.shape
    output = arr.reshape(b + [ds, d // ds, ds, h // ds, ds, w // ds]).sum(axis=(-6, -4, -2))
    return output
```

should give exactly the same output. (to be checked though)

eickenberg: or is it doing something smart that I am missing?
I do recall that periodize_filter was doing something slightly different from periodize, in that it maybe disregarded higher frequencies. But actually I currently don't recall *why* one would want to do that

kymatio/scattering3d/frontend/base_frontend.py
```diff
@@ -1,12 +1,13 @@
 from ...frontend.base_frontend import ScatteringBase
-from ..filter_bank import solid_harmonic_filter_bank, gaussian_filter_bank
+from ..filter_bank import solid_harmonic_filter_bank, gaussian_filter_bank, filter_bank
+from ..utils import compute_padding
 
 
-class ScatteringBase3D(ScatteringBase):
+class ScatteringHarmonicBase3D(ScatteringBase):
```
eickenberg, line 6: Shouldn't this be named `HarmonicScatteringBase3D`

kymatio/scattering3d/frontend/base_frontend.py
```diff
@@ -108,16 +109,52 @@ def create_filters(self):
...
             sample=cls._doc_sample.format(shape=cls._doc_shape))
 
-        cls.scattering.__doc__ = ScatteringBase3D._doc_scattering.format(
+        cls.scattering.__doc__ = ScatteringHarmonicBase3D._doc_scattering.format(
             array=cls._doc_array,
             n=cls._doc_array_n)
 
 
-__all__ = ['ScatteringBase3D']
+class ScatteringBase3D(ScatteringBase):
```
eickenberg, line 124: Does this need a `_document` class method?

kymatio/scattering3d/filter_bank.py
eickenberg: Maybe @janden or @edouardoyallon knows or remembers something about this

eickenberg:
This is looking pretty good to me so far. It would need some more substantial tests. Currently the only test is a smoke test showing that the code runs, but it doesn't check the output.

Maybe the box is a good example to store. A first results test can also be to use the current test, which uses a Dirac, and show that the output of the first layer is the magnitude of the filters

kymatio/scattering3d/filter_bank.py
eickenberg: @edouardoyallon

kymatio/scattering3d/backend/torch_backend.py
eickenberg: since we have the `prepad` refactor PR, this will need to be removed here, because we are merging this after

kymatio/scattering3d/backend/torch_backend.py
```diff
@@ -58,11 +58,126 @@ def compute_integrals(input_array, integral_powers):
...
+
+        self.build()
+
+    def build(self):
+        """Builds the padding module. 
+            
+            Attributes
+            ----------
+            padding_module : ReflectionPad2d
+                Pads the input tensor using the reflection of the input
```
eickenberg, line 89: Are we reflection padding here? I think we need at least 0 padding as an option, if not as the default

kymatio/scattering3d/backend/torch_backend.py
```diff
@@ -58,11 +58,126 @@ def compute_integrals(input_array, integral_powers):
...
+        pad_size_tmp = list(self.pad_size)
+
+        # This handles the case where the padding is equal to the image size
+        if pad_size_tmp[0] == self.input_size[0]:
+            pad_size_tmp[0] -= 1
+            pad_size_tmp[1] -= 1
+        if pad_size_tmp[2] == self.input_size[1]:
+            pad_size_tmp[2] -= 1
+            pad_size_tmp[3] -= 1
+        if pad_size_tmp[4] == self.input_size[2]:
```
eickenberg, line 102: This stuff is reflection-padding-specific. We might not even need it

kymatio/scattering3d/backend/torch_backend.py
```diff
@@ -58,11 +58,126 @@ def compute_integrals(input_array, integral_powers):
...
+
+        Returns
+        -------
+        in_[..., 1:-1, 1:-1, 1:-1, :] : tensor
+            Output tensor. Unpadded input.
+
+    """
+    return in_[..., 1:-1, 1:-1, 1:-1, 0]
+
+class SubsampleFourier(object):
```
eickenberg, line 160: Why would this be a class? Make it a function

PR #627, opened 2020-07-02
TST Test 2d

MuawizChaudhary:
This adds a few tests for the 2d suite of tests. Also splits 2d tests into scattering and backend tests. 

I will be adding batch agnostic tests soon.


tests/scattering2d/test_numpy_backend_2d.py
```diff
@@ -0,0 +1,184 @@
...
+        y = np.random.randn(3, 6, 6) + 1J * np.random.randn(3, 6, 6)
+        z = np.random.randn(3, 6, 6) + 1J * np.random.randn(3, 6, 6)
+
+        w = backend.concatenate((x, y, z))
+
+        assert w.shape == (x.shape[0],) + (3,) + (x.shape[-2:])
+        assert np.allclose(w[:, 0, ...], x)
+        assert np.allclose(w[:, 1, ...], y)
+        assert np.allclose(w[:, 2, ...], z)
+
```
janden, line 183: Remove trailing blank lines.

tests/scattering2d/test_numpy_backend_2d.py
```diff
@@ -0,0 +1,184 @@
...
+        z = pad(x)
+
+        assert z.shape == (1, 8, 8)
+        assert z[0, 2, 2] == x[0, 0, 0]
+        assert z[0, 1, 0] == x[0, 1, 2]
+        assert z[0, 1, 1] == x[0, 1, 1]
+        assert z[0, 1, 2] == x[0, 1, 0]
+        assert z[0, 1, 3] == x[0, 1, 1]
+
+        pad = backend.Pad((2, 2, 2, 2), (4, 4), pre_pad=True)
```
eickenberg, line 26: What is the intended merge order? I see this here, which might conflict with the PR that removes this arg

MuawizChaudhary: Lets start merging Test1d, Test2d. `pre_pad` we'll worry about after.

janden:
Good to merge.

MuawizChaudhary:
Looks like travis passes! Lets merge!

eickenberg:
anybody understand what's up with jenkins?

eickenberg:
OK, I would like to merge this ASAP.
I have just rebased it.

The only issue we currently have is that merging this will break testing for all commits until the last one, because it is only there that the inplace for numpy cdgmm argument is removed in the test.

The inplace argument is removed in the code in an already commit MAINT commit in the dev branch. Ideally we would get the test modification commit in there, but I don't really want to modify the history.
A slightly less bad solution was to bring the MAINT remove inplace test to the top, but unfortunately that is not super easy because of deleted files.

I am a bit hesitant to squash these commits otherwise, because they are semantically quite distinct.

I would suggest merging this as is into dev, and then later on rearranging the commits on dev to bring the test refactoring closer to where the inplace argument is removed and potentially fuse those commits before release.

eickenberg:
I have run the tests locally several times and at least at current head, they pass.

So in the interest of clearing up the PR tracker a bit before the sprint, I am going to merge this now, and we can, if we deem it necessary, rewrite the dev commits just before cutting the next release. I will make an issue reminding us of this.

Issue #628, opened 2020-07-03
MAINT Remove option to not pad


MuawizChaudhary:
Currently in the 2d backend we have the option of [simply not doing a padding](https://github.com/kymatio/kymatio/blob/dev/kymatio/scattering2d/backend/torch_backend.py#L13).  This changes the semantic meaning of the `Pad` class object. We should remove this option, as we want this function to always pad.

janden:
@edouardoyallon I believe you added this at some point. Do you want to discuss where this comes from and what it's used for?

edouardoyallon:
It was used for imagenet: sometimes you might have a signal which is large enough to avoid additional padding, and you just crop the final result(which leads to no border effect). For instance, on imagenet, the images might be 256^2 and but the architecture has to process(for legacy reasons) 224^2 images... so instead, one processes the 256^2 image and one crops the final results, thus it's not worth to pad/unpad. Maybe it's fast enough now. Not sure.

janden:
Ah ok. So we'll leave the option in the scattering transform but take it out of the `Pad` module.

eickenberg:
This looks like something to be addressed pre-sprint or at the latest at the sprint.

MuawizChaudhary:
It looks like this was addressed during the 2021 sprint.

PR #630, opened 2020-07-03
API remove pre_pad option from Pad

MuawizChaudhary:
Currently we are removing the `pre_pad` option from the `Pad` class. Unfortunately, this has led to a change in the core 2d file, with the addition of the real to complex function. This is required in the case of torch, but not in the case of the other frameworks. How do we want to resolve this?

### DIFF
--- a/kymatio/scattering2d/backend/numpy_backend.py
+++ b/kymatio/scattering2d/backend/numpy_backend.py
@@ -11,7 +11,7 @@ from ...backend.base_backend import FFT
 
 
 class Pad(object):
-    def __init__(self, pad_size, input_size, pre_pad=False):
+    def __init__(self, pad_size, input_size):
         """
             Padding which allows to simultaneously pad in a reflection fashion
             and map to complex.
@@ -22,21 +22,16 @@ class Pad(object):
                 size of padding to apply.
             input_size : list of 2 integers
                 size of the original signal
-            pre_pad : boolean
-                if set to true, then there is no padding, one simply converts
-                from real to complex.
 
         """
-        self.pre_pad = pre_pad
         self.pad_size = pad_size
 
     def __call__(self, x):
-        if self.pre_pad:
-            return x
-        else:
-            np_pad = ((self.pad_size[0], self.pad_size[1]), (self.pad_size[2], self.pad_size[3]))
-            output = np.pad(x, ((0,0), np_pad[0], np_pad[1]), mode='reflect')
-            return output
+        paddings = ((0, 0),) * (x.ndim - 2)
+        paddings += ((self.pad_size[0], self.pad_size[1]), (self.pad_size[2], self.pad_size[3]))
+
+        output = np.pad(x, paddings, mode='reflect')
+        return output
 
 
 def unpad(in_):
--- a/kymatio/scattering2d/backend/tensorflow_backend.py
+++ b/kymatio/scattering2d/backend/tensorflow_backend.py
@@ -10,7 +10,7 @@ from ...backend.tensorflow_backend import Modulus, cdgmm, concatenate
 from ...backend.base_backend import FFT
 
 class Pad(object):
-    def __init__(self, pad_size, input_size, pre_pad=False):
+    def __init__(self, pad_size, input_size):
         """
             Padding which allows to simultaneously pad in a reflection fashion
             and map to complex.
@@ -20,19 +20,13 @@ class Pad(object):
                 size of padding to apply.
             input_size : list of 2 integers
                 size of the original signal
-            pre_pad : boolean
-                if set to true, then there is no padding, one simply adds the imaginarty part.
         """
-        self.pre_pad = pre_pad
         self.pad_size = pad_size
 
     def __call__(self, x):
-        if self.pre_pad:
-            return x
-        else:
-            paddings = [[0, 0]] * len(x.shape[:-2])
-            paddings += [[self.pad_size[0], self.pad_size[1]], [self.pad_size[2], self.pad_size[3]]]
-            return tf.cast(tf.pad(x, paddings, mode="REFLECT"), tf.complex64)
+        paddings = [[0, 0]] * len(x.shape[:-2])
+        paddings += [[self.pad_size[0], self.pad_size[1]], [self.pad_size[2], self.pad_size[3]]]
+        return tf.pad(x, paddings, mode="REFLECT")
 
 def unpad(in_):
     """
@@ -46,6 +40,7 @@ def unpad(in_):
         in_[..., 1:-1, 1:-1]
     """
     return in_[..., 1:-1, 1:-1]
+
 
 class SubsampleFourier(object):
     """ Subsampling of a 2D image performed in the Fourier domain.
--- a/kymatio/scattering2d/backend/tensorflow_backend.py
+++ b/kymatio/scattering2d/backend/tensorflow_backend.py
@@ -26,7 +26,7 @@ class Pad(object):
     def __call__(self, x):
         paddings = [[0, 0]] * len(x.shape[:-2])
         paddings += [[self.pad_size[0], self.pad_size[1]], [self.pad_size[2], self.pad_size[3]]]
-        return tf.pad(x, paddings, mode="REFLECT")
+        return tf.cast(tf.pad(x, paddings, mode="REFLECT"), tf.complex64)
 
 def unpad(in_):
     """
--- a/kymatio/scattering2d/backend/torch_backend.py
+++ b/kymatio/scattering2d/backend/torch_backend.py
@@ -13,7 +13,7 @@ from ...backend.base_backend import FFT
 
 
 class Pad(object):
-    def __init__(self, pad_size, input_size, pre_pad=False):
+    def __init__(self, pad_size, input_size):
         """Padding which allows to simultaneously pad in a reflection fashion
             and map to complex.
 
@@ -23,11 +23,7 @@ class Pad(object):
                 Size of padding to apply [top, bottom, left, right].
             input_size : list of 2 integers
                 size of the original signal [height, width].
-            pre_pad : boolean, optional
-                If set to true, then there is no padding, one simply adds the imaginary part.
-
         """
-        self.pre_pad = pre_pad
         self.pad_size = pad_size
         self.input_size = input_size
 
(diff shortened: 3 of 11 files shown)


janden:
As we discussed in the meeting, the proper thing to do with respect to the “casting to complex” is to deal with this in the Torch frontend code. This way it doesn't have to touch the core code.

A more important question is what to do with the `pre_pad` option to the Scattering object. If this PR is merged, it'll be ignored, which changes the API. I'm not sure why the `pre_pad` option arose (@edouardoyallon can answer this one), but we don't want to break backwards compatibility. Instead of ignoring it, we have two options:

- If `pre_pad == True`, we set `self.pad` and `self.unpad` to `None` and have the core code skip these steps. This gets a little ugly, though, and makes that code hard to follow.
- If `pre_pad == True`, we set `self.pad` and `self.unpad` to 'lambda x: x`, which means they don't do anything. It's not the prettiest, but it'll work and won't break the core code. Alternatively, we can create some sort of `Identity` object, which does essentially the same thing, but won't look as strange.

kymatio/scattering2d/backend/numpy_backend.py
```diff
@@ -22,23 +22,16 @@ def __init__(self, pad_size, input_size, pre_pad=False):
...
-        self.pre_pad = pre_pad
         self.pad_size = pad_size
 
     def __call__(self, x):
-        if self.pre_pad:
-            return x
-        else:
-            paddings = (0, 0),
-            paddings += ((self.pad_size[0], self.pad_size[1]), (self.pad_size[2], self.pad_size[3]))
+        paddings = (0, 0),
```
eickenberg, line 30: Am I missing something? Didn't this need to follow the number of axes before the last two? I.e. `paddings = (0, 0), * (x.ndim - 2)`

eickenberg:
I actually don't see where the frontend calls the `scattering` function with the new `pre_pad` argument. Am I missing something?

Also, while we are at it, could we name this `pre_padded` instead of `pre_pad`? `pre_pad` is ambiguous in meaning

kymatio/scattering2d/backend/numpy_backend.py
MuawizChaudhary: We could do it that way, but we don't require this function to be batch agnostic, do we? 

Do you want it to be batch agnostic?

janden:
So according to @edouardoyallon (see https://github.com/kymatio/kymatio/issues/628#issuecomment-655161241), we need the option to set `pre_padded = True` when calling the scattering transform, so let's keep that behavior.

janden:
> Close?

Why?

kymatio/scattering2d/backend/numpy_backend.py
eickenberg: 1. I somehow thought in 2D we typically work with 4 axes. This here implies we work with 3
2. I don't see a problem with having some flexibility where it doesn't cost much to add and leads to readable code

kymatio/scattering2d/backend/tensorflow_backend.py
```diff
@@ -19,19 +19,13 @@ def __init__(self, pad_size, input_size, pre_pad=False):
...
         self.pad_size = pad_size
 
     def __call__(self, x):
-        if self.pre_pad:
-            return x
-        else:
-            paddings = [[0, 0]] * len(x.shape[:-2])
-            paddings += [[self.pad_size[0], self.pad_size[1]], [self.pad_size[2], self.pad_size[3]]]
-            return tf.pad(x, paddings, mode="REFLECT")
+        paddings = [[0, 0]] * len(x.shape[:-2])
```
eickenberg, line 26: see conversation abt numpy backend padding above. Here the padding is adapted to input shape

kymatio/scattering2d/backend/torch_backend.py
```diff
@@ -70,19 +66,23 @@ def __call__(self, x):
...
-            # Note: PyTorch is not effective to pad signals of size N-1 with N
-            # elements, thus we had to add this fix.
-            if self.pad_size[0] == self.input_size[0]:
-                x = torch.cat([x[:, :, 1, :].unsqueeze(2), x, x[:, :, x.shape[2] - 2, :].unsqueeze(2)], 2)
-            if self.pad_size[2] == self.input_size[1]:
-                x = torch.cat([x[:, :, :, 1].unsqueeze(3), x, x[:, :, :, x.shape[3] - 2].unsqueeze(3)], 3)
+        # Note: PyTorch is not effective to pad signals of size N-1 with N
+        # elements, thus we had to add this fix.
+        if self.pad_size[0] == self.input_size[0]:
+            x = torch.cat([x[:, :, 1, :].unsqueeze(2), x, x[:, :, x.shape[2] - 2, :].unsqueeze(2)], 2)
```
eickenberg, line 74: could remove some unsqueeze operations here

eickenberg:
This looks good to me

kymatio/scattering2d/core/scattering2d.py
```diff
@@ -10,11 +10,16 @@ def scattering2d(x, pad, unpad, backend, J, L, phi, psi, max_order,
...
 
     # Define lists for output.
     out_S_0, out_S_1, out_S_2 = [], [], []
-
-    U_r = pad(x)
+    
+    if not pre_pad:
+        U_r = pad(x)
+    else:
+        U_r = to_complex(x)
```
janden, line 22: Remove these. The only use seems to be in Torch where we need to add a singleton dimension at the end. This can be done in the frontend and have `pad` assume the input has that shape to begin with.

kymatio/scattering2d/backend/torch_backend.py
MuawizChaudhary: what would be the smartest way of doing it? atm im thinking `x[:, :, 1, :].reshape(x.shape[:2] + (1,) + x.shape[3:])`

kymatio/scattering2d/frontend/base_frontend.py
```diff
@@ -23,8 +23,12 @@ def build(self):
             raise RuntimeError('The smallest dimension should be larger than 2^J.')
         self.M_padded, self.N_padded = compute_padding(self.M, self.N, self.J)
         # pads equally on a given side if the amount of padding to add is an even number of pixels, otherwise it adds an extra pixel
-        self.pad = self.backend.Pad([(self.M_padded - self.M) // 2, (self.M_padded - self.M+1) // 2, (self.N_padded - self.N) // 2,
-                                (self.N_padded - self.N + 1) // 2], [self.M, self.N], pre_pad=self.pre_pad)
+        if not self.pre_pad:
+            self.pad = self.backend.Pad([(self.M_padded - self.M) // 2, (self.M_padded - self.M+1) // 2, (self.N_padded - self.N) // 2,
+                                (self.N_padded - self.N + 1) // 2], [self.M, self.N])
+        else:
+            self.pad = lambda x: x
```
eickenberg, line 30: Do we need this? Can't we just branch inside the scattering function and not fool it into thinking it is padding?

kymatio/scattering2d/frontend/torch_frontend.py
```diff
@@ -13,6 +13,8 @@ def __init__(self, J, shape, L=8, max_order=2, pre_pad=False,
         ScatteringBase2D._instantiate_backend(self, 'kymatio.scattering2d.backend.')
         ScatteringBase2D.build(self)
         ScatteringBase2D.create_filters(self)
+        if pre_pad:
+            self.pad = lambda x: x.reshape(x.shape + (1,))
```
eickenberg, line 17: Did we discuss this? This seems awkward since the padding seems to be casting to our real data type, so once again the padding is not doing what it was advertised to do.

kymatio/scattering2d/backend/torch_backend.py
eickenberg: `torch.cat([x[:, :, 1:2], x, x[:, :, -2:-1]])` looks like it should do it. Please check

kymatio/scattering2d/frontend/base_frontend.py
eickenberg: I just re-read and learned this is about backwards compatibility

kymatio/scattering2d/frontend/torch_frontend.py
eickenberg: I guess this is also about backwards compatibility?

eickenberg:
OK so if we are keeping the fugliness of having pad do minor non-padding operations because of compatibility reasons, then I think this is good to go. 

Optionally the unsqueeze could be fixed before merging this in

kymatio/scattering2d/core/scattering2d.py
```diff
@@ -1,7 +1,7 @@
 # Authors: Edouard Oyallon, Muawiz Chaudhary
 # Scientific Ancestry: Edouard Oyallon, Laurent Sifre, Joan Bruna
 
-def scattering2d(x, pad, unpad, backend, J, L, phi, psi, max_order,
+def scattering2d(x, pad, unpad, pre_pad, backend, J, L, phi, psi, max_order,
```
eickenberg, line 4: Why are we adding this argument again? It's not being used in the function.

kymatio/scattering2d/frontend/numpy_frontend.py
```diff
@@ -35,7 +35,7 @@ def scattering(self, input):
 
         input = input.reshape((-1,) + signal_shape)
 
-        S = scattering2d(input, self.pad, self.unpad, self.backend, self.J,
+        S = scattering2d(input, self.pad, self.unpad, self.pre_pad, self.backend, self.J,
```
eickenberg, line 38: The `self.pre_pad` doesn't seem to be used.

edouardoyallon:
I'd like to take care of this PR. What's the state? I thought it was interesting to have the possibility to *not* pad the input signal, assuming that this has been done by the user(or that the signal actually needs no padding because we "crop" the image)
@janden @eickenberg ?

edouardoyallon:
ping @all , I think prepad is cool - let me know how you want the code to be and I do it

janden:
@edouardoyallon To be clear, the issue here is not to remove the `pre_pad` option completely, just from the `Pad` class. It doesn't make sense to have a padding function but tell it to actually do nothing. In that case, we shouldn't call the padding function in the first place. I can take a look at the code and see where we are.

kymatio/scattering2d/backend/torch_backend.py
janden: Not in scope for this PR.

kymatio/scattering2d/frontend/torch_frontend.py
janden: I think this is fine.

kymatio/scattering2d/core/scattering2d.py
janden: Removed.

kymatio/scattering2d/frontend/numpy_frontend.py
janden: Removed.

janden:
Changed base to master since this is an internal change and we might as well get it into 0.2.1.

Everything looks good on my end. @eickenberg can you look through?

janden (approved):

janden:
One concern I have here is that we don't test the `pre_pad = False` case. We should probably add a unit test. Right now, we have no assurance that it is doing the right thing…

janden:
Ping @eickenberg

eickenberg (approved):
Apart from a potential test of the `pre_pad=True` case this LGTM for merge

eickenberg:
In the interest of time it can also be merged as is

janden:
Made an issue. Will merge now.

PR #631, opened 2020-07-16
API Backend refactor

MuawizChaudhary:
This PR is in progress. All the tests currently do not run. 

Refactors the named tuple convention out of the 1d.

### DIFF
--- a/kymatio/backend/numpy_backend.py
+++ b/kymatio/backend/numpy_backend.py
@@ -1,57 +1,65 @@
+import numpy
+import scipy.fftpack
 
-   
+
 class NumpyBackend:
-    def __init__(self, name='numpy'):
-        import numpy
-        import scipy.fftpack
-        self.np = numpy
-        self.fft = scipy.fftpack
-        self.name = name
+    _np = numpy
+    _fft = scipy.fftpack
 
-    def input_checks(self, x):
+    name = 'numpy'
+
+    @staticmethod
+    def input_checks(x):
         if x is None:
             raise TypeError('The input should be not empty.')
-    
-    def complex_check(self, x):
-        if not self._is_complex(x):
-            raise TypeError('The input should be complex.')
-    
-    def real_check(self, x):
-        if not self._is_real(x):
-            raise TypeError('The input should be real.')
-    
-    def _is_complex(self, x):
-        return (x.dtype == self.np.complex64) or (x.dtype == self.np.complex128)
-    
-    def _is_real(self, x):
-        return (x.dtype == self.np.float32) or (x.dtype == self.np.float64)
- 
-    def concatenate(self, arrays):
-        return self.np.stack(arrays, axis=1)
 
-    def modulus(self, x):
+    @classmethod
+    def complex_check(cls, x):
+        if not cls._is_complex(x):
+            raise TypeError('The input should be complex.')
+
+    @classmethod
+    def real_check(cls, x):
+        if not cls._is_real(x):
+            raise TypeError('The input should be real.')
+
+    @classmethod
+    def _is_complex(cls, x):
+        return (x.dtype == cls._np.complex64) or (x.dtype == cls._np.complex128)
+
+    @classmethod
+    def _is_real(cls, x):
+        return (x.dtype == cls._np.float32) or (x.dtype == cls._np.float64)
+
+    @classmethod
+    def concatenate(cls, arrays):
+        return cls._np.stack(arrays, axis=1)
+
+    @classmethod
+    def modulus(cls, x):
         """
             This function implements a modulus transform for complex numbers.
-    
+
             Usage
             -----
             x_mod = modulus(x)
-    
+
             Parameters
             ---------
             x: input complex tensor.
-    
+
             Returns
             -------
             output: a real tensor equal to the modulus of x.
-    
+
         """
-        return self.np.abs(x)
-   
-    def cdgmm(self, A, B):
+        return cls._np.abs(x)
+
+    @classmethod
+    def cdgmm(cls, A, B):
         """
             Complex pointwise multiplication between (batched) tensor A and tensor B.
-    
+
             Parameters
             ----------
             A : tensor
@@ -60,22 +68,22 @@ class NumpyBackend:
                 B is a complex tensor of size (M, N) or real tensor of (M, N)
             inplace : boolean, optional
                 if set to True, all the operations are performed inplace
-    
+
             Returns
             -------
             C : tensor
                 output tensor of size (B, C, M, N, 2) such that:
                 C[b, c, m, n, :] = A[b, c, m, n, :] * B[m, n, :]
-    
+
         """
-        if not self._is_complex(A):
+        if not cls._is_complex(A):
             raise TypeError('The first input must be complex.')
-    
+
         if A.shape[-len(B.shape):] != B.shape[:]:
             raise RuntimeError('The inputs are not compatible for '
                                'multiplication.')
-    
-        if not self._is_complex(B) and not self._is_real(B):
+
+        if not cls._is_complex(B) and not cls._is_real(B):
             raise TypeError('The second input must be complex or real.')
-    
+
         return A * B
--- a/kymatio/backend/numpy_backend.py
+++ b/kymatio/backend/numpy_backend.py
@@ -1,73 +1,81 @@
-import numpy as np
 
-def input_checks(x):
-    if x is None:
-        raise TypeError('The input should be not empty.')
+   
+class NumpyBackend:
+    def __init__(self, name='numpy'):
+        import numpy
+        import scipy.fftpack
+        self.np = numpy
+        self.fft = scipy.fftpack
+        self.name = name
 
-def complex_check(x):
-    if not _is_complex(x):
-        raise TypeError('The input should be complex.')
+    def input_checks(self, x):
+        if x is None:
+            raise TypeError('The input should be not empty.')
+    
+    def complex_check(self, x):
+        if not self._is_complex(x):
+            raise TypeError('The input should be complex.')
+    
+    def real_check(self, x):
+        if not self._is_real(x):
+            raise TypeError('The input should be real.')
+    
+    def _is_complex(self, x):
+        return (x.dtype == self.np.complex64) or (x.dtype == self.np.complex128)
+    
+    def _is_real(self, x):
+        return (x.dtype == self.np.float32) or (x.dtype == self.np.float64)
+ 
+    def concatenate(self, arrays):
+        return self.np.stack(arrays, axis=1)
 
-def real_check(x):
-    if not _is_real(x):
-        raise TypeError('The input should be real.')
-
-
-def modulus(x):
-    """
-        This function implements a modulus transform for complex numbers.
-
-        Usage
-        -----
-        x_mod = modulus(x)
-
-        Parameters
-        ---------
-        x: input complex tensor.
-
-        Returns
-        -------
-        output: a real tensor equal to the modulus of x.
-
-    """
-    return np.abs(x)
-
-
-def _is_complex(x):
-    return (x.dtype == np.complex64) or (x.dtype == np.complex128)
-
-
-def _is_real(x):
-    return (x.dtype == np.float32) or (x.dtype == np.float64)
-
-
-def cdgmm(A, B):
-    """
-        Complex pointwise multiplication between (batched) tensor A and tensor B.
-
-        Parameters
-        ----------
-        A : tensor
-            A is a complex tensor of size (B, C, M, N, 2)
-        B : tensor
-            B is a complex tensor of size (M, N) or real tensor of (M, N)
-
-        Returns
-        -------
-        C : tensor
-            output tensor of size (B, C, M, N, 2) such that:
-            C[b, c, m, n, :] = A[b, c, m, n, :] * B[m, n, :]
-
-    """
-
-    if not _is_complex(A):
-        raise TypeError('The first input must be complex.')
-
-    if A.shape[-len(B.shape):] != B.shape[:]:
-        raise RuntimeError('The inputs are not compatible for '
-                           'multiplication.')
-
-    if not _is_complex(B) and not _is_real(B):
-        raise TypeError('The second input must be complex or real.')
-
-    return A * B
+    def modulus(self, x):
+        """
+            This function implements a modulus transform for complex numbers.
+    
+            Usage
+            -----
+            x_mod = modulus(x)
+    
+            Parameters
+            ---------
+            x: input complex tensor.
+    
+            Returns
+            -------
+            output: a real tensor equal to the modulus of x.
+    
+        """
+        return self.np.abs(x)
+   
+    def cdgmm(self, A, B):
+        """
+            Complex pointwise multiplication between (batched) tensor A and tensor B.
+    
+            Parameters
+            ----------
+            A : tensor
+                A is a complex tensor of size (B, C, M, N, 2)
+            B : tensor
+                B is a complex tensor of size (M, N) or real tensor of (M, N)
+            inplace : boolean, optional
+                if set to True, all the operations are performed inplace
+    
+            Returns
+            -------
+            C : tensor
+                output tensor of size (B, C, M, N, 2) such that:
+                C[b, c, m, n, :] = A[b, c, m, n, :] * B[m, n, :]
+    
+        """
+        if not self._is_complex(A):
+            raise TypeError('The first input must be complex.')
+    
+        if A.shape[-len(B.shape):] != B.shape[:]:
+            raise RuntimeError('The inputs are not compatible for '
+                               'multiplication.')
+    
+        if not self._is_complex(B) and not self._is_real(B):
+            raise TypeError('The second input must be complex or real.')
+    
+        return A * B
--- a/kymatio/backend/tensorflow_backend.py
+++ b/kymatio/backend/tensorflow_backend.py
@@ -3,41 +3,15 @@ import tensorflow as tf
 
 
 class TensorFlowBackend(NumpyBackend):
-    def __init__(self):
-        super(TensorFlowBackend, self).__init__(name='tensorflow')
+    name = 'tensorflow'
 
-    def concatenate(self, arrays):
+    @staticmethod
+    def concatenate(arrays):
         return tf.stack(arrays, axis=1)
 
-    def modulus(self, x):
+    @staticmethod
+    def modulus(x):
         norm = tf.abs(x)
-        
+
         return norm
-    
-    def cdgmm(self, A, B):
-        """
-            Complex pointwise multiplication between (batched) tensor A and tensor B.
-            Parameters
-            ----------
-            A : tensor
-                A is a complex tensor of size (B, C, M, N, 2)
-            B : tensor
-                B is a complex tensor of size (M, N) or real tensor of (M, N)
-            inplace : boolean, optional
-                if set to True, all the operations are performed inplace
-            Returns
-            -------
-            C : tensor
-                output tensor of size (B, C, M, N, 2) such that:
-                C[b, c, m, n, :] = A[b, c, m, n, :] * B[m, n, :]
-        """
-        if not self._is_complex(A):
-            raise TypeError('The first input must be complex.')
-    
-        if A.shape[-len(B.shape):] != B.shape[:]:
-            raise RuntimeError('The inputs are not compatible for multiplication.')
-    
-        if not self._is_complex(B) and not self._is_real(B):
-            raise TypeError('The second input must be complex or real.')
-    
-        return A * B
+
--- a/kymatio/backend/tensorflow_backend.py
+++ b/kymatio/backend/tensorflow_backend.py
@@ -1,72 +1,43 @@
+from .numpy_backend import NumpyBackend
 import tensorflow as tf
-import numpy as np
 
 
-def complex_check(x):
-    if not _is_complex(x):
-        raise TypeError('The input should be complex.')
+class TensorFlowBackend(NumpyBackend):
+    def __init__(self):
+        super(TensorFlowBackend, self).__init__(name='tensorflow')
 
-def real_check(x):
-    if not _is_real(x):
-        raise TypeError('The input should be real.')
+    def concatenate(self, arrays):
+        return tf.stack(arrays, axis=1)
 
-def _is_complex(x):
-    return (x.dtype == np.complex64) or (x.dtype == np.complex128)
-
-
-def _is_real(x):
-    return (x.dtype == np.float32) or (x.dtype == np.float64)
-
-
-class Modulus():
-    """This class implements a modulus transform for complex numbers.
-
-    Parameters
-    ----------
-    x: input complex tensor.
-
-    Returns
-    ----------
-    output: a complex tensor equal to the modulus of x.
-
-    Usage
-    ----------
-    modulus = Modulus()
-    x_mod = modulus(x)
-    """
-    def __call__(self, x):
+    def modulus(self, x):
         norm = tf.abs(x)
+        
         return norm
-
-
-def concatenate(arrays, dim):
-    return tf.stack(arrays, axis=dim)
-
-
-def cdgmm(A, B):
-    """
-        Complex pointwise multiplication between (batched) tensor A and tensor B.
-        Parameters
-        ----------
-        A : tensor
-            A is a complex tensor of size (B, C, M, N, 2)
-        B : tensor
-            B is a complex tensor of size (M, N) or real tensor of (M, N)
-        Returns
-        -------
-        C : tensor
-            output tensor of size (B, C, M, N, 2) such that:
-            C[b, c, m, n, :] = A[b, c, m, n, :] * B[m, n, :]
-    """
-
-    if not _is_complex(A):
-        raise TypeError('The first input must be complex.')
-
-    if A.shape[-len(B.shape):] != B.shape[:]:
-        raise RuntimeError('The inputs are not compatible for multiplication.')
-
-    if not _is_complex(B) and not _is_real(B):
-        raise TypeError('The second input must be complex or real.')
-
-    return A * B
-
+    
+    def cdgmm(self, A, B):
+        """
+            Complex pointwise multiplication between (batched) tensor A and tensor B.
+            Parameters
+            ----------
+            A : tensor
+                A is a complex tensor of size (B, C, M, N, 2)
+            B : tensor
+                B is a complex tensor of size (M, N) or real tensor of (M, N)
+            inplace : boolean, optional
+                if set to True, all the operations are performed inplace
+            Returns
+            -------
+            C : tensor
+                output tensor of size (B, C, M, N, 2) such that:
+                C[b, c, m, n, :] = A[b, c, m, n, :] * B[m, n, :]
+        """
+        if not self._is_complex(A):
+            raise TypeError('The first input must be complex.')
+    
+        if A.shape[-len(B.shape):] != B.shape[:]:
+            raise RuntimeError('The inputs are not compatible for multiplication.')
+    
+        if not self._is_complex(B) and not self._is_real(B):
+            raise TypeError('The second input must be complex or real.')
+    
+        return A * B
--- a/kymatio/backend/torch_backend.py
+++ b/kymatio/backend/torch_backend.py
@@ -1,28 +1,7 @@
-from torch.autograd import Function
 import torch
+from torch.autograd import Function
 
-BACKEND_NAME = 'torch'
-
-def input_checks(x):
-    if x is None:
-        raise TypeError('The input should be not empty.')
-
-    contiguous_check(x)
-
-def complex_check(x):
-    if not _is_complex(x):
-        raise TypeError('The input should be complex (i.e. last dimension is 2).')
-
-def real_check(x):
-    if not _is_real(x):
-        raise TypeError('The input should be real.')
-
-def _is_complex(x):
-    return x.shape[-1] == 2
-
-def _is_real(x):
-    return x.shape[-1] == 1
-
+ 
 class ModulusStable(Function):
     """Stable complex modulus
 
@@ -104,6 +83,7 @@ class ModulusStable(Function):
             The gradient with respect to the input.
         """
         x, output = ctx.saved_tensors
+
         if ctx.dim is not None and ctx.keepdim is False and x.dim() != 1:
             grad_output = grad_output.unsqueeze(ctx.dim)
             output = output.unsqueeze(ctx.dim)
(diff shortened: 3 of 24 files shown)


tests/scattering1d/test_tensorflow_backend_1d.py
```diff
@@ -1,7 +1,7 @@
 import pytest
 import numpy as np
-
-from kymatio.scattering1d.backend.tensorflow_backend import backend
```
janden: No changes here.

tests/scattering1d/test_torch_backend_1d.py
```diff
@@ -12,13 +12,14 @@
         skcuda_available = True
 except:
     Warning('torch_skcuda backend not available.')
+skcuda_available = False
```
janden, line 15: Why? Where's the `torch_skcuda` backend anyhow?

tests/scattering1d/test_torch_backend_1d.py
```diff
@@ -176,6 +177,8 @@ def test_unpad():
     # test unpading of a random tensor
     x = torch.randn(8, 4, 1)
 
+    Torch1DBackend()
```
janden, line 180: What's going on here?

kymatio/backend/numpy_backend.py
```diff
@@ -1,24 +1,35 @@
...
+    if not _is_complex(x):
+        raise TypeError('The input should be complex.')
+
+
+def real_check(x):
+    if not _is_real(x):
+        raise TypeError('The input should be real.')
+
+
+def _is_complex(x):
```
janden, line 19: Why are these moved outside? Don't they depend on the `np` module we're using (regular NumPy vs. Jax)?

kymatio/backend/tensorflow_backend.py
```diff
@@ -1,49 +1,54 @@
...
+    return (x.dtype == np.float32) or (x.dtype == np.float64)
+   
+ 
+class TensorFlowBackend():
+    def __init__(self):
+        self.name = 'tensorflow'
+        self.modulus = Modulus()
+  
+
+    def concat(self, arrays, dim):
```
janden, line 50: Why not `concatenate` like we have elsewhere?

kymatio/backend/torch_backend.py
```diff
@@ -3,163 +3,167 @@
...
+
+
+def input_checks(x):
+    if x is None:
+        raise TypeError('The input should be not empty.')
+
+    contiguous_check(x)
+
+
+def complex_check(x):
```
janden, line 134: Why are these not in the class?

kymatio/scattering1d/backend/jax_backend.py
```diff
@@ -0,0 +1,5 @@
+
```
janden, line 1: Are we including the Jax backend in this PR?

kymatio/scattering1d/backend/numpy_backend.py
```diff
@@ -1,12 +1,12 @@
 # Authors: Edouard Oyallon, Joakim Anden, Mathieu Andreux
 BACKEND_NAME = 'numpy'
 
-from ...backend.numpy_backend import NumpyBaseBackend
+from ...backend.numpy_backend import NumpyBackend, complex_check, real_check
```
janden, line 4: If checks are included in the class, we wouldn't need these extra imports here.

tests/scattering1d/test_torch_backend_1d.py
```diff
@@ -173,19 +173,22 @@ def test_subsample_fourier(backend, device, random_state=42):
...
 
 
-def test_unpad():
-    # test unpading of a random tensor
-    x = torch.randn(8, 4, 1)
+@pytest.mark.parametrize("backend", backends)
+@pytest.mark.parametrize("device", devices)
+def test_unpad(backend, device):
+    if backend.name == "torch_skcuda" and device == "cpu":
+        pytest.skip()
```
janden, line 180: Why are we adding this here? Can avoid these skips using the backend–device tuple scheme.

tests/scattering1d/test_torch_scattering1d.py
```diff
@@ -13,7 +13,7 @@
     if torch.cuda.is_available():
         from skcuda import cublas
         import cupy
-        skcuda_available = True
+#        skcuda_available = True
```
janden, line 16: ?

kymatio/backend/numpy_backend.py
MuawizChaudhary: We use them in the frontend. Should I just use the back end variable to call the appropriate function in the frontend?

kymatio/backend/tensorflow_backend.py
MuawizChaudhary: so we call this function in the other files. 

before, when we had the namedtuple convention we would be able to do this:
`backend.concatenate = lambda x: concatenate(x, -2)`

Now, from the inheritance of this backend here, we have to worry about this:

`def concatenate(self, x):
    return self.concatenate(x, -2)`

this makes the code a bit scary, I don't know whats going on, additionally it was crashing with this convention I described above.

kymatio/backend/torch_backend.py
MuawizChaudhary: called in frontend

kymatio/backend/torch_backend.py
```diff
@@ -1,230 +1,242 @@
...
+    def contiguous_check(self, x):
+        if not x.is_contiguous():
+            raise RuntimeError('Tensors must be contiguous.')
+    
 
-    if B.device.type == 'cuda':
-        if A.device.type == 'cuda':
-            if A.device.index != B.device.index:
-                raise TypeError('Input and filter must be on the same GPU.')
+    def cdgmm(self, A, B, inplace=False):
```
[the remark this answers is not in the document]
MuawizChaudhary: that was another PR that hasn't been merged yet.

kymatio/scattering1d/backend/jax_backend.py
MuawizChaudhary: My bad.

kymatio/scattering1d/backend/numpy_backend.py
MuawizChaudhary: True.

kymatio/scattering1d/backend/numpy_backend.py
```diff
@@ -1,115 +1,126 @@
...
-    """
-    if (pad_left >= x.shape[-1]) or (pad_right >= x.shape[-1]):
-        raise ValueError('Indefinite padding size (larger than tensor).')
+from ...backend.numpy_backend import NumpyBaseBackend
+
+
+class Numpy1DBackend(NumpyBaseBackend):
+    def __init__(self, np):
+        super(Numpy1DBackend, self).__init__(np)
+        self.np = np
```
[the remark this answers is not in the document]
MuawizChaudhary: Correct.

kymatio/scattering1d/backend/torch_backend.py
```diff
@@ -3,129 +3,122 @@
...
+from ...backend.torch_backend import TorchBaseBackend
 
 
-def subsample_fourier(x, k):
-    """Subsampling in the Fourier domain
+class Torch1DBackend(TorchBaseBackend):
+    def __init__(self):
+        super(Torch1DBackend, self).__init__()
+        self.name = 'torch'
+        pass
```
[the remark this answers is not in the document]
MuawizChaudhary: not needed.

tests/scattering1d/test_torch_scattering1d.py
MuawizChaudhary: Just temporary.

tests/scattering1d/test_torch_backend_1d.py
MuawizChaudhary: I'll be pushing it soon enough.

kymatio/backend/torch_backend.py
```diff
@@ -1,230 +1,242 @@
...
+            Returns
+            -------
+            output : tensor
+                A tensor with the same dimensions as x, such that output[..., 0]
+                contains the complex modulus of x, while output[..., 1] = 0.
+        """
+        def __call__(self, x):
+            complex_contiguous_check(x)
+            norm = modulus(x)[..., None]
+            return norm
```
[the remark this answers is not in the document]
MuawizChaudhary: I'll create an issue!

kymatio/backend/torch_backend.py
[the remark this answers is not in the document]
MuawizChaudhary: Actually you are right.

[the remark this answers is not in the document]
MuawizChaudhary: I based this PR based off of the 1d test pr. The 1d PR had not been rebased ontop of the change. How do we want to proceed?

kymatio/backend/numpy_backend.py
eickenberg: I don't understand this answer or its implications.

I guess the question is whether jax can handle numpy dtype objects or not

kymatio/backend/numpy_backend.py
```diff
@@ -1,78 +1,87 @@
...
-        raise TypeError('The second input must be complex or real.')
-
-    if inplace:
-        return np.multiply(A, B, out=A)
-    else:
-        return A * B
+    return (x.dtype == numpy.float32) or (x.dtype == numpy.float64)
+
+    
+class NumpyBackend():
```
eickenberg, line 27: parentheses not necessary

kymatio/backend/numpy_backend.py
```diff
@@ -1,78 +1,87 @@
...
+    def __init__(self, np):
+        self.np = np
+        self.name = 'numpy'
```
eickenberg, line 30: so: The numpy object is passed in, but the `self.name` will be inherited?

will we also be having a cupy backend then?

kymatio/backend/tensorflow_backend.py
```diff
@@ -1,23 +1,5 @@
...
-    if not _is_real(x):
-        raise TypeError('The input should be real.')
-
-def _is_complex(x):
-    return (x.dtype == np.complex64) or (x.dtype == np.complex128)
-
-
-def _is_real(x):
-    return (x.dtype == np.float32) or (x.dtype == np.float64)
+import tensorflow as tf
```
eickenberg, line 2: Do we need to put the tf backend into a class? What inherits from it? Can't it just stay as is?

kymatio/backend/tensorflow_backend.py
eickenberg: I don't follow

kymatio/backend/tensorflow_backend.py
```diff
@@ -41,36 +23,60 @@ def __call__(self, x):
...
 
-    if not _is_complex(B) and not _is_real(B):
-        raise TypeError('The second input must be complex or real.')
 
-    return A * B
+def _is_real(x):
+    return (x.dtype == np.float32) or (x.dtype == np.float64)
+   
+ 
+class TensorFlowBackend():
```
eickenberg, line 44: parentheses not necessary

kymatio/backend/torch_backend.py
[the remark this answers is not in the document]
eickenberg: such that the end result is that there is no more `inplace` stuff. I would suggest removing it here. Otherwise it might end up showing up again when this gets merged

kymatio/scattering1d/backend/jax_backend.py
```diff
@@ -0,0 +1,5 @@
...
+from .numpy_backend import NumpyBackend1D
+import jax.numpy 
+
+backend = NumpyBackend1D(jax.numpy)
```
eickenberg, line 5: This thing, according to what I saw above will still have `self.name` be `numpy`

kymatio/scattering1d/backend/numpy_backend.py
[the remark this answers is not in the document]
eickenberg: can't we pass in the numpy/jax object via inheritance?

kymatio/scattering1d/backend/numpy_backend.py
```diff
@@ -1,115 +1,126 @@
...
+        -------
+        res : tensor
+            The input tensor periodized along the next to last axis to yield a
+            tensor of size x.shape[-2] // k along that dimension.
+        """
+        self.complex_check(x)
+    
+        y = x.reshape(-1, k, x.shape[-1] // k)
+    
+        res = y.mean(axis=(-2,))
```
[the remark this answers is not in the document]
eickenberg: `axis=-2`

kymatio/scattering1d/backend/tensorflow_backend.py
```diff
@@ -1,109 +1,104 @@
...
-backend.pad = pad
-backend.rfft = rfft
-backend.irfft = irfft
-backend.ifft = ifft
-backend.concatenate = lambda x: concatenate(x, -2)
+import tensorflow as tf
+from ...backend.tensorflow_backend import TensorFlowBackend, complex_check, real_check
+
+
+class TensorFlowBackend1D(TensorFlowBackend):
```
eickenberg, line 7: again, why do we need this for tf?

kymatio/scattering1d/backend/tensorflow_backend.py
```diff
@@ -1,109 +1,104 @@
...
+        -------
+        res : tensor
+            The input tensor periodized along the next to last axis to yield a
+            tensor of size x.shape[-2] // k along that dimension.
+        """
+        complex_check(x)
+    
+        y = tf.reshape(x, (-1, k, x.shape[-1] // k))
+    
+        return tf.reduce_mean(y, axis=(-2,))
```
eickenberg, line 36: `axis=-2`

kymatio/backend/tensorflow_backend.py
janden: Just for consistency.

tests/scattering1d/test_numpy_backend_1d.py
```diff
@@ -0,0 +1,117 @@
+import pytest
```
eickenberg, line 1: are these tests from this pr or from the tests refactoring pr? I don't see why the tests would change with this refactoring

kymatio/backend/numpy_backend.py
janden: > the self.name will be inherited?

That's probably not a good idea. (What is `backend.name` used for, anyhow?)

> will we also be having a cupy backend then?

Should be relatively straightforward if we want to go there (don't think we do, however).

kymatio/backend/torch_backend.py
janden: Same here.

kymatio/backend/tensorflow_backend.py
janden: The second one won't work since inheritance in Python clobbers the method in the superclass (I think). How about this?
```
def contatenate(self, arrays):
    return tf.stack(arrays, dim=1)
```
The way we've set things up, the second dimension (`dim = 1`) is always the coefficient dimension, so we always want to concatenate along there, independent of the dimension of the signals. So there's only one `concatenate` that is used by all the backends.

tests/scattering1d/test_numpy_backend_1d.py
MuawizChaudhary: This is from the test refactoring PR. We should merge those as soon as possible.

kymatio/scattering1d/backend/numpy_backend.py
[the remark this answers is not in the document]
janden: Yes that's what I meant to say. Pass it to the constructor in the superclass, then we don't have to assign it here.

kymatio/backend/torch_backend.py
janden: Can't it get them from the backend? I agree that these functions belong more to the frontend, but they're currently organized into the backend, so let's keep them there.

tests/scattering1d/test_torch_scattering1d.py
janden: Ok.

kymatio/scattering1d/backend/tensorflow_backend.py
janden: He means the class structure. I think it's good to have for consistency between the backends, but it's not strictly necessary. (Although it would simplify things if we have `TensorFlowBackend` inherit from `NumPyBackend`, since they use the same input check functions.)

eickenberg:
I'd like to discuss today the differences between making changes by subclassing and making changes by passing a constructor argument.

kymatio/backend/tensorflow_backend.py
janden: 3D is a world unto itself, I'm afraid. It may warrant a special `concatenate_3d` that does something different.

tests/scattering1d/test_torch_backend_1d.py
MuawizChaudhary: could you explain this for me during the meeting?

janden:
@MuawizChaudhary Can you rebase this on top of dev? There's a lot of conflicts that need to be resolved and I don't want us to have to review it twice.

MuawizChaudhary:
If you squashed all the changes, I'll re base it shortly!

janden:
> If you squashed all the changes, I'll re base it shortly!

The changes in dev from the 1D test PR? Yeah they're squashed.

janden:
Tests fail. Please fix and we'll review.

kymatio/scattering1d/backend/tensorflow_backend.py
eickenberg: Ah is this why I see tf backend inherit from numpy backend?
How sure are we that this works and doesn't expose a lot of numpy functionality in the tf backend?
I don't even see the input check functions work properly because I don't see where the np object is set.
Is this passing tests? Are there even tests checking these things?

kymatio/backend/numpy_backend.py
```diff
@@ -1,73 +1,81 @@
...
-        raise RuntimeError('The inputs are not compatible for '
-                           'multiplication.')
-
-    if not _is_complex(B) and not _is_real(B):
-        raise TypeError('The second input must be complex or real.')
-
-    return A * B
+   
+class NumpyBackend:
+    def __init__(self, name='numpy'):
```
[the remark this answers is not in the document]
MuawizChaudhary: We need it so that we can specify whats the name of the backend.  We dont have to if its nD numpy backend since we already specified the name here.

kymatio/backend/numpy_backend.py
```diff
@@ -1,73 +1,81 @@
...
+        import numpy
```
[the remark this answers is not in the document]
MuawizChaudhary: I believe it shouldn't trigger a numpy import unless we call the init functions. 

I thought the best idea was to import it inline, I believe that might have been what @janden wanted.

kymatio/backend/tensorflow_backend.py
```diff
@@ -1,72 +1,43 @@
...
+    def __init__(self):
+        super(TensorFlowBackend, self).__init__(name='tensorflow')
 
-def real_check(x):
-    if not _is_real(x):
-        raise TypeError('The input should be real.')
+    def concatenate(self, arrays):
+        return tf.stack(arrays, axis=1)
 
-def _is_complex(x):
```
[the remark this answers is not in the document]
MuawizChaudhary: This functionality is in the numpy backend.

kymatio/backend/tensorflow_backend.py
```diff
@@ -1,72 +1,43 @@
+from .numpy_backend import NumpyBackend
 import tensorflow as tf
-import numpy as np
 
 
-def complex_check(x):
-    if not _is_complex(x):
-        raise TypeError('The input should be complex.')
+class TensorFlowBackend(NumpyBackend):
```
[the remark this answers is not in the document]
MuawizChaudhary: We inherit for the complex and real check functions. tf and np share the same code. 

I over ride the fft functions in the 1d backend.

kymatio/backend/tensorflow_backend.py
```diff
@@ -1,72 +1,43 @@
...
+    def __init__(self):
+        super(TensorFlowBackend, self).__init__(name='tensorflow')
```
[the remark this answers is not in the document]
MuawizChaudhary: the line above?

kymatio/backend/tensorflow_backend.py
```diff
@@ -1,72 +1,43 @@
...
+        if not self._is_complex(A):
+            raise TypeError('The first input must be complex.')
+    
+        if A.shape[-len(B.shape):] != B.shape[:]:
+            raise RuntimeError('The inputs are not compatible for multiplication.')
+    
+        if not self._is_complex(B) and not self._is_real(B):
+            raise TypeError('The second input must be complex or real.')
+    
+        return A * B
```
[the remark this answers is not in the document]
MuawizChaudhary: You have a point.

kymatio/backend/torch_backend.py
```diff
@@ -115,111 +95,116 @@ def backward(ctx, grad_output):
...
+    def __init__(self):
+        self.name = 'torch'
+ 
+    def input_checks(self, x):
+        if x is None:
+            raise TypeError('The input should be not empty.')
+    
+        self.contiguous_check(x)
+    
+    def complex_check(self, x):
```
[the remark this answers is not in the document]
MuawizChaudhary: I feel like this is a smart idea.

kymatio/backend/torch_skcuda_backend.py
```diff
@@ -2,82 +2,83 @@
...
-        m, n = B.nelement() // 2, A.nelement() // B.nelement()
-        lda = m
-        ldc = m
-        incx = 1
-        handle = torch.cuda.current_blas_handle()
-        stream = torch.cuda.current_stream()._as_parameter_
-        cublas.cublasSetStream(handle, stream)
-        cublas.cublasCdgmm(handle, 'l', m, n, A.data_ptr(), lda, B.data_ptr(), incx, C.data_ptr(), ldc)
-        return C
+class TorchSKcudaBackend:
```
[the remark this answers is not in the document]
MuawizChaudhary: we inherit the torch backend from the 1d torch backend when we create the TorchSKcuda 1D backend class. The torch skcuda base backend is also inherited.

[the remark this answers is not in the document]
MuawizChaudhary: I recommend checking `kymatio/scattering1d/backend/torch_skcuda_backend.py`.

kymatio/scattering1d/backend/numpy_backend.py
```diff
@@ -1,116 +1,98 @@
...
-    real_check(x)
-    return scipy.fftpack.fft(x)
-
-
-def irfft(x):
-    complex_check(x)
-    return scipy.fftpack.ifft(x).real
+        y = x.reshape(-1, k, x.shape[-1] // k)
+    
+        res = y.mean(axis=-2)
```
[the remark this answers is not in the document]
janden: This is what was done before so it's likely correct.

tests/general/test_torch_backend.py
```diff
@@ -1,16 +1,16 @@
...
-from kymatio.backend.torch_backend import ModulusStable, modulus
+from kymatio.backend.torch_backend import ModulusStable, TorchBackend
 
 
 def test_modulus(random_state=42):
     """
     Tests the stability and differentiability of modulus
     """
-
+    backend = TorchBackend()
```
janden, line 10: Why not just import `backend` from `kymatio.backend.torch_backend`? It's already been instantiated and that's what the main code will use.

tests/scattering1d/test_tensorflow_backend_1d.py
```diff
@@ -1,6 +1,6 @@
 import pytest
 import numpy as np
-
+import tensorflow as tf
```
janden, line 3: Why do we need this?

tests/scattering1d/test_tensorflow_backend_1d.py
```diff
@@ -106,14 +106,11 @@ def coefficent(n):
     y_r = (x_r * coefficents).sum(-1)
 
     z = backend.rfft(x_r)
-    # increase tolerance here as tensorflow fft is slightly inaccurate due to
```
janden, line 109: Why is this going away?

tests/scattering1d/test_torch_backend_1d.py
janden: We'd have a list of compatible backends and devices and loop over those. So if skcuda is available, we'd have the list: `[(torch_backend, 'cpu'), (torch_backend, 'gpu'), (torch_skcuda_backend, 'gpu')` and parametrize the tests with that. This is implemented in the 2D Torch tests. No need to do it right now, but we should probably do this in the 1D and 3D tests as well.

janden (dismissed):
Looks good. Just a few things to address, then we can extend this to 2D and 3D.

eickenberg:
OK, I just rebased this.

I will need some help to figure out what the intention was in some of these changes.

The global picture is that we are introducing classes for backends instead of the named tuples.

Please take a look at the changes I needed to make in my first commit on top of this to make the tests for numpy backend 2d pass: https://github.com/kymatio/kymatio/pull/631/commits/8ac8a8b42794881bb8c0aa8661bdec29aec4c482

basically I had to write a bunch of attributes to the instantiated class. I left the original commented for comparison.

Question: Why is it that we are defining half of the backend in the class and half in the 2d backend file? Is it because those parts are the specific stuff? Should we potentially be doing this using inheritance?

Happy for any input and opinions @janden @MuawizChaudhary

eickenberg:
OK, I have kludged my way to making the tests pass.

For the specific backends I would suggest using inheritance rather than writing properties, since that is not much better than the named tuples we had before.

Please let me know if you think that makes sense and I'll make those changes. Ideally we get this one merged before the sprint (or at the latest at the beginning of it).

janden:
Still fails, but for legit reasons (torch_skcuda backend has not been updated to the new architecture).

janden:
So tests pass now on Jenkins. Since it's been a while, I want to go over everything again to see what needs to be done before merging.

eickenberg:
Awesome. I will read through the updated code now.
One thing that I have had simmering for a bit is the instantiation. Do we really need to instantiate these backends?
If not, we could set all the methods to `staticmethod` and remove the `self`objects because they are not needed

edouardoyallon:
the fft failure for python 3.6 *only*, with conda and tf sounds like some future promising fun!

kymatio/backend/numpy_backend.py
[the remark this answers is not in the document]
eickenberg: Yeah I think nothing is really stopping us from using a global import.

I'd generally also like to get away as much as possible from making an instance. Ideally all functions are static on the class and we use inheritance to pass down those static functions

kymatio/backend/tensorflow_backend.py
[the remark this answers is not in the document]
eickenberg: need to check whether we can unify cdgmm across dimensionalities

kymatio/backend/torch_backend.py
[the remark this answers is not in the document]
eickenberg: I guess this is the base class actually.
But we might be able to use one more level of inheritance to deal with the two different complex number formulations

kymatio/backend/torch_backend.py
```diff
@@ -115,111 +95,117 @@ def backward(ctx, grad_output):
...
 
-        Parameters
-        ---------
-        x : tensor
-            Complex torch tensor.
+    def complex_contiguous_check(self, x):
+        self.complex_check(x)
+        self.contiguous_check(x)
+    
+    def contiguous_check(self, x):
```
eickenberg, line 121: yeah I'm not happy with these unused `self` objects

kymatio/frontend/numpy_frontend.py
```diff
@@ -11,7 +11,7 @@ def scattering(self, x):
     def __call__(self, x):
         """This method is an alias for `scattering`."""
 
-        input_checks(x)
+        self.backend.input_checks(x)
```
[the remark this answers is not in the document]
eickenberg: you mean because `scattering` assumes checks to have been done?

tests/scattering1d/test_tensorflow_backend_1d.py
eickenberg: yeah why?

tests/scattering1d/test_torch_backend_1d.py
eickenberg: good idea

eickenberg:
@janden is it OK if I go through the backends and remove any unnecessary use of `self`in function definitions? I.e. functions that do not depend on state should not need a `self` object as input

kymatio/backend/numpy_backend.py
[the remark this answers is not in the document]
janden: > Yeah I think nothing is really stopping us from using a global import.

Agree. That's how it is for the other backends, as far as I can tell.

> I'd generally also like to get away as much as possible from making an instance. Ideally all functions are static on the class and we use inheritance to pass down those static functions

I remember inheriting static functions as being slightly tricky to deal with when it comes to inheritance, but this may be a false memory. Overall, I think it's simple enough to have these singleton instances since it allows us to store some internal state (like these module references).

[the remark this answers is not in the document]
eickenberg: why can't we just write those into the class?
I'm happy to keep any state that is necessary, but would avoid it if it's easy

janden:
@eickenberg So all the backends have been refactored now to have either class methods or stack methods. There are no object instantiations (as far as I can tell). Please have a look and let me know what you think.

kymatio/scattering3d/backend/tensorflow_backend.py
```diff
@@ -7,116 +7,109 @@
...
-backend.name = 'tensorflow'
-backend.cdgmm3d = cdgmm3d
-backend.rfft = rfft
-backend.ifft = ifft
-backend.modulus = Modulus()
-backend.modulus_rotation = modulus_rotation
-backend.compute_integrals = compute_integrals
-backend.concatenate = concatenate
+
+backend = TensorFlowBackend3D()
```
eickenberg, line 115: instantiation not necessary, correct?

tests/general/test_torch_backend.py
eickenberg: especially now, since it's actually not instantiated

eickenberg: though that said, it might be more modular to do it like above, just without the parentheses

tests/scattering1d/test_numpy_backend_1d.py
eickenberg: hmm,  i believe this is merged, so I'm surprised to see the diff below

eickenberg (commented):
This looks close to mergeable. At this point I think if we address all the remaining comments, we should be good to go

eickenberg (approved):
OK, I think I'm done for today.

I have addressed a number of outstanding code comments in the PR.

The remaining ones I am not sure how to address, but I have the feeling they are minor.

I did an interactive rebase in order to consolidate @MuawizChaudhary 's summer work with my commits from last week so that in aggregate the tests pass on that commit. Just in case this wasn't a good idea, I have the unsquashed branch fully backed up on my computer. You can also reconstruct the previous status quo by using what you have on your machine and adding my most recent commits on top.

There is one remaining puzzle: I found (and had commented on in the PR at some point) several changes that looked like they were from a previous test refactoring project. I had thought that these might be duplicates, but for most of them, I can't find a trace of a second version of them anywhere in the history. If I am right, then we should make sure to keep these commits, because they do work on the tests.

From my side this can be merged as is, since merging will just add the test commits, and the remaining comments are minor. However, it is also possible to address the comments and split off the TST commits in order to make this PR only about the backend refactor.

I am fine with either. Since it's the last sprint day tomorrow, maybe we can get this in and then try to add the jax backends?

eickenberg:
OK, as discussed with @janden, I squashed all the straggling TST commits together and hilariously all those commits boiled down to these very few lines removing the tolerance from tensorflow fft test https://github.com/kymatio/kymatio/pull/631/commits/a6dee636542e6f8f8ffa739b8be63f49f39fb857

I will now proceed to remove this commit, because we don't actually want that change.

janden:
:+1:

PR #640, opened 2021-01-12
Fix for multi-GPU scattering

anakin-datawalker:
This is a fix for multi-GPU scattering issues as previously raised. It was tested on torch 1.7, cuda 10.2 and 11.0

### DIFF
--- a/kymatio/scattering2d/frontend/torch_frontend.py
+++ b/kymatio/scattering2d/frontend/torch_frontend.py
@@ -33,7 +33,7 @@ class ScatteringTorch2D(ScatteringTorch, ScatteringBase2D):
             if not isinstance(c, int):
                 continue
 
-            self.phi[c] = self.register_single_filter(phi, n)
+            self.register_single_filter(phi, n)
             n = n + 1
 
         for j in range(len(self.psi)):
@@ -41,7 +41,7 @@ class ScatteringTorch2D(ScatteringTorch, ScatteringBase2D):
                 if not isinstance(k, int):
                     continue
 
-                self.psi[j][k] = self.register_single_filter(v, n)
+                self.register_single_filter(v, n)
                 n = n + 1
 
     def load_single_filter(self, n, buffer_dict):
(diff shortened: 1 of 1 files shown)


edouardoyallon:
The error message is slightly weird ( https://jenkins.flatironinstitute.org/job/kymatio/job/PR-640/1/console ) but as travis is working, I guess it's mainly linked to CUDA.

janden:
Looks like it has to do with dependencies not installing properly. Since we haven't run the PR pipeline for a while, it's likely that some packages have been updated in the meantime. Will take a closer look.

kymatio/scattering2d/backend/torch_skcuda_backend.py
```diff
@@ -11,7 +11,7 @@
 from ...backend.torch_backend import _is_complex
 from ...backend.torch_skcuda_backend import cdgmm
 
-@cupy.util.memoize(for_each_device=True)
+@cupy._util.memoize(for_each_device=True)
```
janden, line 14: (Next time, it's probably better to put this in a separate PR, but this is good for now.)

janden (approved):
This looks good. Just need to sort out the testing issues and see if it passes.

janden:
Note: Changed merge branch to dev.

janden:
Once #641 is merged, we'll rebase this one on top and try again. That should make the Jenkins issue go away.

janden:
Rebased this onto current dev and reworded the commit message. Let's see how this does.

janden:
Looks like it's going through. @anakin-datawalker can you provide a short snippet that demonstrates multi-GPU usage?

anakin-datawalker:
Sure. I had provided the following code in the original issue, which raises the error "TypeError: Input and filter must be on the same GPU." when executed on current kymatio master but which runs correctly with the proposed PR. Would that be sufficient ?

```
import torch
from kymatio.torch import Scattering2D
scattering = Scattering2D(J=2, shape=(32, 32), max_order=1)
scattering = torch.nn.DataParallel(scattering).cuda()
x = torch.rand(8,8,32,32).cuda()
scattering(x)
```
Best

janden:
That should be sufficient. Let me try it out. What did you say the minimum version of Torch was for which FFT supports DataParallel?

anakin-datawalker:
I am almost sure it was torch 1.5.

eickenberg:
This change looks very good to me, but I am having slight troubles fully understanding what is going on.

What I would expect is that there was an error of the type that we had filters registered as buffers, but we collected them in a dictionary for ease of manipulation. Then when it came to using built-in pytorch mechanisms to transport to different GPUs, e.g. `DataParallel`, the dictionary did not get updated and its contents still pointed to the tensors on the original GPU.

I see that one correction is not to write the registered buffer to the dictionary, and probably to create the dictionary when requested, but I can't connect all the dots here from looking at the code.
Is my assessment from above correct, or is there something else at play? If my assessment is correct, then how have we survived the equivalent problem of going from cpu to gpu?

edouardoyallon:
> This change looks very good to me, but I am having slight troubles fully understanding what is going on.
> 
> What I would expect is that there was an error of the type that we had filters registered as buffers, but we collected them in a dictionary for ease of manipulation. Then when it came to using built-in pytorch mechanisms to transport to different GPUs, e.g. `DataParallel`, the dictionary did not get updated and its contents still pointed to the tensors on the original GPU.
> 
> I see that one correction is not to write the registered buffer to the dictionary, and probably to create the dictionary when requested, but I can't connect all the dots here from looking at the code.
> Is my assessment from above correct, or is there something else at play? If my assessment is correct, then how have we survived the equivalent problem of going from cpu to gpu?

Me neither but I guess wes hould acknowledge both pytorch fix and @anakin-datawalker 's work. If they're using this on their cluster, since it at least makes the code slightly more clean, then I'm in favor to merge it.

PR #641, opened 2021-01-13
TST Fix Jenkins

janden:
This PR gets Jenkins up and running again. There are four main contributions:
- Upgrade `pip` before calling `pip install`. This ensures that we can take advantage of the newer `manylinux2010` and `manylinux2014` wheels. The default version (v9) only installs `manylinux1` wheels and when it can't find them, tries to compile from source, which is a bad idea.
- Invoke `pip` using `python3 -m pip` instead of `pip3` since this is a more robust approach.
- Output version information for `pip` and the installed packages. This should help debugging in the future.
- Fixes usage of internal `cupy` components, which have moved from `cupy.utils` to `cupy._utils` in v8.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -17,6 +17,8 @@ pipeline {
 	HOME = "$WORKSPACE/build"
       }
       steps {
+	sh 'python3 -m pip --version'
+	sh 'python3 -m pip freeze'
 	sh 'python3 -c "import torch; torch.cuda.current_device()"'
 	sh 'python3 -c "import tensorflow as tf; tf.test.is_gpu_available()"'
 	sh 'python3 -m venv --system-site-packages --without-pip $HOME'
--- a/kymatio/scattering1d/backend/torch_skcuda_backend.py
+++ b/kymatio/scattering1d/backend/torch_skcuda_backend.py
@@ -6,7 +6,13 @@ from ...backend.torch_backend import contiguous_check, complex_check
 
 BACKEND_NAME = 'torch_skcuda'
 
-@cupy.util.memoize(for_each_device=True)
+# As of v8, cupy.util has been renamed cupy._util.
+if hasattr(cupy, '_util'):
+    memoize = cupy._util.memoize
+else:
+    memoize = cupy.util.memoize
+
+@memoize(for_each_device=True)
 def load_kernel(kernel_name, code, **kwargs):
     code = Template(code).substitute(**kwargs)
     kernel_code = cupy.cuda.compile_with_cache(code)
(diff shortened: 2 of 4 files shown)


janden:
Looks like part of the issue is that the Jenkins Docker cache was cleared since our last successful run this summer. Now it's downloading all the packages and reinstalling them. The reason it fails is that `scikit-learn` dropped support for the `manylinux1` platform in v0.24. This wouldn't normally be a problem, but we're currently using the system version of `pip`, which seems to be v9.0.1, which doesn't support the newer `manylinux2010` and `manylinux2014` platforms. Upgrading `pip` prior to installing the packages should let us use the newer binary wheels, so compilation should no longer be necessary.

janden:
Code coverage goes down because of unexecuted lines when `cupy.util` doesn't exist, but this shouldn't be a big deal. (We should look for ways of dealing with this with special pragma tags, though.)

edouardoyallon:
> * Fixes usage of internal `cupy` components, which have moved from `cupy.utils` to `cupy._utils` in v8.

I'm slightly concerned by this one. In general(to my limited knowledge) the underscore of "_utils" implies that one is doing a dirty trick.. no?

janden:
> I'm slightly concerned by this one. In general(to my limited knowledge) the underscore of "_utils" implies that one is doing a dirty trick.. no?

It is (and was). Hopefully we'll be able to purge all of this stuff once Torch supports complex numbers.

edouardoyallon:
> > I'm slightly concerned by this one. In general(to my limited knowledge) the underscore of "_utils" implies that one is doing a dirty trick.. no?
> 
> It is (and was). Hopefully we'll be able to purge all of this stuff once Torch supports complex numbers.

Torch does now.

edouardoyallon (approved):
It's good to me.

eickenberg:
Switching `cupy.util` to `cupy._util` means that they are making `util` officially a private module. This will mean that they get to make any change they want at any time without warning. We'll need to keep our eyes open for these changes and ideally transition away from using it in the medium term.

janden:
Right. My point is that this has always been the case. The other function we use in cupy is `cupy.cuda.compile_with_cache` which is undocumented and similarly subject to change (it's not part of the official API).

The goal here is just to get the code running as it was before with the newer version of cupy. Long-term, I agree: we should get rid of this dependency on cupy. Hopefully that will be possible once we move to complex numbers in Torch.

janden:
I'm going to merge this so we can get started on #640 (and other PRs). I'll make an issue on cupy flakiness and we can continue the discussion there.

eickenberg:
Yeah I forgot to say that this also looks good to go for me. Thanks for taking care of this!

PR #651, opened 2021-02-10
TST Try GH actions for pip tests

janden:
Adds configuration files for two workflows: one for pip install and one for conda install. Ideally, it should just pop up as a check in the PR, but we'll see.

### DIFF
--- a/.github/workflows/conda.yml
+++ b/.github/workflows/conda.yml
@@ -1,0 +1,43 @@
+name: Kymatio Conda CI
+
+on:
+    - push
+    - pull_request
+
+jobs:
+    build:
+        runs-on: ubuntu-latest
+
+        strategy:
+            matrix:
+                python-version: [3.5, 3.6, 3.7]
+
+        steps:
+            - uses: actions/checkout@v1
+            - name: Add Conda to path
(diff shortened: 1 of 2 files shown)


janden:
@edouardoyallon @eickenberg If this looks good to you, I suggest we merge it ASAP to get the GH Actions CI working for the other PRs.

eickenberg:
This is great. Thanks for setting this up!

Is there any way we can test over several pytorch versions?

If that is not straight-forward, let's just merge this in, because it's orthogonal to everything else

eickenberg:
Is merging into dev useful here? Will the github actions be triggered on that already or do we need to merge into master?

eickenberg (approved):
approved

janden:
> Is there any way we can test over several pytorch versions?

Yes but this takes more time, obviously, since we have limited capacity. My vote would be to remove the conda tests, but we can discuss that later. Another possibility is if we could have more tests run in parallel, which might be relevant once we get self-hosted runners going on @edouardoyallon's machine.

> Is merging into dev useful here? Will the github actions be triggered on that already or do we need to merge into master?

I don't know. We're about to find out. They show up on my fork: https://github.com/janden/kymatio/pull/3

PR #652, opened 2021-02-10
[DO NOT CLOSE AND DO NOT MERGE] Adding pytorch complex numbers

### LINKED ISSUE #462: Separate padding from casting to complex

edouardoyallon:
Fix #462 

I'm replacing in the torch backend using https://pytorch.org/docs/stable/complex_numbers.html

- [X] .contiguous() isn't really required (thought we could cosntraint a function to output a contiguous tensor if a contiguous tensor is passed
- [x] autograd doesn't work for some reasons linked to pytorch (to my understanding)
- [X] 1-2-3D
- [ ] deleting the torch skcuda backend if the speed is favorable (or close to) - Closes #327
- [X] casting in complex numbers is done separately from padding


eickenberg (commented):
Can you rebase this on to the backend refactoring PR? We'll need to merge that first.

As it stands, I am a bit hesitant about this change for the following reason:

- using torch 1.7 all the tests pass with the real view implementation. this means that the real view is forward and backward compatible.
- differentiability is broken? Seems like we should wait until this is fixed?
- all of the complex number stuff seems to be internal. It never interacts with the user. Shouldn't this change come with e.g. the `WaveletTransform` module, where complex numbers are user-facing? Seems more compelling.

I will look into ideas to isolate the breaking points and writing helpers into two files that can be selectively imported to solve the problem transparently.

For now this PR should either be rebased on the backend refactor wait until that one is merged and then rebase.

edouardoyallon:
> Can you rebase this on to the backend refactoring PR? We'll need to merge that first.
ok

> As it stands, I am a bit hesitant about this change for the following reason:
> 
> * using torch 1.7 all the tests pass with the real view implementation. this means that the real view is forward and backward compatible.
> * differentiability is broken? Seems like we should wait until this is fixed?

It's fixed  and was just a bug-- we just needed to use the dev version of pytorch.

> * all of the complex number stuff seems to be internal. It never interacts with the user. Shouldn't this change come with e.g. the `WaveletTransform` module, where complex numbers are user-facing? Seems more compelling.

I think it's simpler to add first those complex numbers

> I will look into ideas to isolate the breaking points and writing helpers into two files that can be selectively imported to solve the problem transparently.
> 
> For now this PR should either be rebased on the backend refactor wait until that one is merged and then rebase.

My view on it is that it shoudln't be more different than numpy (which is using native complex) A real improvement, which I will not implement, would be to use the hermitian symmetry. This would lead to significant speed up but yet it is not friendly with phase harmonics strategies.

janden:
> It's fixed and was just a bug-- we just needed to use the dev version of pytorch.

Ok great. So that means that we can't use complex numbers until 1.8 is out? We can't tell users to run pytorch dev.

edouardoyallon:
I'll leave this PR to check from time to time if pytorch became suddenly efficient for what we do.

janden:
> I'll leave this PR to check from time to time if pytorch became suddenly efficient for what we do.

Better to close the PR if we have no intention of merging it. If you want to leave it open on your personal repo, you can.

lostanlen:
@edouardoyallon @MuawizChaudhary shall we close ?

MuawizChaudhary:
Pro closing, its a bit stale and I think the code base in dev changed a bunch. If need be, we can recover commits to save on work.

janden:
> @edouardoyallon @MuawizChaudhary shall we close ?

Yes.

PR #657, opened 2021-02-10
MAINT Rename T → N in 1D

### LINKED ISSUE #656: T → N
As part of our discussion on separating the wavelet scale from the averaging scale, we realized that the variable T is already being used in 1D for the size of the signals. In the 2D code (and in scatnet, I think), this is denoted N, so I suggest we rename T to N. This would then free up T as the averaging scale parameter.

janden:
Currently, the signal length is denoted by T in 1D, which is easily
confused with the averaging scale 2 ** J and inconsistent with the
notation for 2D, where the shape is given by M, N. As a result, we
change the notation to N.

Closes #656.

### DIFF
--- a/kymatio/scattering1d/frontend/base_frontend.py
+++ b/kymatio/scattering1d/frontend/base_frontend.py
@@ -42,9 +42,9 @@ class ScatteringBase1D(ScatteringBase):
 
         # check the shape
         if isinstance(self.shape, numbers.Integral):
-            self.T = self.shape
+            self.N = self.shape
         elif isinstance(self.shape, tuple):
-            self.T = self.shape[0]
+            self.N = self.shape[0]
             if len(self.shape) > 1:
                 raise ValueError("If shape is specified as a tuple, it must "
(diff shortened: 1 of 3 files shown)


kymatio/scattering1d/frontend/base_frontend.py
```diff
@@ -42,9 +42,9 @@ def build(self):
 
         # check the shape
         if isinstance(self.shape, numbers.Integral):
-            self.T = self.shape
+            self.N = self.shape
         elif isinstance(self.shape, tuple):
-            self.T = self.shape[0]
+            self.N = self.shape[0]
```
lostanlen, line 47: so self.N is basically an alias for self.shape[0] throughout the library?

janden: Something like that. Would be good to investigate this further. Can you make an issue?

lostanlen: ok

PR #659, opened 2021-02-10
save kymatio configuration as a keras-compatible dict

lostanlen:
Fixes #637

PR #660, opened 2021-02-11
Update master with new testing regime and cupy bugfix

janden:
These commits are all cherry-picked from dev, but should go into the bugfix release 0.2.1 so we're bringing them into master now.


janden:
@lostanlen Question: How does this work with your fancy branching scheme? Do we need to update 0.2.X with this one or create a new 0.2.1 branch?

lostanlen:
There's no need for a bugfix `0.2.1` branch. Just update `0.2.X` IMO

janden:
Cool, but this still goes into master, right? Then we update 0.2.X to point to the new master. It's only when we start bringing in 0.3 changes that master will start diverging from 0.2.X.

lostanlen:
Yes. In a sense, now that we also have `dev`, master is purely cosmetic (but useful) at this point. it's only a synonym for "latest stable release with bugfixes".
To be clear, i didn't anticipate that we'll have a dev branch one day, so i imagined that master would serve as our dev ...

janden:
Ah I see. Cool. Anyhow looks like tests are good, so just need review and merge (?), then we can rebase #659 on top and see how it does.

Jenkinsfile
```diff
@@ -10,7 +10,7 @@ pipeline {
       agent {
 	dockerfile {
 	  dir 'tools'
-	  args '--gpus 2'
+	  args '--gpus 1'
 	}
```
edouardoyallon, line 14: Are we still using jenkins?

edouardoyallon (approved):

Jenkinsfile
janden: Yes we are.

janden: (We need it for the GPU tests.)

PR #664, opened 2021-02-11
MAINT Update Python requirements

janden:
We no longer support 3.5, so minimum version is 3.6.


eickenberg (approved):
This LGTM. I added an email address change. Unfortunately the commit message is a bit generic. Feel free to change that. Otherwise this is good to go.

PR #665, opened 2021-02-11
TST Combine GHA workflows into one

janden:
Simplifies displaying results and navigating.


edouardoyallon (approved):
looks good. Maybe we could add something in the test unit to see if the local machine has AVS and then run it with or without tf?

PR #673, opened 2021-03-04
Stride

pawarrick:
As discussed in the Kymatio Sprint 2021, these changes to the numpy 1-D frontend allow the low-pass filter sigma_low time extent to be specified with the new parameter T.  Note the previous use of "T" for the input signal extent has been changed to "N".  Two new test cases have been added to tests/scattering1d/test_numpy_scattering1d.py.


kymatio/scattering1d/core/scattering1d.py
```diff
@@ -73,8 +77,11 @@ def scattering1d(x, pad, unpad, backend, J, psi1, psi2, phi, pad_left=0,
     # compute the Fourier transform
     U_0_hat = rfft(U_0)
 
+    #T = 2**(J-5)
+    T_log2 = math.floor(math.log2(T))
```
lostanlen, line 81: i would prefer log2_T as a variable name

kymatio/scattering1d/filter_bank.py
```diff
@@ -547,10 +549,14 @@ def calibrate_scattering_filters(J, Q, r_psi=math.sqrt(0.5), sigma0=0.1,
                                             alpha=alpha)
     xi2, sigma2, j2 = compute_params_filterbank(sigma_low, 1, r_psi=r_psi,
                                             alpha=alpha)
+    ## PAW
+    #sigma_low = sigma_low * math.pow(2, 5)
+    sigma_low = sigma0/T
```
lostanlen, line 554: it's a bit confusing that sigma_low is defined twice in this function: once for calling compute_params_filterbank and a second time as a return value. Perhaps we should clarify what's going on here?

kymatio/scattering1d/core/scattering1d.py
```diff
@@ -8,7 +10,7 @@ def scattering1d(x, pad, unpad, backend, J, psi1, psi2, phi, pad_left=0,
     Parameters
     ----------
     x : Tensor
-        a torch Tensor of size `(B, 1, N)` where `N` is the temporal size
+        a torch Tensor of size `(B, 1, T)` where `T` is the temporal size
```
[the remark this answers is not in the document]
pawarrick: yes

kymatio/scattering1d/core/scattering1d.py
```diff
@@ -73,8 +77,11 @@ def scattering1d(x, pad, unpad, backend, J, psi1, psi2, phi, pad_left=0,
     # compute the Fourier transform
     U_0_hat = rfft(U_0)
 
+    #T = 2**(J-5)
+    T_log2 = math.floor(math.log2(T))
+    
     # Get S0
-    k0 = max(J - oversampling, 0)
+    k0 = max(T_log2 - oversampling, 0)
```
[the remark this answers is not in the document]
pawarrick: these max/min index limit constrainst were all thanks to @janden

kymatio/scattering1d/core/scattering1d.py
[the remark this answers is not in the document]
eickenberg: `N` comes from the input shape and `T` is the desired output length?

tests/scattering1d/test_numpy_scattering1d.py
```diff
@@ -27,9 +28,93 @@ def test_Scattering1D(self, backend):
...
+        T=2**(J-sigma_low_scale_factor)
+        scattering2 = Scattering1D(J, N, Q, T=T, backend=backend, frontend='numpy')
+        Sx2 = scattering2(x)
+        
+        # for now, just be sure that the output shape is different from default
+        # should we add this new result to the test data?
+        #print('Sx1 shape: ' + str(Sx1.shape) + ' Sx2.shape: ' + str(Sx2.shape))
+        assert Sx2.shape == (Sx1.shape[0], Sx1.shape[1], Sx1.shape[2]*2**(sigma_low_scale_factor))
+    
+    def Scattering1D_filter_factory_T_plot(self, N, J, T, phi_f, psi1_f, msg):
```
eickenberg, line 71: Can go into an example?

kymatio/scattering1d/core/scattering1d.py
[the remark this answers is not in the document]
lostanlen: `N` comes from the input shape. `T` is the length of `phi`. Thus the stride is (proportional to) `T` and the output length is (proportional to) `N/T`. Do you confirm @pawarrick ?

kymatio/scattering1d/filter_bank.py
pawarrick: I have clarified this in the coming revision

tests/scattering1d/test_numpy_scattering1d.py
pawarrick: ok sure, examples/1d is a better place for the plots

pawarrick: done

kymatio/scattering1d/core/scattering1d.py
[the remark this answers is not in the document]
pawarrick: yes exactly and the change in output length is what I test in test_numpy_scattering1d.test_Scattering1D_T()

kymatio/scattering1d/core/scattering1d.py
pawarrick: done

lostanlen (approved):

OverLordGoldDragon:
Was wondering where this is, surprised not already implemented - should work merging before or after JFTS.

lostanlen:
@OverLordGoldDragon could you please review this PR and tell us what you think?

OverLordGoldDragon:
@lostanlen On it. I'll familiarize to make merge with JFTS smoother. Think it's better to merge this first, then edit relevant parts in JFTS for compatibility; if this merges after, it'll break JFTS and we'll have to rebase it anyway. It'll also ease introducing the `F` parameter (max frequential invariant).

lostanlen:
@OverLordGoldDragon that makes sense. Stride is an important feature and a long-awaited one (by @pawarrick , @AC-29, et al.) so let's merge it sooner rather than later

OverLordGoldDragon:
Reviewed, [made PR](https://github.com/pawarrick/kymatio/pull/1) addressing everything except 3 and 5. 

 1. The `min(max(` logic on subsampling is unsafe against `oversampling`; think it's better to preserve current behavior than run input checks on `oversampling`.
 2. `T` lacks an input check; should it ever be allowed to exceed `2**J`, and by how much? If _yes_:
     - a. Should warn users of potentially excessive oversampling per padding input beyond `nextpow2(N)`
     - b. `J_tentative` should be set per `max(N, T)`
     - c. Seems subsampling factors are unaffected as such a lowpass is more bandlimited than any other filter
     - d. Docs will need changing
     - e. I decided "no" in my fork, seems much trouble without much use
 3. `t_max_phi` logic is incorrect: `phi_f[0]` is no longer necessarily the filter with greatest temporal support. This isn't cleanly resolved: Morlet has greater temporal than Gauss for same `sigma`, particularly for lower `xi`. One approach is to compute temporal support for both, Morlet at `sigma_min` and `xi_min`, and `phi_f[0]`, and take max.
 4.  `self.T` needs to be set only once in `ScatteringBase1D`

Style / docstrings:

 5. `_J` vs `_j` vs `_T`: there's a `k1_J` variable, and `# convolve with phi_J` comment, yet the filter is taken at `J - k1`, which is `j` rather than `J`. After this PR it's `log2_T`, which is too long to rename things to; instead, it should be `j`, as it already denotes the subsampling "depth" or "stage".
     - Applicable to `sigma0` docstring in `filter_bank.py`
     - This is a "style" change so skipped for now
 6. I think "for subsampling" is awkwardly worded, as subsampling is more a inclusive step than main purpose of `T` (which is to control amount of invariance), and an optional one at that.
 7.  `J_scattering` incorrect docs: _"2**J_scattering corresponds to the averaging support of the low-pass filter"_ no longer so.

OverLordGoldDragon (commented):
Reviewed, [made PR](https://github.com/pawarrick/kymatio/pull/1) addressing everything except 3 and 5. 

 1. The `min(max(` logic on subsampling is unsafe against `oversampling`; think it's better to preserve current behavior than run input checks on `oversampling`.
 2. `T` lacks an input check; should it ever be allowed to exceed `2**J`, and by how much? If _yes_:
     - a. Should warn users of potentially excessive oversampling per padding input beyond `nextpow2(N)`
     - b. `J_tentative` should be set per `max(N, T)`
     - c. Seems subsampling factors are unaffected as such a lowpass is more bandlimited than any other filter
     - d. Docs will need changing
     - e. I decided "no" in my fork, seems much trouble without much use
 3. `t_max_phi` logic is incorrect: `phi_f[0]` is no longer necessarily the filter with greatest temporal support. This isn't cleanly resolved: Morlet has greater temporal support than Gauss for same `sigma`, particularly for lower `xi`. One approach is to compute temporal support for both, Morlet at `sigma_min` and `xi_min`, and `phi_f[0]`, and take max.
 4.  `self.T` needs to be set only once in `ScatteringBase1D`

Style / docstrings:

 5. `_J` vs `_j` vs `_T`: there's a `k1_J` variable, and `# convolve with phi_J` comment, yet the filter is taken at `J - k1`, which is `j` rather than `J`. After this PR it's `log2_T`, which is too long to rename things to; instead, it should be `j`, as it already denotes the subsampling "depth" or "stage".
     - Applicable to `sigma0` docstring in `filter_bank.py`
     - This is a "style" change so skipped for now
 6. I think "for subsampling" is awkwardly worded, as subsampling is more an inclusive step than main purpose of `T` (which is to control amount of invariance), and an optional one at that.
 7.  `J_scattering` incorrect docs: _"2**J_scattering corresponds to the averaging support of the low-pass filter"_ no longer so.

OverLordGoldDragon:
1. Added correctness test to ensure `T` controls amount of invariance as intended.
     - unsure where to put test, created `test_correctness.py`
 2. `min(a, b) - c` cleaner than `min(a - c, b - c)`
 3. Small room for improvement for `out_type = "list"`: `min(j1, log2_T)` is such that max subsampling doesn't alias _lowpass's_ freq support. But this isn't needed: narrowest support is controlled by `psi[-1]` (when `T < 2**J`), so we lose no information by subsampling per `j1, j2` (within compact support tolerance). Though this barely saves space and adds code complexity - not advocating, just noting.

examples/1d/plot_filters_T.py
```diff
@@ -0,0 +1,59 @@
+import pytest
```
janden, line 1: Incorporate into `plot_filters.py`?

tests/scattering1d/test_numpy_scattering1d.py
```diff
@@ -27,9 +28,66 @@ def test_Scattering1D(self, backend):
...
+ 
+        # adjust T
+        sigma_low_scale_factor = 2
+        T=2**(J-sigma_low_scale_factor)
+        scattering2 = Scattering1D(J, N, Q, T=T, backend=backend, frontend='numpy')
+        Sx2 = scattering2(x)
+        
+        # for now, just be sure that the output shape is different from default
+        # should we add this new result to the test data?
+        #print('Sx1 shape: ' + str(Sx1.shape) + ' Sx2.shape: ' + str(Sx2.shape))
```
janden, line 68: Remove.

tests/scattering1d/test_numpy_scattering1d.py
```diff
@@ -27,9 +28,66 @@ def test_Scattering1D(self, backend):
...
+        for j in Js:
+            J = j
+            for i in sigma_low_scale_factor:
+                T=2**(J-i)
+                if i == 0:
+                    default_str = ' (default)'
+                else: 
+                    default_str = ''
+                phi_f, psi1_f, psi2_f, _ = scattering_filter_factory(np.log2(N), J, Q, T)
+                #msg = 'J=' + str(J) + ', T=' +str(T) + default_str + ': LP-filter width $\sigma_{low}$=' + str(phi_f['sigma']) + ' (=0.1/T)'
```
janden, line 90: Remove (and line below).

tests/scattering1d/test_numpy_scattering1d.py
```diff
@@ -27,9 +28,66 @@ def test_Scattering1D(self, backend):
...
+                T=2**(J-i)
+                if i == 0:
+                    default_str = ' (default)'
+                else: 
+                    default_str = ''
+                phi_f, psi1_f, psi2_f, _ = scattering_filter_factory(np.log2(N), J, Q, T)
+                #msg = 'J=' + str(J) + ', T=' +str(T) + default_str + ': LP-filter width $\sigma_{low}$=' + str(phi_f['sigma']) + ' (=0.1/T)'                
+                #print(msg)
+                assert(phi_f['sigma']==0.1/T)
+
```
janden, line 93: Remove blank line.

kymatio/scattering1d/frontend/base_frontend.py
```diff
@@ -51,9 +51,13 @@ def build(self):
                                  "have exactly one element")
         else:
             raise ValueError("shape must be an integer or a 1-tuple")
-        
+
         if self.T is None:
             self.T = 2**(self.J)
+        elif self.T > 2**(self.J):
+            raise ValueError("The temporal support T of the low-pass filter "
```
janden, line 58: Why?

janden (changes requested):

kymatio/scattering1d/frontend/base_frontend.py
OverLordGoldDragon: [See 2](https://github.com/kymatio/kymatio/pull/673#pullrequestreview-626744057). Should there be _no_ constraint on `T`? Unsure how this plays out with `T >> N`, but more work needs to be done.

janden: Constraint should be whatever was in place previously for `2 ** J`.

OverLordGoldDragon: Then that's what this accomplishes. The idea's with this PR is to allow `T != 2**J`, but implementation has catching up to do if `T > 2**J`.

janden: But having `T > 2 ** J` is desirable in certain cases (@lostanlen does this quite often, from what I recall). The constraint before was that `2 ** J` cannot be much larger than `N` (I forget what the exact limit was), that same constraint should be applied to `T` instead.

OverLordGoldDragon:
@janden The constraint right now is `2 ** J <= N`; if this is exceeded, we subsample more samples than exist.

I agree `T > 2**J` can be desired. What needs to happen:

  1. Limit on maximum possible subsampling (make `len(time) == 1`?)
  2. Pad by more than `N`; currently code claims this breaks things with PyTorch, but I don't know what that's about.
  3. Adjust `ind_start`, `ind_end` per 1 and 2

It's also clumsy for me to work through PRs to a PR, currently there's [one waiting](https://github.com/pawarrick/kymatio/pull/2/files) - maybe merge this first then I take it from there?

OverLordGoldDragon:
And what of _minimum_ `T`? If it's any `T > 0`, subsampling logic needs to account for the case of `supp(phi) > supp(psi[-1])`.

pawarrick:
I just added the stride support to Tensorflow and Keras front ends

OverLordGoldDragon:
This PR cannot complete without some of my other PRs and things I've not PR'd that are in JTFS since they rely on existing PRs being merged; can either lump them all into this PR or merge this first.

OverLordGoldDragon:
I will complete Stride logic for both `T` and `F` in JTFS.

OverLordGoldDragon:
[PR](https://github.com/pawarrick/kymatio/pull/3) makes `J` redundant in `core/scattering1d.py`, but I kept it for now.

Moved `log2_T` to frontend for consistency with JTFS which does it per necessity with `log2_F`.

OverLordGoldDragon:
@lostanlen Can `T` ever be `>N`? `N` is already global average pooling

lostanlen:
@lostanlen Can T ever be >N? N is already global average pooling

No, it's safe so assume that N<T.

janden:
So where are we on this? Is this ready for review? Are there any outstanding questions that we need to resolve?

OverLordGoldDragon:
@janden It's ready for review, but [cannot complete here](https://github.com/kymatio/kymatio/pull/673#issuecomment-830507875) (namely padding by more than `N` and computing said padding).

There is a question but can't act on it here: what to do in case of `T == N`? This is global averaging, which doesn't care whether we pad millions or none at all, so seems we should exclude it from computation of padding amounts and only use `psi`.

Might also be a few docs missing but no point on perfecting it here, it'll change later anyway.

OverLordGoldDragon:
> doesn't care whether we pad millions or none at all

it matters for `pad_mode != 'zero'` but not for boundary effects, this is a question of what prior to impose, and "more" is worse, so `T==N` shouldn't _increase_ padding determined from `psi` alone.

pawarrick:
@janden from the point of view of introducing the new T parameter to control the degree of time invariance, the main thing I see as missing is a torch test to mirror the new keras test in 
kymatio/tests/scattering1d/test_keras_scattering1d.py

OverLordGoldDragon:
@pawarrick Handled it; test failure is not related to this PR.

@janden A question is whether to test for output values with changed `T`; I think we should. I'd need the script generates `test_data_1d.npz`.

pawarrick:
@OverLordGoldDragon: excellent.  
Yes, I think having the output values for the changed T is a final thing to add to the keras and torch 1D tests.

kymatio/scattering1d/core/scattering1d.py
```diff
@@ -64,7 +68,7 @@ def scattering1d(x, pad, unpad, backend, J, psi1, psi2, phi, pad_left=0,
 
     # S is simply a dictionary if we do not perform the averaging...
     batch_size = x.shape[0]
-    kJ = max(J - oversampling, 0)
+    kJ = max(log2_T - oversampling, 0)
```
eickenberg, line 71: not directly related to this PR but this note is to point out that `kJ` means basically nothing to an uninformed reader. Should keep these to a minimum. 

I get from the docstring that it has something to do with determining the cropping at the end, but it would be great if this were clearer

kymatio/scattering1d/core/scattering1d.py
```diff
@@ -92,7 +96,8 @@ def scattering1d(x, pad, unpad, backend, J, psi1, psi2, phi, pad_left=0,
         # Convolution + downsampling
         j1 = psi1[n1]['j']
 
-        k1 = max(j1 - oversampling, 0)
+        sub1_adj = min(j1, log2_T) if average else j1
```
eickenberg, line 99: what does `sub1_adj` mean?

kymatio/scattering1d/core/scattering1d.py
```diff
@@ -107,7 +112,7 @@ def scattering1d(x, pad, unpad, backend, J, psi1, psi2, phi, pad_left=0,
 
         if average:
             # Convolve with phi_J
-            k1_J = max(J - k1 - oversampling, 0)
+            k1_J = max(log2_T - k1 - oversampling, 0)
```
eickenberg, line 115: `2 ** k1_J` is a stride I assume?

kymatio/scattering1d/core/scattering1d.py
```diff
@@ -92,7 +96,8 @@ def scattering1d(x, pad, unpad, backend, J, psi1, psi2, phi, pad_left=0,
         # Convolution + downsampling
         j1 = psi1[n1]['j']
 
-        k1 = max(j1 - oversampling, 0)
+        sub1_adj = min(j1, log2_T) if average else j1
+        k1 = max(sub1_adj - oversampling, 0)
 
         assert psi1[n1]['xi'] < 0.5 / (2**k1)
```
eickenberg, line 102: prod code doesn't use `assert` but rather raises specific errors

kymatio/scattering1d/filter_bank.py
```diff
@@ -409,14 +409,13 @@ def compute_xi_max(Q):
...
+def compute_params_filterbank(sigma_min, Q, r_psi=math.sqrt(0.5), alpha=5.):
     """
     Computes the parameters of a Morlet wavelet filterbank.
 
     This family is defined by constant ratios between the frequencies and
     width of adjacent filters, up to a minimum frequency where the frequencies
-    are translated.
-    This ensures that the low-pass filter has the largest temporal support
+    are translated. sigma_min specifies the smallest frequential width
     among all filters, while preserving the coverage of the whole frequency
```
eickenberg, line 419: not sure if the `while preserving the coverage[...]` still makes sense here

kymatio/scattering1d/filter_bank.py
```diff
@@ -426,11 +425,10 @@ def compute_params_filterbank(sigma_low, Q, r_psi=math.sqrt(0.5), alpha=5.):
 
     Parameters
     ----------
-    sigma_low : float
-        frequential width of the low-pass filter. This acts as a
-        lower-bound on the frequential widths of the band-pass filters,
-        so as to ensure that the low-pass filter has the largest temporal
-        support among all filters.
+    sigma_min : float
+        This acts as a lower-bound on the frequential widths of the band-pass
```
eickenberg, line 429: `lower bound`

eickenberg (commented):
not sure I can be of help reviewing this given current time constraints.

(by this I mean: I am not familiar enough with the codebase anymore to be able to decide high-level things. I am generally worried about the increasing differences to the other dimensionalities)

kymatio/scattering1d/core/scattering1d.py
```diff
@@ -28,6 +29,9 @@ def scattering1d(x, pad, unpad, backend, J, psi1, psi2, phi, pad_left=0,
         The array `phi[j]` is a real-valued filter.
     J : int
         scale of the scattering
+    log2_T : int
```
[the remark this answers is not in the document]
OverLordGoldDragon: `T` can be non-dyadic. `log2_T` is also reused so eases readability / searchability.

kymatio/scattering1d/core/scattering1d.py
OverLordGoldDragon: Agreed

kymatio/scattering1d/core/scattering1d.py
OverLordGoldDragon: "subsampling 1 adjusted", didn't think much on this

kymatio/scattering1d/core/scattering1d.py
OverLordGoldDragon: Yes

kymatio/scattering1d/core/scattering1d.py
OverLordGoldDragon: `assert` is dropped in production code (compiled with "optimize" flag) for performance, yes

I think asserts make sense here for readability, if all tests pass there's little chance of deployment failure. Though it'd help to _explain_ why we assert, e.g. `# stability theorem, see [paper]`

Also unrelated to this PR

kymatio/scattering1d/filter_bank.py
OverLordGoldDragon: Yes: if an implementation is CQT-only, limiting `sigma_min` limits `xi_min` and prevents tiling whole frequency axis.

janden:
> @janden It's ready for review, but [cannot complete here](https://github.com/kymatio/kymatio/pull/673#issuecomment-830507875) (namely padding by more than `N` and computing said padding).

So what's the problem, exactly? That a large `T` would require a large amount of padding that we don't currently support? How would this be different from `2 ** J` being large?

> Yes, I think having the output values for the changed T as a better assert is a final thing to add to the numpy, keras and torch 1D tests.

I don't know about this. Adding more binary blobs to the repo should only be done when absolutely necessary. Besides, that test is only there to make sure that the core code aligns with the reference data. A way around this would be to ensure that the `T = 2 ** J` where `J` is the value used for the reference data, then pick `J  - 1` and check even samples.

kymatio/scattering1d/core/scattering1d.py
```diff
@@ -1,4 +1,5 @@
-def scattering1d(x, pad, unpad, backend, J, psi1, psi2, phi, pad_left=0,
+
+def scattering1d(x, pad, unpad, backend, J, log2_T, psi1, psi2, phi, pad_left=0,
```
janden, line 2: `J` is no longer used and can be removed.

kymatio/scattering1d/core/scattering1d.py
janden, line 99: I keep stumbling over these. Maybe a clearer way to formulate things would be
```
k1 = max(j1 - oversampling, 0)
if average:
    k1 = min(k1, kJ)
```
Similar code would work for `k2` below.

The idea being that we want to downsample by `k1`, which is `j1 - oversampling` normally, but if we're averaging, we don't want to downsample by more than the final scale, which is `kJ`.

kymatio/scattering1d/filter_bank.py
janden: Ok so let's skip the last part then, and have `The sigma_min parameter specifies the smallest frequential width among all filters.`

pawarrick:
@janden, that solution that avoids another blob should work, I'll implement that.

kymatio/scattering1d/filter_bank.py
```diff
@@ -722,10 +734,10 @@ def scattering_filter_factory(J_support, J_scattering, Q, r_psi=math.sqrt(0.5),
         psi2_f[n2]['j'] = j2
     phi_f['xi'] = 0.
     phi_f['sigma'] = sigma_low
-    phi_f['j'] = 0
+    phi_f['j'] = log2_T
```
janden, line 737: Why?

kymatio/scattering1d/frontend/base_frontend.py
janden: Looks like this is resolved?

kymatio/scattering1d/frontend/base_frontend.py
```diff
@@ -66,12 +76,12 @@ def build(self):
         self.pad_left, self.pad_right = compute_padding(self.J_pad, self.N)
         # compute start and end indices
         self.ind_start, self.ind_end = compute_border_indices(
-            self.J, self.pad_left, self.pad_left + self.N)
+            self.log2_T, self.pad_left, 2**self.J_pad - self.pad_right)
```
janden, line 79: I'm not following the last argument here. Was this a bug previously?

tests/scattering1d/test_correctness.py
```diff
@@ -0,0 +1,62 @@
...
+
+def _l1l2(x):
+    return np.sum(np.sqrt(np.mean(x**2, axis=1)), axis=0)
+
+def l1l2(x0, x1):
+    """Coeff distance measure; Thm 2.12 in https://arxiv.org/abs/1101.2286"""
+    return _l1l2(x1 - x0) / _l1l2(x0)
+
+
+if __name__ == '__main__':
```
janden, line 58: Remove.

tests/scattering1d/test_keras_scattering1d.py
```diff
@@ -0,0 +1,60 @@
...
+
+        model1 = Model(inputs1, scat1)
+        model1.compile(optimizer='adam',
+                      loss='sparse_categorical_crossentropy',
+                      metrics=['accuracy'])
+        Sg1 = model1.predict(x)
+
+        # for now, just be sure that the output shape is different from default
+        # should we add this new result to the test data?
+        #print('Sx1 shape: ' + str(Sx1.shape) + ' Sx2.shape: ' + str(Sx2.shape))
```
janden, line 54: Remove.

tests/scattering1d/test_keras_scattering1d.py
```diff
@@ -0,0 +1,60 @@
...
+        assert Sg1.shape == (Sg0.shape[0], Sg0.shape[1], Sg0.shape[2]*2**(sigma_low_scale_factor))
+
+
+if __name__ == '__main__':
```
janden, line 58: Remove.

janden (changes requested):

kymatio/scattering1d/filter_bank.py
pawarrick: In fact this breaks because T is in the function signature and log2_T is not defined here

pawarrick:
just one update here moving the log2_t assignment. sorry for the noise.

kymatio/scattering1d/core/scattering1d.py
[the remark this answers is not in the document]
eickenberg: so if T can be non-dyadic, why specify it as its log?

kymatio/scattering1d/core/scattering1d.py
eickenberg: > I think asserts make sense here for readability

`raise` adds even more readability

kymatio/scattering1d/filter_bank.py
eickenberg: If the sentence makes sense it can stay. It's just that I don't fully follow it.
The fourier-signal reciprocity is also not helping with interpreting the words min and max

tests/scattering1d/test_keras_scattering1d.py
```diff
@@ -0,0 +1,60 @@
...
+from tensorflow.keras.layers import Input, Flatten, Dense
+from kymatio.keras import Scattering1D
+from tensorflow.keras.models import Model
+import os
+import numpy as np
+import io
+
+backends = []
+
+class TestScattering1DKeras:
```
eickenberg, line 11: class necessary?

tests/scattering1d/test_correctness.py
OverLordGoldDragon: Why? This is for local debugging.

kymatio/scattering1d/frontend/base_frontend.py
OverLordGoldDragon: Existing logic fails with padding by more than `N`.

OverLordGoldDragon: > Looks like this is resolved?

Old version is highlighted, see [latest](https://github.com/kymatio/kymatio/blob/1ee37e84ac3c0050440f64910a395fab97a30bf8/kymatio/scattering1d/frontend/base_frontend.py#L58).

kymatio/scattering1d/filter_bank.py
OverLordGoldDragon: @pawarrick I don't follow, tests pass for me. `log2_T` is always defined:

```python
    log2_T = math.floor(math.log2(T))
    if max_subsampling is None:
```

OverLordGoldDragon: @janden Because `0` is wrong?

kymatio/scattering1d/core/scattering1d.py
OverLordGoldDragon: Agreed, but in this or separate PR?

kymatio/scattering1d/filter_bank.py
OverLordGoldDragon: @janden @eickenberg This commit doesn't change the relevance of that line.

In CQT, if we limit `sigma_min`, we also limit the lowest center frequency, so we will not tile the entire frequency axis. If docs make no mention, it can mislead that "lower `J`" -> "less coverage".

That's why the line was there in the first place, was it not?

kymatio/scattering1d/core/scattering1d.py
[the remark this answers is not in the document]
OverLordGoldDragon: Because subsampling, and sampling filters, is only done on dyadic scales. The "non-dyadic" part is for setting `sigma_low`, i.e. the amount of imposed time-shift invariance does not have to be in powers of 2.

It also sets up for `log2_F` in JTFS which is reused a _lot_ more and serves more diverse logic.

kymatio/scattering1d/core/scattering1d.py
OverLordGoldDragon: This doesn't scale well. It's more complicated in terms of expanding the definition into its full expression, namely substituting for `kJ`. It's also not `kJ` but `klog_2T`. 

It's fair for Scattering1D, but JTFS uses [more `_adj` logic](https://github.com/OverLordGoldDragon/kymatio/blob/8449dcc6aa620cddfa7ba24e6c3936343bb1d3a7/kymatio/scattering1d/core/timefrequency_scattering.py#L416) that gets out of hand if rewritten this way - not just in more raw code, but in how we think of the logic. Note there isn't even a `k` there, because subsampling is convoluted per the "trimming vs subsampling" distinction; a `kJ` equivalent would be a "higher level" term which obfuscates intent. I would have a much harder time with it.

Then doing `sub1_adj` in time scattering makes it consistent and reading JTFS more familiar.

OverLordGoldDragon: Bottom line, I can do your version for time scattering, but am against reformulating JTFS the same way.

OverLordGoldDragon:
> That a large `T` would require a large amount of padding that we don't currently support? 

Yes

> How would this be different from `2 ** J` being large?

It's not. Currently we suffer _a lot_ boundary effects with `2 ** J == N`.


> I don't know about this. Adding more binary blobs to the repo should only be done when absolutely necessary. 

It's necessary in JTFS, and there's [7 of them](https://github.com/lostanlen/kymatio/tree/fix-672/tests/scattering1d). Then maybe a better storage format? Something with top-level keys mapping to separate tests.

OverLordGoldDragon:
Alternatively use delimiter syntax like

```
0:J
0:x
...
1:J
1:x
...
```

where `0` and `1` are test numbers, explained under test's docstring (or elsewhere). Then

```python
data = np.load(...)
test_nums = np.unique([k.split(':')[0] for k in data.files])
for test_num in test_nums:
    J = data[test_num + ':J']
```
Can instead do `test_name`, e.g. `'standard', 'small_T'

kymatio/scattering1d/filter_bank.py
OverLordGoldDragon: @pawarrick Nevermind you moved it - good catch.

tests/scattering1d/test_correctness.py
janden: Then run `pytest tests/scattering1d/test_correctness.py`.

kymatio/scattering1d/frontend/base_frontend.py
janden: > Old version is highlighted, see [latest](https://github.com/kymatio/kymatio/blob/1ee37e84ac3c0050440f64910a395fab97a30bf8/kymatio/scattering1d/frontend/base_frontend.py#L58).

Ok. Is that a yes or no?

janden:
> It's not. Currently we suffer _a lot_ boundary effects with `2 ** J == N`.

Ok then that's a separate issue. We can't solve all the problems of Kymatio in one PR. Please make an issue and we can address it separately.

> It's necessary in JTFS, and there's [7 of them](https://github.com/lostanlen/kymatio/tree/fix-672/tests/scattering1d). Then maybe a better storage format? Something with top-level keys mapping to separate tests.

The issue is not storage format or number, but repo size. I agree that it'll be necessary to add binary blobs for JTFS to test against reference data, but their total size should be minimal. It took a lot of work getting the original blobs down to size and making sure they are correct. Once they're in master, they're there forever.

tests/scattering1d/test_correctness.py
OverLordGoldDragon: That doesn't permit interactive debugging.

janden: Then run `pytest --trace`, `pytest --pdb`, or add those lines in when you're debugging. At any rate, this doesn't belong in the repo.

kymatio/scattering1d/frontend/base_frontend.py
OverLordGoldDragon: I just confirmed they're 100% identical (aside for `log2_T`) - my confusion, will remove.

janden: Ok. Resolving.

pawarrick:
A way around this would be to ensure that the `T = 2 ** J` where `J` is the value used for the reference data, then pick `J - 1` and check even samples.

@janden I attempted a test like this (committed in comments in the section "new adjust T test").  Keeping T=2**6 and adjusting J from 6 to 5 reduces the number of paths but maintains the same number of time samples. Comparing the scattering coefficients of the paths is successful for the order 0 path but not for the comparable order 1 and 2 paths.  I also tried keeping J constant at 6 and reducing T to 2**5, giving twice the number of time samples, but comparing every other samples did not compare either.   

Is this the type of test you had in mind to reuse the blob data? Other thoughts?

examples/1d/plot_filters_T.py
```diff
@@ -0,0 +1,59 @@
...
+import os
+import numpy as np
+import io
+
+backends = []
+
+from kymatio.scattering1d.backend.numpy_backend import backend
+backends.append(backend)
+
+class PlotFiltersT:
```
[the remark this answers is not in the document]
pawarrick: @eickenberg thanks. All your comments committed (whoa... first time since 1989 I've been called out for _too_ much OO :))

tests/scattering1d/test_keras_scattering1d.py
pawarrick: Done, but kept assertion (for now)

tests/scattering1d/test_keras_scattering1d.py
pawarrick: changed to function

kymatio/scattering1d/frontend/base_frontend.py
OverLordGoldDragon: Yes resolved

OverLordGoldDragon:
@janden @pawarrick `T` controls imposed time invariance, there won't be strided equality. It's why I found the PR's name confusing.

The 7 files in JTFS I'd call 'minimal'. There are implementation aspects that will pass every test but still change output values slightly - and it's very important to know if a code edit does this. For example, once we implement padding by more than `N` (and if set as default), this'll only break the output test. Changed lowpass width has other few such interactions in `Scattering1D`.

tests/scattering1d/test_correctness.py
OverLordGoldDragon: Alright. Will remove last before merge, I might still use it till then.

tests/scattering1d/test_keras_scattering1d.py
janden: Ok. Assert is good and should stay.

janden:
> Is this the type of test you had in mind to reuse the blob data? Other thoughts?

Sort of. You'd have to match the subset of paths that exist in both and compare those. Ideally they should line up.

> The 7 files in JTFS I'd call 'minimal'. There are implementation aspects that will pass every test but still change output values slightly - and it's very important to know if a code edit does this. For example, once we implement padding by more than `N` (and if set as default), this'll only break the output test. Changed lowpass width has other few such interactions in `Scattering1D`.

We'll have to take that discussion when we get there. All I'm saying is that the goal is to include as little binary data as necessary.

kymatio/scattering1d/filter_bank.py
janden: This sentence doesn't make sense logically. Before it was “This ensures A, while satisfying B.” Now it's “A specifies X, while satisfying B.” The first is a constrained guarantee, while the second is… I don't know?

pawarrick:
> @janden @pawarrick `T` controls imposed time invariance, there won't be strided equality. It's why I found the PR's name confusing.

@OverLordGoldDragon You're right that the purpose of this PR is to introduce `T` to control the degree of time invariance.  The initial name "stride" was not the most descriptive but after some discussion we stuck with it for better or for worse.

pawarrick:
> > Is this the type of test you had in mind to reuse the blob data? Other thoughts?
> 
> Sort of. You'd have to match the subset of paths that exist in both and compare those. Ideally they should line up.

@janden, that's what the commented code in the test intends to do [.](https://github.com/pawarrick/kymatio/blob/stride/tests/scattering1d/test_numpy_scattering1d.py#L65). 
But the paths that exist in both do not line up for some reason.  I will look again to see if I can see why.

OverLordGoldDragon:
> You'd have to match the subset of paths that exist in both and compare those. Ideally they should line up.

No paths in first or second order will be equal. It's what I meant: lesser `T` -> narrower `phi` -> changed values with `average=True`.

kymatio/scattering1d/filter_bank.py
OverLordGoldDragon: Agreed it's a bit awkward, but I think the idea's conveyed. The alternative is `sigma_min limits the smallest frequential width`, which is actually more accurate, `min(sigma1s) == sigma_min` isn't guaranteed.

Also I didn't get pinged for this, Github's notifs aren't reliable - @janden to be sure.

OverLordGoldDragon: Clarifying, `sigma_min` is guaranteed with `Q > 1` via "intermediate" wavelets.

janden:
Rebased using
```
git rebase --onto dev old-dev
```

pawarrick:
> > You'd have to match the subset of paths that exist in both and compare those. Ideally they should line up.
> 
> No paths in first or second order will be equal. It's what I meant: lesser `T` -> narrower `phi` -> changed values with `average=True`.

The test keeps `T` constant but changes `J` so should be same filter.   They do compare actually. Found my error in selecting comparable order2 paths (they're not necessarily consecutive).

OverLordGoldDragon:
@pawarrick I only see `phi_f['sigma'] == 0.1/T` tested as you describe, there's no comparing scattering paths - only shapes are tested in the scattering test . Also, changing only `J` just tests that `T` is set independent of it - it's something, but doesn't test the effects on actual computation.

pawarrick:
> @pawarrick I only see `phi_f['sigma'] == 0.1/T` tested as you describe, there's no comparing scattering paths - only shapes are tested in the scattering test . Also, changing only `J` just tests that `T` is set independent of it - it's something, but doesn't test the effects on actual computation.

We're talking about the test test_numpy_scattering_1d [here](https://github.com/pawarrick/kymatio/blob/stride/tests/scattering1d/test_numpy_scattering1d.py#L65)

OverLordGoldDragon:
I see, it'll uncomment. Yes this can pass, but a test for changed `T` is preferable.

pawarrick:
fixed test_numpy_scattering1d.py adjust J test to check comparable paths only

pawarrick:
@janden I've done all the changes on my list.  Anything else pending for this PR?

pawarrick:
@OverLordGoldDragon `test.test_torch_scattering1d.test_T` is asserting at line 430:  `assert np.allclose(Sg0, Sx0)`

OverLordGoldDragon:
@pawarrick That's... not good - I'll take a look

OverLordGoldDragon:
@pawarrick I'm trying to switch to this branch but it's showing a ton of merge conflicts - are you actually able to run the tests?

pawarrick:
@OverLordGoldDragon I needed a fresh clone of the repo because of merge conflicts as well.  Maybe due to @janden's rebase above?

> Rebased using
> 
> ```
> git rebase --onto dev old-dev
> ```

OverLordGoldDragon:
I don't get how a branch existed in a conflicted state; resolved everything in `dev`'s favor. I also recall running this branch after @janden's rebase... unsure what it's all about. Commit history is a mess now, don't know how to revert

OverLordGoldDragon:
@pawarrick The assertion fails (and that's a good thing).

A subset of paths can be indexed, but I still advocate for a new binary entry.

OverLordGoldDragon:
(Clarifying, subset of paths in the original idea of increasing `J` while keeping `T` same.)

pawarrick:
let's update approval status

lostanlen:
(1) is this the first PR we can merge among the list of open PRs?
(2) how can we do that?

eickenberg:
I think we need to push out the release first.
I believe it was ready but there was a discussion necessary about some
inheritance stuff iirc. There should be a comment by @janden somewhere
about that

pawarrick:
> 

Hi @eickenberg and @lostanlen, getting this PR finalized was one of my new year's tasks after getting the PMEA article submitted... Which release are we talking about?  In the meantime I want to check that all my the stride tests that I wrote still work the way I intended...

pawarrick:
I updated to ensure that my tests work as intended.  This is ready for merging as far as I can tell.

lostanlen:
@eickenberg shall we merge this to dev?

lostanlen:
@pawarrick i'm realizing that @janden requested some debug mechanisms and commented code to be removed. could you do that? then i'll resolve his comments and make a build

pawarrick:
I think this was done ages ago, but I'll check again

lostanlen:
ah, you're right, i'm looking at an outdated review. it was never resolved though, and i'm nervous at the thought of dismissing it

could you please double-check? thanks!

examples/1d/plot_filters_T.py
pawarrick: plot_filters_T.py was incorporated into plot_filters.py

examples/1d/plot_filters_T.py
```diff
@@ -0,0 +1,59 @@
...
+from kymatio import Scattering1D
+from kymatio.scattering1d.filter_bank import scattering_filter_factory
+import os
+import numpy as np
+import io
+
+backends = []
+
+from kymatio.scattering1d.backend.numpy_backend import backend
+backends.append(backend)
```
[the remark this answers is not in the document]
pawarrick: was done

examples/1d/plot_filters_T.py
```diff
@@ -0,0 +1,59 @@
...
+          for psi_f in psi1_f:
+              plt.plot(np.arange(N)/N, psi_f[0], 'b')
+          plt.xlim(0, 0.5)
+          plt.xlabel(r'$\omega$', fontsize=18)
+          plt.ylabel(r'$\hat\psi_j(\omega)$', fontsize=18)
+          
+          plt.legend()
+          plt.savefig(plot_dir + '/order1filters_T' + str(T) + 'J' + str(J) +  '.pdf', bbox_inches='tight', orientation='landscape')
+    
+    def Scattering1D_filter_factory_T(self, backend):
```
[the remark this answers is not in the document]
pawarrick: was resolved

examples/1d/plot_filters_T.py
```diff
@@ -0,0 +1,59 @@
...
+          plt.figure()
+          plt.plot(np.arange(N)/N, phi_f[0], 'r', label=msg)
+          for psi_f in psi1_f:
+              plt.plot(np.arange(N)/N, psi_f[0], 'b')
+          plt.xlim(0, 0.5)
+          plt.xlabel(r'$\omega$', fontsize=18)
+          plt.ylabel(r'$\hat\psi_j(\omega)$', fontsize=18)
+          
+          plt.legend()
+          plt.savefig(plot_dir + '/order1filters_T' + str(T) + 'J' + str(J) +  '.pdf', bbox_inches='tight', orientation='landscape')
```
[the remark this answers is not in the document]
pawarrick: was resolved to only show, not save

examples/1d/plot_filters_T.py
```diff
@@ -0,0 +1,59 @@
...
+        for j in Js:
+            J = j
+            for i in sigma_low_scale_factor:
+                T=2**(J-i)
+                if i == 0:
+                    default_str = ' (default)'
+                else: 
+                    default_str = ''
+                phi_f, psi1_f, psi2_f, _ = scattering_filter_factory(np.log2(N), J, Q, T)
+                msg = 'J=' + str(J) + ', T=' +str(T) + default_str + ': LP-filter width $\sigma_{low}$=' + str(phi_f['sigma']) + ' (=0.1/T)'
```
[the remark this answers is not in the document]
pawarrick: will resolve this style issue after merge

tests/scattering1d/test_correctness.py
pawarrick: this was resolved

tests/scattering1d/test_keras_scattering1d.py
pawarrick: resolved

tests/scattering1d/test_keras_scattering1d.py
pawarrick: resolved

pawarrick:
I have gone over this PR history and marked issues as "resolved" where I am sure.  There are a few issues mentioned in thread with @OverLordGoldDragon issues that I am not sure about.

lostanlen:
@pawarrick we have some serious failing tests

pawarrick:
This is what I was referring to earlier about the build being blocked.  
I'll look at the 1D problems first, I haven't touched any 2D code but there may be some dependencies in common.

pawarrick:
I will look over these test failures later today and then we can assess how ready the branch is to merge.

lostanlen:
```
return tf.cast(tf.pad(x, paddings, mode="REFLECT"), tf.complex64)
```

?

lostanlen:
https://github.com/pawarrick/kymatio/commit/048c23ed9a6af99f0b6a4017d2e748dbf3c53a3a

maybe no longer necessary?

pawarrick:
1D and 2D tests in folder test all pass on my machine.  Now the build is aborting for some other reason related to a timeout.

pawarrick:
pip and conda CIs build successfully, self-hosted CIs are queued?

pawarrick:
@lostanlen, do you know what is causing this blocked queue?

rastegah:
> @lostanlen, do you know what is causing this blocked queue?

@lostanlen mentioned that the Travis CI has not been triggered for some time ... the server that hosts it it may be down? it's managed by @eickenberg ?

tests/scattering1d/test_correctness.py
OverLordGoldDragon: @pawarrick Not exactly

pawarrick: @OverLordGoldDragon do you mean the point about interactive debugging?  I can run pytest in my IDE fine without main scaffolding.  In any event, this should not get in the way of the merge in my opinion.

OverLordGoldDragon (approved):
Assuming all tests pass, this looks good to merge. If we missed anything, it's not hard to fix later.

Currently the following are unresolved. However,

 - [1](https://github.com/kymatio/kymatio/pull/673#discussion_r640979502) can revisit in a style session
 - [2](https://github.com/kymatio/kymatio/pull/673#issuecomment-849951584) I can pass for now. `T` is indirectly tested by `test_correctness`, but I still advocate for a dedicated test
 - [3](https://github.com/kymatio/kymatio/pull/673#issuecomment-1017778925) I have the entire PR open, all resolved and unresolved comments visible; Ctrl + F "inher" only points to your comment @eickenberg 
 - [4](https://github.com/kymatio/kymatio/pull/673#discussion_r808423072) since all tests pass, no need at the moment, I can remove the code from `test_correctness.py`.

Worst case, JTFS first order scattering essentially duplicates the time scattering here, and nothing broken will get past `test_jtfs.py`. If we can consider the above resolved for now, then this should merge.

pawarrick:
Hi Michael @eickenberg, do you have any news on the self hosting issue? Thanks.

lostanlen:
@pawarrick @pawarrick i spoke with @eickenberg yesterday and he's OK with disabling Jenkins for PRs merging to the dev branch. 

So let's merge this

OverLordGoldDragon:
Removed my testing code, now should be merge-ready.

lostanlen:
Michael has disabled the out-of-order Jenkins CI.

@eickenberg is "squash and merge" the correct way?

pawarrick:
Excellent

PR #685, opened 2021-03-14
Support of torch 18/legacy 17

edouardoyallon:


eickenberg (approved):
Code looks the same to me as the previous PR, yes?

Apart from the change to the github actions and the modifications of the tests?

If so, then I think this is good to go as a fix on master.

edouardoyallon:
> Code looks the same to me as the previous PR, yes?
> 
> Apart from the change to the github actions and the modifications of the tests?
> 
> If so, then I think this is good to go as a fix on master.

Yes, it should be close from before(there is stilla bug and I'm adding a bit of doc)

lostanlen:
voilà !! thanks for the impressive work @edouardoyallon

PR #686, opened 2021-03-15
TST Restore torch<1.3 for jenkins

janden:


edouardoyallon (approved):
Fair enough but why?

eickenberg:
As long as it doesn't become a huge annoyance, it seems fine to have jenkins running on the versions that it can handle

janden:
> Fair enough but why?

Because hopefully soon we'll have new GPUs on it and it will run.

PR #694, opened 2021-03-19
TST So long, Circle CI

janden:


edouardoyallon (approved):
so byebye circle CI

janden:
…it's time that we began
to laugh and cry and cry and laugh about it all again…

PR #696, opened 2021-03-19
TST Add py35 to GHA pip workflow

janden:


janden:
It's fine, we just have to run torch==1.6 for this one. I don't know how to do it in GHA but will figure it out.

Since 0.2.0 was released with py35 compatibility, we need to make sure that 0.2.1 does as well, then we can drop it for 0.3.

edouardoyallon:
oh, I see the point.

edouardoyallon (approved):

PR #699, opened 2021-03-20
TST Set max-parallel to one for self-hosted

janden:
These tests always seem to run out of memory, so it's probably safer to
run them one at a time to reduce flakiness.


janden:
Still fails… Would be better if we could run these tests on a machine that is less busy. It's time-consuming to have to deal with all these sporadic test failures.

janden:
Will be dealt with in #701 instead.

PR #702, opened 2021-03-24
`scipy.fftpack` -> `scipy.fft`

OverLordGoldDragon:
`.fft` is more performant per my benchmarks, and scipy [recommends against](https://docs.scipy.org/doc/scipy/reference/fftpack.html) `.fftpack`. As used throughout kymatio, it's just a drop-in replacement.


lostanlen:
Thanks @OverLordGoldDragon . Might be good to run this through `asv`?
Tagging @edouardoyallon

janden:
IIRC when we first wrote this, `scipy.fft` wasn't supported across all versions yes. Makes sense to drop this in now. It would be better to pull this into dev, though, since the backend infrastructure has changed quite a bit.

OverLordGoldDragon:
@lostanlen I'm unfamiliar with `asv` but I benched it a fair bit on my 8-core i7 with all FFTs.

> better to pull this into dev

Works for me, but I'm rather confused on how to navigate the progress of this repository (`dev` being far behind `master`, for one); any ETA for its merger?

janden:
> but I'm rather confused on how to navigate the progress of this repository 

Long story short. dev has had some more long-term restructuring to make the backend more flexible and avoid redundancy. In parallel, we had to bring some changes into master to ensure compatibility with the latest torch (where the FFT interface changed) and fix a bug in Keras. The plan is to cut 0.2.1 from master in the next few days.
Unfortunately, since the changes in master and dev both deal with the backend, the rebase is going to be quite hairy, so the plan is to close out as many of the PRs into dev as possible, *then* rebase onto master (and eventually merge in when 0.3 is ready). This way, there will only be one big rebase. In terms of functionality, however, the difference is not huge (as @lostanlen points out).
Hope this clears things up a bit.

OverLordGoldDragon:
@janden Thanks. Will re-open into `dev`.

lostanlen:
Awesome. Many thanks for this @OverLordGoldDragon 🚀

PR #703, opened 2021-03-24
`scipy.fftpack` -> `scipy.fft`

OverLordGoldDragon:
[Reopened](https://github.com/kymatio/kymatio/pull/702). -- `.fft` is more performant per my benchmarks, and scipy [recommends against](https://docs.scipy.org/doc/scipy/reference/fftpack.html) `.fftpack`. As used throughout kymatio, it's just a drop-in replacement.

### DIFF
--- a/examples/2d/plot_filters.py
+++ b/examples/2d/plot_filters.py
@@ -8,7 +8,7 @@ from colorsys import hls_to_rgb
 import matplotlib.pyplot as plt
 import numpy as np
 from kymatio.scattering2d.filter_bank import filter_bank
-from scipy.fftpack import fft2
+from scipy.fft import fft2
 
 
 ###############################################################################
--- a/kymatio/backend/numpy_backend.py
+++ b/kymatio/backend/numpy_backend.py
@@ -1,10 +1,10 @@
 import numpy
-import scipy.fftpack
+import scipy.fft
 
 
 class NumpyBackend:
     _np = numpy
-    _fft = scipy.fftpack
+    _fft = scipy.fft
 
     name = 'numpy'
 
--- a/kymatio/scattering1d/filter_bank.py
+++ b/kymatio/scattering1d/filter_bank.py
@@ -1,7 +1,7 @@
 import numpy as np
 import math
 import warnings
-from scipy.fftpack import ifft
+from scipy.fft import ifft
 
 def adaptive_choice_P(sigma, eps=1e-7):
     """
--- a/kymatio/scattering2d/filter_bank.py
+++ b/kymatio/scattering2d/filter_bank.py
@@ -1,5 +1,5 @@
 import numpy as np
-from scipy.fftpack import fft2, ifft2
+from scipy.fft import fft2, ifft2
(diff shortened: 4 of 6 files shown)


eickenberg (approved):
Thanks, this looks great.
Curious: How much difference do you see in your benchmarks?
If you have pyfftw installed, could you also compare to that?

eickenberg:
Some tests failing for one particular test run. Anybody have an idea why this is?

OverLordGoldDragon:
`scipy.fft` also supports `workers` kwarg that specifies number of CPU threads to use, a great speedup. If kymatio rather not multithread by default, it could be a kwarg, or - in `ssqueezepy`, I allow on-fly switching with `os.environ['SSQ_PARALLEL'] = '1'`, [here](https://github.com/OverLordGoldDragon/ssqueezepy/blob/dtype/ssqueezepy/utils/fft_utils.py#L136).

@eickenberg I don't recall exactly, but it was definitely substantial - further, `scipy.fftpack.rfft`, which @lostanlen will need, returns a [_real array_](https://docs.scipy.org/doc/scipy-1.5.4/reference/generated/scipy.fftpack.rfft.html) all out of order, requiring post-processing steps, whereas `fft.rfft` returns proper complex.

pyFFTW always beats `scipy.fft` with enough planning, sometimes significantly (50%); just don't do `FFTW_EXHAUSTIVE` or it may run for hours and not do much better than `_PATIENT` with few minutes... (was disappointed). Though it's a bit [clumsy](https://github.com/OverLordGoldDragon/ssqueezepy/blob/dtype/ssqueezepy/utils/fft_utils.py#L111) to manage for both developers and users for best results, and without full minutes of planning I found it to be just as fast or even slower than `scipy.fft`.

janden:
> Some tests failing for one particular test run. Anybody have an idea why this is?

It's likely due to torch 1.8 incompatibility. That's not been brought into dev yet.

I think we should support multithreading by default (pretty sure this is what you get with torch/TF if you run on CPU). Regarding FFTW, the plan was to add that as a potential backend for NP at some point, IIRC.

OverLordGoldDragon:
> Regarding FFTW

Frankly I don't think it's worth the hassle and would prioritize other things; max-threaded `scipy.fft` gets close anyway, and both get thwarted by GPU torch. 

Since kymatio takes FFT in for-loops, there's no risk of 100% CPU spikes, so multithreaded default seems sensible; just add like `scipy.fft.fft(x, threads=multiprocessing.cpu_count())`.

edouardoyallon:
> > Some tests failing for one particular test run. Anybody have an idea why this is?
> 
> It's likely due to torch 1.8 incompatibility. That's not been brought into dev yet.

 didn't read the logs but don't forget they are run on shared machines... slurm comin' soon

edouardoyallon (approved):

eickenberg:
Also `asv` hopefully coming soon to automatically benchmark these kinds of things, but I think here the change is definitely not going harm performance.

PR #706, opened 2021-03-28
DEMO slurm

edouardoyallon:
TODO:
- [x] running one unit-test on slurm
- [x] if the job is cancel by gha, use scancel (more extensive tests to do, i ain't sure of the behavior yet..)
- [x] display the output (err and out) of a given job
- [x] fail if the job has failed during its execution


edouardoyallon:
I'm not sure how to solve it but an issue is that anconda takes about 32G

OverLordGoldDragon:
This is to monitor resource usage, if I understand correctly?

janden:
What's the status on this?

edouardoyallon:
Tests seem good, should this be closed?

PR #714, opened 2021-04-07
Style

OverLordGoldDragon:
As bonus I added paths and `[skip ci]` to skip redundant builds.

### DIFF
--- a/.github/workflows/conda.yml
+++ b/.github/workflows/conda.yml
@@ -1,11 +1,28 @@
 name: Kymatio Conda CI
 
 on:
-    - push
-    - pull_request
+    push:
+        paths:
+          - "kymatio/*"
+          - "tests/*"
+          - ".github/workflows/*"
+          - ".coveragerc"
+          - "requirements.txt"
+          - "requirements_optional.txt"
+          
+    pull_request:
+        paths:
+          - "kymatio/*"
+          - "tests/*"
+          - ".github/workflows/*"
+          - ".coveragerc"
+          - "requirements.txt"
+          - "requirements_optional.txt"
 
 jobs:
     build-conda:
+        if: "! contains(toJSON(github.event.commits.*.message), '[skip ci]')"
+
         runs-on: ubuntu-20.04
 
         strategy:
@@ -52,4 +69,7 @@ jobs:
             - name: Set up Kymatio
               run: conda run -n test-env python3 setup.py develop
             - name: Test
-              run: conda run -n test-env pytest --cov=kymatio
+              run: >
+                  conda run -n test-env pycodestyle --max-line-length=85 kymatio
+                  
+                  conda run -n test-env pytest --cov=kymatio
--- a/examples/2d/plot_filters.py
+++ b/examples/2d/plot_filters.py
@@ -8,7 +8,7 @@ from colorsys import hls_to_rgb
 import matplotlib.pyplot as plt
 import numpy as np
 from kymatio.scattering2d.filter_bank import filter_bank
-from scipy.fftpack import fft2
+from scipy.fft import fft2
 
 
 ###############################################################################
--- a/kymatio/backend/numpy_backend.py
+++ b/kymatio/backend/numpy_backend.py
@@ -1,10 +1,10 @@
 import numpy
-import scipy.fftpack
+import scipy.fft
 
 
 class NumpyBackend:
     _np = numpy
-    _fft = scipy.fftpack
+    _fft = scipy.fft
 
     name = 'numpy'
 
--- a/kymatio/scattering1d/filter_bank.py
+++ b/kymatio/scattering1d/filter_bank.py
@@ -1,7 +1,7 @@
 import numpy as np
 import math
 import warnings
-from scipy.fftpack import ifft
+from scipy.fft import ifft
 
 def adaptive_choice_P(sigma, eps=1e-7):
     """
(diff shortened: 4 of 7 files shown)


OverLordGoldDragon:
Unsure why it lumped together with `fft` commit

.github/workflows/conda.yml
```diff
@@ -1,11 +1,28 @@
...
+          - "kymatio/*"
+          - "tests/*"
+          - ".github/workflows/*"
+          - ".coveragerc"
+          - "requirements.txt"
+          - "requirements_optional.txt"
 
 jobs:
     build-conda:
+        if: "! contains(toJSON(github.event.commits.*.message), '[skip ci]')"
```
lostanlen, line 24: nice

lostanlen (approved):
LGTM. @janden thoughts?

OverLordGoldDragon:
hmm, it didn't build, think it skips the initial commit - will add some dummies; should make sure we make it to `pycodestyle`

janden:
> LGTM. @janden thoughts?

What's the argument against black again? Ideally, we'd want something that can be easily auto-formatted to reduce friction with new PRs.

OverLordGoldDragon:
> auto-formatted

[autopep8](https://github.com/hhatto/autopep8)

> argument against black

It does not allow making any exceptions to rules (non-issue if you agree with all of them, but I personally don't)

OverLordGoldDragon:
Maybe one of `pip` or `conda` shouldn't do `pycodestyle` so we know it's only style error

OverLordGoldDragon:
Can configure [autopep8 Action](https://github.com/marketplace/actions/autopep8) to make a PR if it finds anything to "correct", or commit directly. I'd run `pycodestyle` locally first to see the kinds of things it complains about and whether some are worth exempting.

edouardoyallon:
Do you think it is possible to separate in several commits the PR you propose? the reason is that if we revert a commit, it'll revert several features. Also it's difficult to code reiview codes which interact with each other.

OverLordGoldDragon:
@edouardoyallon Yeah; one should branch off of `upstream/dev` rather than their own `dev`. This one's extras is just `scipy.fft`, which is already merged (though Github doesn't recognize it).

janden:
> It does not allow making any exceptions to rules (non-issue if you agree with all of them, but I personally don't)

Well that's the selling point, no? What rules in particular do you find problematic?

OverLordGoldDragon:
Mainly vertical align and spacing

```python
short = function(var1, variable2, the_variable3)
longer_name = function(var1, var2, the_variable3)
```
vs
```python
short       = function(var1, variable2, the_variable3)
longer_name = function(var1, var2,      the_variable3)
```

<hr>

```python
out = a * b ** c + d / e - f ** g + h * m ** (1 / n)
```
vs
```python
out = a * b**c + d/e - f**g + h * m**(1/n)
```

<hr>

```python
psihs_hf = [wav(scale=scale)[:N//2 + 1] for wav in wavs]
psis = [wav.psifn(scale=scale) for wav in wavs]
```
vs
```python
psihs_hf = [wav(scale=scale)[:N//2 + 1] for wav in wavs]
psis     = [wav.psifn(scale=scale)      for wav in wavs]
```

<hr>

```python
k1 = max(j1 - oversampling, 0)  # what we subsampled in 1st-order
k2 = max(j2 - k1 - oversampling, 0)  # what we subsample now in 2nd
```
vs
```python
k1 = max(j1 - oversampling, 0)       # what we subsampled in 1st-order
k2 = max(j2 - k1 - oversampling, 0)  # what we subsample now in 2nd
```

<hr>

**Won't allow module-level imports unless on top of file**; example [ssqueezepy/visuals.py](https://github.com/OverLordGoldDragon/ssqueezepy/blob/master/ssqueezepy/visuals.py#L943), I'd have to duplicate lots to get around this, or significantly change codebase (and make it worse overall).

OverLordGoldDragon:
It looks worse in practice; Python's rules are other languages' errors, and vice versa, so we're not quite breaking laws of nature. There's still plenty to enforce outside these.

janden:
So which one of the examples above are black and which are not? Personally, I'd prefer the former. As for module-level imports, there's a good reason to keep them in one place, no?

OverLordGoldDragon:
Former is Black. In example I linked, putting them on top would error per recursive imports (`a` imports `b` but `b` imports `a`). This can be remedied, but net-cost is against simplicity and readability; there's a reason PyTorch, TensorFlow, and other mature libraries don't do linting - they become more an impedance than they help. Currently this doesn't afflict Kymatio, but it might eventually.

Maybe let's get others' take on this also, styling is subjective; which of my code blobs do you prefer, @eickenberg @edouardoyallon @lostanlen ?

MuawizChaudhary:
> Former is Black. In example I linked, putting them on top would error per circular imports (`a` imports `b` but `b` imports `a`). This can be remedied, but net-cost is against simplicity and readability; there's a reason PyTorch, TensorFlow, and other mature libraries don't do linting - they become more an impedance than they help. Currently this doesn't afflict Kymatio, but it might eventually.
> 
> Maybe let's get others' take on this also, styling is subjective; which of my code blobs do you prefer, @eickenberg @edouardoyallon @lostanlen @MuawizChaudhary ?

Not a fan. Prefer the former.

lostanlen:
Would be good to merge this before the hackathon on May 20 :)

PR #717, opened 2021-04-11
Visuals

OverLordGoldDragon:
Mainly PR to allow others to viz coeffs with latest version of `timefrequency_scattering.py`, but I write it to reasonably allow `dev` merger.

Added visuals primitives `imshow` and `plot` that greatly reduce duplication; users don't need to use these, currently neither documented nor organized., but devs are welcome to fix.

### DIFF
--- a/.github/workflows/conda.yml
+++ b/.github/workflows/conda.yml
@@ -1,28 +1,11 @@
 name: Kymatio Conda CI
 
 on:
-    push:
-        paths:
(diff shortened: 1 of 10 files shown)


OverLordGoldDragon:
Actually I'll keep latest visuals on JTFS branch then drop it before merging; this PR can then be reviewed for visuals merger.

PR #724, opened 2021-04-15
Update jenkins environment with torch 1.7.1

dylex:
This adds a custom build of pytorch 1.7.1 for the older GPUs on the jenkins machine.
I also pinned torchvision 0.8.2 to match the torch version.
Tests seem to be passing.

If you'd prefer we could make this a separate test and run both torch 1.2 and 1.7 separately.


janden (approved):
Thank you for fixing this. Will make things a lot easier.

janden:
@edouardoyallon Is this ok to merge?

PR #726, opened 2021-04-18
[WIP] Allow non-CQT filter failure

OverLordGoldDragon:
Current behavior is to error if `get_normalizing_factor` is too small (i.e. wavelet is actually a pure tone) on _any wavelet_. Since ideally we work with CQT-only anyway, this is unnecessarily strict - and problematic in JTFS where `2**J_support` is typically small (~256, 128), limiting flexibility with `Q_fr`.

PR allows filterbanks whose _intermediate_, or non-CQT, filters fail: `for q in range(1, num_intermediate + 1)`.

### Todo
 1. `meta` will no longer map to output coefficients in all cases, nor will an accurate mapping be possible without explicitly checking for non-CQT failure (by computing them). Options:
      - A. Compute said filters
      - B. Get info from already-computed filters (makes method no longer standalone)
 2. Docs

### DIFF
--- a/kymatio/scattering1d/filter_bank.py
+++ b/kymatio/scattering1d/filter_bank.py
@@ -461,6 +461,7 @@ def compute_params_filterbank(sigma_low, Q, r_psi=math.sqrt(0.5), alpha=5.):
     xi = []
     sigma = []
     j = []
+    is_cqt = []
 
     if sigma_max <= sigma_low:
         # in this exceptional case, we will not go through the loop, so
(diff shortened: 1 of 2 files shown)


janden:
Is this WIP? If so, will convert to draft.

OverLordGoldDragon:
@janden There's a decision to be made on handling `meta` (see Todo 1); I favor A since it's not much compute and keeps the method standalone.

OverLordGoldDragon:
I can no longer reproduce this, grid searched over wide range of `N, Q, J`; seems it was exclusive to `compute_minimum_support_to_pad` which built the filterbank with half the appropriate `N`. Behavior _is_ still reproduced in `compute_minimum_support_to_pad`, but already accounted for in new logic that starts with `N` and doubles until `halfwidth < N/2`. 

Worth looking out for in future changes, but currently serves no purpose while adding complexity so closing.

PR #727, opened 2021-04-18
Introduce `pad_mode`

OverLordGoldDragon:
Allow user to choose between `reflect, symmetric, zero` padding.

While a fair default, `reflect` isn't all-purpose best; unless derivatives at boundaries are smooth, it may do more harm than good (I've seen in testing ridge extraction in `ssqueezepy`). Further, it is uniquely detrimental to JTFS, where lowpass captures opposite spin generated by `reflect` (confirmed with @lostanlen).

Current options are bottlenecked by TensorFlow; it can be configured per-backend. Most flexibly, a user can pass a `lambda`. For now I went with barebones that includes the important case of `zero` padding - most primitive, but also least priors, making it arguably general best. (P.S., answer to my 'quiz' at bottom [here](https://github.com/kymatio/kymatio/discussions/710) is boundary effects).

### DIFF
--- a/kymatio/scattering1d/backend/numpy_backend.py
+++ b/kymatio/scattering1d/backend/numpy_backend.py
@@ -32,7 +32,7 @@ class NumpyBackend1D(NumpyBackend):
         return res
 
     @classmethod
-    def pad(cls, x, pad_left, pad_right, padtype='reflect'):
+    def pad(cls, x, pad_left, pad_right, pad_mode='reflect'):
         """Pad real 1D tensors
         1D implementation of the padding function for real PyTorch tensors.
         Parameters
@@ -46,7 +46,7 @@ class NumpyBackend1D(NumpyBackend):
         pad_right : int
             amount to add on the right of the tensor (at the end of the temporal
             axis).
-        padtype : str
+        pad_mode : str
             name of padding to use.
         Returns
         -------
@@ -55,13 +55,13 @@ class NumpyBackend1D(NumpyBackend):
         """
         if (pad_left >= x.shape[-1]) or (pad_right >= x.shape[-1]):
             raise ValueError('Indefinite padding size (larger than tensor).')
-        if padtype == 'zero':
-            padtype = 'constant'
+        if pad_mode == 'zero':
+            pad_mode = 'constant'
 
         paddings = ((0, 0),) * len(x.shape[:-1])
         paddings += (pad_left, pad_right),
 
-        output = cls._np.pad(x, paddings, mode=padtype)
+        output = cls._np.pad(x, paddings, mode=pad_mode)
 
         return output
 
--- a/kymatio/scattering1d/backend/numpy_backend.py
+++ b/kymatio/scattering1d/backend/numpy_backend.py
@@ -46,6 +46,8 @@ class NumpyBackend1D(NumpyBackend):
         pad_right : int
             amount to add on the right of the tensor (at the end of the temporal
             axis).
+        padtype : str
+            name of padding to use.
         Returns
         -------
         output : tensor
--- a/kymatio/scattering1d/backend/numpy_backend.py
+++ b/kymatio/scattering1d/backend/numpy_backend.py
@@ -32,7 +32,7 @@ class NumpyBackend1D(NumpyBackend):
         return res
 
     @classmethod
-    def pad(cls, x, pad_left, pad_right):
+    def pad(cls, x, pad_left, pad_right, padtype='reflect'):
         """Pad real 1D tensors
         1D implementation of the padding function for real PyTorch tensors.
         Parameters
@@ -53,11 +53,13 @@ class NumpyBackend1D(NumpyBackend):
         """
         if (pad_left >= x.shape[-1]) or (pad_right >= x.shape[-1]):
             raise ValueError('Indefinite padding size (larger than tensor).')
+        if padtype == 'zero':
+            padtype = 'constant'
 
         paddings = ((0, 0),) * len(x.shape[:-1])
         paddings += (pad_left, pad_right),
 
-        output = cls._np.pad(x, paddings, mode='reflect')
+        output = cls._np.pad(x, paddings, mode=padtype)
 
         return output
 
--- a/kymatio/scattering1d/frontend/base_frontend.py
+++ b/kymatio/scattering1d/frontend/base_frontend.py
@@ -11,7 +11,7 @@ compute_meta_scattering, precompute_size_scattering)
 
 class ScatteringBase1D(ScatteringBase):
     def __init__(self, J, shape, Q=1, max_order=2, average=True,
-            oversampling=0, vectorize=True, out_type='array', padtype='reflect',
+            oversampling=0, vectorize=True, out_type='array', pad_mode='reflect',
             backend=None):
         super(ScatteringBase1D, self).__init__()
         self.J = J
@@ -22,7 +22,7 @@ class ScatteringBase1D(ScatteringBase):
         self.oversampling = oversampling
         self.vectorize = vectorize
         self.out_type = out_type
-        self.padtype = padtype
+        self.pad_mode = pad_mode
         self.backend = backend
 
     def build(self):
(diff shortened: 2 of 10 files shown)


lostanlen:
Cool. If you want i can write the unit tests for this?

lostanlen:
Small nitpick but i would have liked `padding_mode` better than `padtype`.

The word `type` has a different connotation: `str`, `int`, `float`, etc.
Besides, NumPy has `mode`, Keras has `mode`, and TensorFlow has `mode`. 

So perhaps call it `mode` in internal `pad` functions and `padding_mode` in the constructor?

changhongw:
> Small nitpick but i would have liked `padding_mode` better than `padtype`.
> 
> The word `type` has a different connotation: `str`, `int`, `float`, etc.
> Besides, NumPy has `mode`, Keras has `mode`, and TensorFlow has `mode`.
> 
> So perhaps call it `mode` in internal `pad` functions and `padding_mode` in the constructor?

Vote for `padding_mode`.

changhongw:
> Allow user to choose between `reflect, symmetric, zero` padding.
> 
> While a fair default, `reflect` isn't all-purpose best; unless derivatives at boundaries are smooth, it may do more harm than good (I've seen in testing ridge extraction in `ssqueezepy`). Further, it is uniquely detrimental to JTFS, where lowpass captures opposite spin generated by `reflect` (confirmed with @lostanlen).
> 
> Current options are bottlenecked by TensorFlow; it can be configured per-backend. Most flexibly, a user can pass a `lambda`. For now I went with barebones that includes the important case of `zero` padding - most primitive, but also least priors, making it arguably general best. (P.S., answer to my 'quiz' at bottom [here](https://github.com/kymatio/kymatio/discussions/710) is boundary effects).
> 
> ### Todo
> * Docs (frontend): unsure where to "put" it, docstr layout is strange
> * Testing: unsure if I missed anything in some frontend

Nice idea in the long run!

OverLordGoldDragon:
> unit tests

Sure

> `padtype`

Agreed; how about `padmode`, or `pad_mode`? `padding_mode` rather lengthy

lostanlen:
`torch.Conv1D` has `padding_mode` so consistency with them would be nice


> `padding_mode` rather lengthy

`p+TAB+ENTER`

OverLordGoldDragon:
> `torch.Conv1D` has `padding_mode`

[librosa](https://librosa.org/doc/latest/generated/librosa.stft.html)

OverLordGoldDragon:
Numpy and TensorFlow use `mode`, TensorFlow's `Conv1D` uses `padding`, librosa uses `pad_mode` - can't match them all. I find `pad_mode` more attractive per brevity, changed to it for now - but I'm fine if we insist on anything else.

> unit tests

Unless you had something in mind, I don't see a need here; "testing" wasn't the best summary of my bullet, rather someone familiar with `1d` should see if I missed anything (but I can do later). With JTFS, both coverage and correctness are accounted since JTFS defaults to `zero`.

lostanlen:
ok nevermind let's do `pad_mode`

I'm under the impression that this branch is not covered by unit tests?

```python
# check pad_mode
if self.pad_mode not in ('reflect', 'symmetric', 'zero'):
    raise ValueError("`pad_mode` must be one of: reflect, symmetric, "
                     "zero (got %s)" % self.pad_mode)
```

OverLordGoldDragon:
@lostanlen You're right.

For such things I'd recommend a dedicated `test_exceptions.py`; Kymatio's coverage currently sits at 87%, these arg checks do indeed add up.

lostanlen:
https://github.com/OverLordGoldDragon/kymatio/pull/4

yes a dedicated test file for exceptions would be nice
but for now i have put it under `test_numpy_scattering1d.py`

janden:
Are we sure that `zero`, `reflect` and `symmetric` all are valid padding modes for the various frontends? It seems to line up for TF and NP, but Torch doesn't seem to have `zero` and `symmetric`. (Also I find `reflect` and `symmetric` to be synonymous, but perhaps that's just me…) We should maybe also specify in the docs what these different padding modes mean.

OverLordGoldDragon:
> `symmetric`

@janden Just noticed torch lacks this... but `zero` is supported through `constant` (handled internally). Will change; currently only needed addon is `zero` anyway.

As for docs, where do I put them? Arg doc layout is strange.

janden:
> @janden Just noticed torch lacks this... but `zero` is supported through `constant` (handled internally). Will change; currently only needed addon is `zero` anyway.

Makes sense. What about `symmetric`?

> As for docs, where do I put them? Arg doc layout is strange.

Try here: https://github.com/kymatio/kymatio/blob/776ad5988728117dba73a9a8c7a6cbef0f88593a/kymatio/scattering1d/frontend/base_frontend.py#L279

OverLordGoldDragon:
@janden Only `zero` and `reflect` will be supported. Alts:

 1. Check per-backend
 2. Allow `lambda`

Idea with `2` is, instead of passing `scattering1d(pad=self.backend.pad)`, we pass `pad=self.pad_fn`, where

```python
self.pad_fn = (pad_mode if isinstance(pad_mode, FunctionType) else
               self.backend.pad)
```
and we document `pad_mode` func to have input signature `pad_fn(x, pad_left, pad_right, pad_mode)`.

Simple to implement, maximum flexibility. Should I proceed?

lostanlen:
i like it

janden:
> Only `zero` and `reflect` will be supported.

Ok got it.

> Allow `lambda`

I'm not sure. If you think it could be done cleanly, we can try.

The issue would be that the scattering1d core would try to pass `pad_mode` as an argument to that lambda function. It might be simpler to just do a partial substitution in the frontend and pass the padding function (with fixed padding mode) to the core code without `pad_mode` as another argument. Does that make sense?

OverLordGoldDragon:
The workaround is nice; if looks good I'll do docs/frontends @janden

OverLordGoldDragon:
And what of _minimum_ `T`? If it's any `T > 0`, subsampling logic needs to account for the case of `supp(phi) > supp(psi[-1])`.

kymatio/scattering1d/backend/numpy_backend.py
```diff
@@ -46,18 +46,22 @@ def pad(cls, x, pad_left, pad_right):
...
+            name of padding to use.
         Returns
         -------
         output : tensor
             The tensor passed along the third dimension.
         """
         if (pad_left >= x.shape[-1]) or (pad_right >= x.shape[-1]):
             raise ValueError('Indefinite padding size (larger than tensor).')
+        if pad_mode == 'zero':
+            pad_mode = 'constant'
```
MuawizChaudhary, line 59: Why dont we just name pad_mode to constant from the start?

kymatio/scattering1d/frontend/base_frontend.py
```diff
@@ -51,6 +54,21 @@ def build(self):
         else:
             raise ValueError("shape must be an integer or a 1-tuple")
 
+        # check `pad_mode`, set `pad_fn`
+        if isinstance(self.pad_mode, FunctionType):
+            def pad_fn(x):
+                return self.pad_mode(x, self.pad_left, self.pad_right)
```
MuawizChaudhary, line 60: Why?

kymatio/scattering1d/frontend/base_frontend.py
```diff
@@ -175,6 +193,13 @@ def output_size(self, detail=False):
             `'array'`, the output is a large array containing the
             concatenation of all scattering coefficients. Defaults to
             `'array'`.
+        pad_mode : str (default 'reflect') / function, optional
+            Name of padding scheme to use, one of (`x = [1, 2, 3]`):
+                - zero:    [0, 0, 1, 2, 3, 0, 0, 0]
```
MuawizChaudhary, line 198: just name it constant?

MuawizChaudhary (commented):
Why dont we rename `padding_mode='zero'` to `'constant'`?

MuawizChaudhary:
> > Only `zero` and `reflect` will be supported.
> 
> Ok got it.
> 
> > Allow `lambda`
> 
> I'm not sure. If you think it could be done cleanly, we can try.
> 
> The issue would be that the scattering1d core would try to pass `pad_mode` as an argument to that lambda function. It might be simpler to just do a partial substitution in the frontend and pass the padding function (with fixed padding mode) to the core code without `pad_mode` as another argument. Does that make sense?

not a fan of this lambda stuff.

MuawizChaudhary:
We'll refactor the padding in a future release.

Issue #723, opened 2021-04-15
TypeError: 'module' object is not callable


marcogallotta:
Getting the following error using v0.2 on Ubuntu 20.10 using PyTorch. I was getting the same kind of error in a different line using the latest dev I built yesterday.

All I'm really doing right now is:

```
self.scattering = Scattering1D(J=8, shape=T, Q=12) 
out = self.scattering.forward(X)
```

where `X` is a 2D tensor.

```Traceback (most recent call last):
  File "./torch_net.py", line 147, in <module>
    output = net(inputs)
  File "/usr/local/lib/python3.8/dist-packages/torch/nn/modules/module.py", line 889, in _call_impl
    result = self.forward(*input, **kwargs)
  File "./torch_net.py", line 26, in forward
    out = self.scattering.forward(X)
  File "/usr/local/lib/python3.8/dist-packages/kymatio/frontend/torch_frontend.py", line 23, in forward
    return self.scattering(x)
  File "/usr/local/lib/python3.8/dist-packages/kymatio/scattering1d/frontend/torch_frontend.py", line 114, in scattering
    S = scattering1d(x, self.backend.pad, self.backend.unpad, self.backend, self.J, self.psi1_f, self.psi2_f, self.phi_f,\
  File "/usr/local/lib/python3.8/dist-packages/kymatio/scattering1d/core/scattering1d.py", line 78, in scattering1d
    U_0_hat = fft(U_0, 'C2C')
  File "/usr/local/lib/python3.8/dist-packages/kymatio/backend/base_backend.py", line 58, in __call__
    return self.fft_forward(x, direction=direction, inverse=inverse)
  File "/usr/local/lib/python3.8/dist-packages/kymatio/backend/base_backend.py", line 53, in fft_forward
    output = self.fft(x)
  File "/usr/local/lib/python3.8/dist-packages/kymatio/scattering1d/backend/torch_backend.py", line 134, in <lambda>
    fft = FFT(lambda x: torch.fft(x, 1, normalized=False),
TypeError: 'module' object is not callable```

**Comment by dminskiy** [comment]
I have the same issue with torch 1.8.1, while it all works fine with torch 1.7.1
torch.fft was depreciated in this version of torch: https://github.com/pytorch/pytorch/issues/42175
Annoyingly, when you try to use new versions of these functions like this: 
```
fft = FFT(lambda x: torch.fft.fft(x, 2, norm='backward'),
[...]

PR #728, opened 2021-04-18
Expose `Q2`

OverLordGoldDragon:
For JTFS use cases with `Q2 > 1`, recommended by @changhongw and @lostanlen

Exposes to `Scattering1D` and all frontend derivatives, since JTFS reuses `Scattering1D`. Updated docstrings everywhere in 1D.

### DIFF
--- a/kymatio/scattering1d/frontend/base_frontend.py
+++ b/kymatio/scattering1d/frontend/base_frontend.py
@@ -10,13 +10,12 @@ compute_meta_scattering, precompute_size_scattering)
 
 
 class ScatteringBase1D(ScatteringBase):
-    def __init__(self, J, shape, Q=1, Q2=1, max_order=2, average=True,
+    def __init__(self, J, shape, Q=1, max_order=2, average=True,
             oversampling=0, vectorize=True, out_type='array', backend=None):
         super(ScatteringBase1D, self).__init__()
         self.J = J
         self.shape = shape
-        self.Q = Q
-        self.Q2 = Q2
+        self.Q = Q if isinstance(Q, tuple) else (Q, 1)
         self.max_order = max_order
         self.average = average
         self.oversampling = oversampling
@@ -72,7 +71,7 @@ class ScatteringBase1D(ScatteringBase):
     def create_filters(self):
         # Create the filters
         self.phi_f, self.psi1_f, self.psi2_f, _ = scattering_filter_factory(
-            self.J_pad, self.J, self.Q, Q2=self.Q2, normalize=self.normalize,
+            self.J_pad, self.J, self.Q, normalize=self.normalize,
             criterion_amplitude=self.criterion_amplitude,
             r_psi=self.r_psi, sigma0=self.sigma0, alpha=self.alpha,
             P_max=self.P_max, eps=self.eps)
@@ -88,8 +87,7 @@ class ScatteringBase1D(ScatteringBase):
         meta : dictionary
             See the documentation for `compute_meta_scattering()`.
         """
-        return compute_meta_scattering(self.J, self.Q, Q2=self.Q2,
-                                       max_order=self.max_order)
+        return compute_meta_scattering(self.J, self.Q, max_order=self.max_order)
 
     def output_size(self, detail=False):
         """Get size of the scattering transform
@@ -110,7 +108,7 @@ class ScatteringBase1D(ScatteringBase):
         """
 
         return precompute_size_scattering(
-            self.J, self.Q, Q2=self.Q2, max_order=self.max_order, detail=detail)
+            self.J, self.Q, max_order=self.max_order, detail=detail)
 
     _doc_shape = 'N'
 
@@ -263,10 +261,13 @@ class ScatteringBase1D(ScatteringBase):
         J : int
             The maximum log-scale of the scattering transform. In other words,
             the maximum scale is given by :math:`2^J`.
-        {param_shape}Q : int >= 1
+        {param_shape}Q : int >= 1 / tuple[int]
             The number of first-order wavelets per octave. Defaults to `1`.
-        Q2 : int >= 1
-            The number of second-order wavelets per octave. Defaults to `1`.
+            If tuple, sets `Q = (Q1, Q2)`, where `Q2` is the number of
+            second-order wavelets per octave (which defaults to `1`).
+                - Q1: For audio signals, a value of `>= 12` is recommended in
+                order to separate partials.
+                - Q2: Recommended `== 1` for most (`Scattering1D`) applications.
         max_order : int, optional
             The maximum order of scattering coefficients to compute. Must be
             either `1` or `2`. Defaults to `2`.
--- a/kymatio/scattering1d/frontend/base_frontend.py
+++ b/kymatio/scattering1d/frontend/base_frontend.py
@@ -264,8 +264,9 @@ class ScatteringBase1D(ScatteringBase):
             The maximum log-scale of the scattering transform. In other words,
             the maximum scale is given by :math:`2^J`.
         {param_shape}Q : int >= 1
-            The number of first-order wavelets per octave (second-order
-            wavelets are fixed to one wavelet per octave). Defaults to `1`.
+            The number of first-order wavelets per octave. Defaults to `1`.
+        Q2 : int >= 1
+            The number of second-order wavelets per octave. Defaults to `1`.
         max_order : int, optional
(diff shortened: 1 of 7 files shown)


lostanlen:
Thanks. i think that the API we had in mind with @Janden and others would be to pass a tuple `Q=(Q1, Q2)` instead of passing `Q1` and `Q2` as separate kwargs
At least this is what we did in ScatNet

But i'm also fine with merging it this way and revisiting the API later

OverLordGoldDragon:
@lostanlen Considered it, but this is clearer to document (and talk about). They're distinct parameters, best made explicit.

OverLordGoldDragon:
I'm mainly for explicitness in `TimeFrequencyScattering`, where we make room to describe various use cases; no need for it in `Scattering1D`. Former could pass to latter as a tuple (though, it creates some internal inconsistency with arg handling).

lostanlen:
it's passig on `pip` and failing on `conda`. related to the infamous `torch.fft module is not callable` issue (#723)

merge?

OverLordGoldDragon:
@lostanlen I reproduce #723 on dev but not master, so dev also has refactoring to do relative to master. Should we be recommending `master` over `v0.2.0`, or `dev`, to users? @edouardoyallon 

I'm fine with merging.

changhongw:
> For JTFS use cases with `Q2 > 1`, recommended by @changhongw and @lostanlen
> 
> Exposes to `Scattering1D` and all frontend derivatives, since JTFS reuses `Scattering1D`. Updated docstrings everywhere in 1D.
> 
> ### Todo
> 1. Expose to `Scattering1D`, or only internally?

This depends on the developing team's long-term plan for Kymatio? @lostanlen Passing Q1 and Q2 as a tuple Q=(Q1, Q2) helps the user be aware of the changeability of Q2 and the scattering path. If in the feature, we would like to explore higher order scattering, i.e. to have a Q3 (filters per octave of the third-order time scattering), it could be naturally Q=(Q1, Q2, Q3)? Similarly, we could also have Q_fr = (Q_fr1, Q_fr2). In such cases, a tuple would be better than many separate variables for the argument of `Scattering1D`? 

For the second-order, either works for me as well. The current way introduces fewer changes to `Scattering1D`.

OverLordGoldDragon:
> if in the future ... higher order

Agreed, then tuple's certainly preferable.

I'm not strongly disposed either way, though any higher order for JTFS seems a long shot, so maybe explicit `Q2` just for JTFS.

lostanlen:
i've never implemented third-order JTFS. i imagine it's very intensive computationally. not a priority for Kymatio IMO.
I would prefer something along the lines of
`TimeFrequencyScattering -> Conv2D -> Scattering2D -> Conv2D` ...

So i'm fine with `TimeFrequencyScattering` having exactly two kwargs: `Q` and `Q2`.

Third-order time scattering could be useful for some non-audio applications, but the code is nowhere near ready for that yet. Maybe we won't ever have it (https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it)
Let's wait till we have a specific use case before worrying about `Q=(Q1,Q2,Q3)`.

In conclusion i agree with @OverLordGoldDragon 's design

waiting on @janden's review and then we can merge

janden:
I don't quite follow the argument for separate kwargs. Seems like the tuple approach would be cleaner here (after all, it's `Q`, not `Q1`). For internals, we could separate, but for API I think a tuple is more appropriate.

OverLordGoldDragon:
@janden Current plan is `Q`-only (`Q1`) or tuple for `Scattering1D` and dedicated for `TimeFrequencyScattering`. I'm for making `Q2` explicit so it's more obvious the option exists, and since it's clearer to document and refer to.

Let me know of the agreement, I'll implement - either will work for me.

janden:
> Current plan is `Q`-only (`Q1`) or tuple for `Scattering1D`

I'm not following exactly. From the code, I'm not seeing a tuple, but two arguments: `Q` and `Q2`. Again, I think it would be simpler to have a tuple argument here.

> dedicated for `TimeFrequencyScattering`

Dedicated `Q2` or dedicated `Q_fr`? I think the latter would make more sense here in terms of compatibility with TS.

OverLordGoldDragon:
> I'm not seeing a tuple

Right, I've yet to change it. Then I'll make `Q` a tuple in `Scattering1D`.

> Dedicated `Q2` or dedicated `Q_fr`? I think the latter would make more sense here in terms of compatibility with TS.

Dedicated both.

So then: tuple for both, or dedicate `Q2` in JTFS?

janden:
I'd say tuple `Q` for both to make things compatible, then another argument `Q_fr` to control the frequency Q in JTFS.

OverLordGoldDragon:
@janden Adjusted.

lostanlen:
i need this feature urgently. considering that the test failure is unrelated (#723), i'd be in favor of merging (to `dev`)

kymatio/scattering1d/frontend/base_frontend.py
```diff
@@ -283,8 +287,9 @@ def output_size(self, detail=False):
             The maximum log-scale of the scattering transform. In other words,
             the maximum scale is given by `2 ** J`.
         {param_shape}Q : int
-            The number of first-order wavelets per octave (second-order
```
janden, line 286: Needs to be fixed to align with docstring above.

OverLordGoldDragon:
@lostanlen Use the JTFS branch?

lostanlen:
@OverLordGoldDragon good idea, thanks

kymatio/scattering1d/frontend/keras_frontend.py
```diff
@@ -7,15 +7,16 @@
 
 
 class ScatteringKeras1D(ScatteringKeras, ScatteringBase1D):
-    def __init__(self, J, Q=1, max_order=2, oversampling=0):
```
janden, line 10: Shouldn't this just be `Q` to allow tuples to pass through? Same with the other frontends.

janden (changes requested):
Seems like a few `Q2`s slipped through.

kymatio/scattering1d/frontend/base_frontend.py
OverLordGoldDragon: Same align as original, no?

kymatio/scattering1d/frontend/keras_frontend.py
OverLordGoldDragon: Missed frontends

OverLordGoldDragon:
JTFS is ahead of most these PRs, I no longer replicate every change. Adjusted this one.

kymatio/scattering1d/frontend/base_frontend.py
janden: This one talks about Q, Q2, while the above docstring talks about tuples.

janden:
> JTFS is ahead of most these PRs, I no longer replicate every change. Adjusted this one.

So are these PRs meant to be merged or just staging for JTFS?

kymatio/scattering1d/frontend/base_frontend.py
OverLordGoldDragon: ah _that_ 'align'

OverLordGoldDragon:
Latter; former involves figuring out a merge order and successively fixing conflicts - redundant effort. I can guarantee the PRs accurately reflect what'll be merged with JTFS. (but if anyone's willing to do it I won't object)

I suggest instead keeping the PRs open and referring to them while reviewing JTFS

janden:
Ok.

First, if a PR is not meant to be reviewed and merged, please mark it as a draft and state as much. Furthermore, if there's a discussion about a particular aspect of JTFS that needs to happen, it's better to take that through issues or the discussion board or discord.

Second, I don't think it's a good idea to merge JTFS in one mega-PR. It currently amounts to 4000 additional lines and it will be unmanageable to review it properly. If merging these smaller PRs is equivalent, that's preferable.

OverLordGoldDragon:
So I misunderstood how concurrent merges work.

Yes, I favor merging these PRs first, and can recommend a merge order. #673 should come first (but maybe after #748).

> issues or the discussion board or discord.

I find it better to keep things all in one place. Else there's thread duplication and need to check many sources for one thing.

There's already plenty in discussions and discord regarding JTFS, the PRs are in "almost ready" state at worst.

MuawizChaudhary (commented):
LGTM

PR #729, opened 2021-04-18
Error upon no second-order coeffs w/ `max_order=2`

OverLordGoldDragon:
Users shouldn't think they're getting any second-order when they aren't. Likewise for joint coefficients in JTFS (which instead breaks internally rather than silently).

### DIFF
--- a/tests/scattering1d/test_numpy_scattering1d.py
+++ b/tests/scattering1d/test_numpy_scattering1d.py
@@ -33,3 +33,8 @@ class TestScattering1DNumpy:
 
         Sx = scattering(x)
         assert np.allclose(Sx, Sx0)
+
+        with pytest.raises(ValueError) as record:
+            sc = Scattering1D(
+                shape=(512,), J=2, max_order=2, backend=backend, frontend='numpy')
+        assert "max_order=1" in record.value.args[0]
--- a/tests/scattering1d/test_numpy_scattering1d.py
+++ b/tests/scattering1d/test_numpy_scattering1d.py
@@ -34,7 +34,6 @@ class TestScattering1DNumpy:
         Sx = scattering(x)
         assert np.allclose(Sx, Sx0)
 
-        with pytest.raises(ValueError) as record:
-            sc = Scattering1D(
-                shape=(512,), J=2, max_order=2, backend=backend, frontend='numpy')
-        assert "max_order=1" in record.value.args[0]
+        sc = Scattering1D(shape=(512,), J=2, max_order=2, backend=backend,
+                          frontend='numpy')
+        assert sc._no_second_order_filters
--- a/kymatio/scattering1d/frontend/base_frontend.py
+++ b/kymatio/scattering1d/frontend/base_frontend.py
@@ -51,6 +51,11 @@ class ScatteringBase1D(ScatteringBase):
         else:
             raise ValueError("shape must be an integer or a 1-tuple")
 
+        # check that we get any second-order coefficients if max_order==2
+        if self.max_order == 2 and np.isnan(self.meta()['n'][-1][1]):
+            raise ValueError("configuration yields no second-order coefficients; "
+                             "try increasing `J`, or set `max_order=1`.")
+
         # Compute the minimum support to pad (ideally)
         min_to_pad = compute_minimum_support_to_pad(
             self.N, self.J, self.Q, r_psi=self.r_psi, sigma0=self.sigma0,
--- a/kymatio/scattering1d/frontend/base_frontend.py
+++ b/kymatio/scattering1d/frontend/base_frontend.py
@@ -55,7 +55,8 @@ class ScatteringBase1D(ScatteringBase):
         meta = ScatteringBase1D.meta(self)
         if self.max_order == 2 and np.isnan(meta['n'][-1][1]):
             raise ValueError("configuration yields no second-order coefficients; "
-                             "try increasing `J`, or set `max_order=1`.")
+                             "try increasing `J`, or (`Scattering1D` only) "
+                             "set `max_order=1`.")
 
         # Compute the minimum support to pad (ideally)
         min_to_pad = compute_minimum_support_to_pad(
--- a/kymatio/scattering1d/frontend/base_frontend.py
+++ b/kymatio/scattering1d/frontend/base_frontend.py
@@ -52,7 +52,8 @@ class ScatteringBase1D(ScatteringBase):
             raise ValueError("shape must be an integer or a 1-tuple")
 
         # check that we get any second-order coefficients if max_order==2
-        if self.max_order == 2 and np.isnan(self.meta()['n'][-1][1]):
+        meta = ScatteringBase1D.meta(self)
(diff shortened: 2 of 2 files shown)


lostanlen:
when does this happen? `max_order=2` and `J` very small?
perhaps add a unit test to trigger this ValueError?

OverLordGoldDragon:
@lostanlen Yes, e.g. `J=4` (which isn't _too_ small for some inputs).

I don't see any exception handling tests in `tests/`, but this is guaranteed to work so long as `compute_meta_scattering()` is accurate.

OverLordGoldDragon:
Updated variant ensures correct `self.meta()` is used (e.g. it's overridden in JTFS and stuff breaks)

OverLordGoldDragon:
`Scattering1D`'s error precedes joint's, no clean way to display JTFS-specific message so I clarify that `max_order=1` is only for `Scattering1D`.

OverLordGoldDragon:
@lostanlen Neat unit test, thanks.

janden:
I'm not sure this would be desirable behavior. It's `max_order=2`, not `order=2`. We allow coefficients up to `max_order` to be computed. The fact that we don't reach the maximum shouldn't be a cause for error.

OverLordGoldDragon:
@janden Fair for time scattering (then I'd throw a warning so users know they're not getting any second-order coeffs), but this implies _no_ joint coefficients for JTFS, think we should be stricter there.

janden:
I think behavior should be the same for TS and JTFS here. If you're working at scales small enough that you don't need a second order (which is rare enough anyhow), you don't want to have the software yelling at you. Should we similarly error if JTFS is called with `max_order=1` since that is also quite useless?

OverLordGoldDragon:
@janden `max_order` shouldn't be an arg in JTFS, it does nothing.

If we neither error nor warn, there will be confusion; `max_order=2` will most often deliver second-order, and when it doesn't, users have no way of knowing unless they've read the docs. This is a "silent failure"; similarly we could not error nor warn about `get_normalizing_constant` and expect users to understand that some filters may be numerically unstable.

janden:
> `max_order` shouldn't be an arg in JTFS, it does nothing.

Why not? Would make it consistent with all the other interfaces.

> users have no way of knowing unless they've read the docs. 

Or check the meta output. Again, if they're working with such small scales, the whole point is that they don't *need* second-order coefficients. It's not a failure, it's expected behavior.

OverLordGoldDragon:
> `max_order`

Then default to `None` and warn if it's anything else (i.e. user passed)?

> it's expected behavior

The idea's to be beginner-friendly, it's not an obvious detail. It also adds much complication throughout all of JTFS codebase (frontents, `timefrequency_scattering`) to account for this case and basically do `Scattering1D` without joint coefficients; that's much work without good reason, I'll let others handle this if erroring is so undesired.

janden:
> Then default to `None` and warn if it's anything else (i.e. user passed)?

No default to `2` just like the other frontends and act accordingly.

> The idea's to be beginner-friendly, it's not an obvious detail.

I'm not following. This is such a corner case that it's unlikely to be relevant in most beginner applications. It's more likely something that will be triggered during more advanced experiments (computing quantities for different values of `Q`) with high `oversampling` ratios, etc. In that case, it's more of an issue to have the user “fix” the desired behavior by slotting in a TS object because JTFS refuses to just give you the first-order coefficients.

As an analogy, consider `np.fft.fft`. What happens if you give it a vector of length one or two? You don't get Fourier coefficients, really, just the sum (and the difference). It's a silly thing to do, I agree, but it's well-defined. Or calling `scipy.signal.convolve` with a Dirac delta function as its second argument. Probably not what most people would expect when they perform a convolution, but still well-defined. Of course the second-order scattering transform is more complicated than that, but my point is that this is just a corner case of the definition.

> It also adds much complication throughout all of JTFS codebase (frontents, `timefrequency_scattering`) to account for this case and basically do `Scattering1D` without joint coefficients

The core doesn't have to. Just call the corresponding `Scattering1D` frontend.

OverLordGoldDragon:
If we sub with `Scattering1D`, code that interacts with JTFS may fail due to lack of JTFS-specific attributes. Then a change of instance docs is also due, since JTFS's have nothing to do with S1D's. All this happening silently is just bad software design; I don't follow what's wrong with just erroring. A user invokes JTFS to get JTFS coefficients, if they want time scattering they can just use `Scattering1D`.

If it's decided then I'll let you handle it once JTFS is finished.

> No default to 2 just like the other frontends and act accordingly.

It'd be like

```python
if max_order is None:
    max_order = 2
elif max_order != 2:
    raise
```
There's then the issue of time vs frequency order. If we're going full consistency, then `max_order_fr` is also due exposing? And if they default to not `None` directly in function signature, should we not check they haven't changed? That's two long kwargs that do nothing, and either add code in handling or do nothing silently. Consistency is nice, but so is non-redundancy.

janden:
Do we have any other inputs on this one? Looks like @OverLordGoldDragon and I are at an impasse here. Ping @eickenberg @edouardoyallon @lostanlen.

OverLordGoldDragon:
@janden I agree to not error nor warn for `Scattering1D`, but stand by erroring for JTFS; to do this cleanly I introduced an internal `_no_second_order_filters` that's written at `build()`.

I'm also against exposing `max_order` or `max_order_fr` in JTFS; args that do nothing add noise to user interface (and mislead unless we throw warning). Internal consistency is maintained by setting and passing them internally, while frontend is kept clean.

I welcome others' input; upon disagreement I can expose `max_order` & `_fr`. @eickenberg @edouardoyallon @lostanlen @MuawizChaudhary

tests/scattering1d/test_numpy_scattering1d.py
```diff
@@ -33,3 +33,7 @@ def test_Scattering1D(self, backend):
 
         Sx = scattering(x)
         assert np.allclose(Sx, Sx0)
+
+        sc = Scattering1D(shape=(512,), J=2, max_order=2, backend=backend,
+                          frontend='numpy')
+        assert sc._no_second_order_filters
```
MuawizChaudhary, line 39: Please incorporate this test for other backends.

MuawizChaudhary:
I agree with @janden that we should expose `max_order`. @OverLordGoldDragon Throw a error if you find it necessary for JTFS. 

I think I understand why Joakim is against throwing a warning for when `max_order=2` does not compute the 2nd order. However, until this discussion, I always thought that the scattering would compute the 2nd order if you specified `max_order=2`. 

@janden let say we extended scattering to take an arbitary `max_order=k`. If we only compute up to a `k'` ordering I think that is information the user should be told about. I get your examples with the fft and convolve, but I think that the complexity of scattering gives some credence to what @OverLordGoldDragon is saying. Additionally, those inputs to the functions you mention seem very simple. Is it also simple with scattering? 

I'm ambivilent ultimately. I think if the software was extended to calculate up to a `max_ordering=k`, we should let users know if it does not calculate up to that, but here it sounds like this is a rare occurance when `max_order=2`.

PR #738, opened 2021-05-04
Agnostic padding

OverLordGoldDragon:
Pad modes implemented in backend-agnostic manner, permitting any pad length.

Currently only torch is tested. Problem is, it [lacks negative stride](https://github.com/pytorch/pytorch/issues/229) and makes a copy (though just once). So it's optimally done via CuPy.

However, padding isn't currently used anywhere except on input (which is very fast relative to transform itself), so there's a workaround: convert to and pad in numpy, revert. This'll support all of numpy's large variety of paddings for any length.

Wasn't sure where to put this so I made an `agnostic_backend.py`. Not deploy-ready, awaiting review.

### DIFF
--- a/kymatio/scattering1d/backend/agnostic_backend.py
+++ b/kymatio/scattering1d/backend/agnostic_backend.py
@@ -1,6 +1,5 @@
 
-def pad(x, pad_left, pad_right, pad_mode='reflect', axis=-1,
-        backend_name='numpy'):
+def pad(x, pad_left, pad_right, pad_mode='reflect', axis=-1, out=None):
     """Backend-agnostic padding.
 
     Parameters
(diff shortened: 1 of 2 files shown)


OverLordGoldDragon:
TensorFlow being a pain (as usual), can't assign to tensor, and I'm not familiar with performant workarounds. Dunno what to do here except restrict to `max_pad_factor = 1` for `pad_mode="reflect"` and fallback on `tf.pad`.

OverLordGoldDragon:
> A layer with functionality common to all backends that can rely on ducktyping can indeed be added, but it should import and inherit from the backends, or provide a mixin. Crucially: without importing the backend packages, which makes it not agnostic.

> inherit from the backends, or provide a mixin

That seems much uncleaner. What's the point? 

> makes it not agnostic.

It can't be 100% agnostic as some tensor manipulations aren't accessible as the tensor's attributes, but it's minimal enough to call "agnostic".

MuawizChaudhary:
This agnostic backend seems to be a mess with all the if statements for specific frameworks. The current way we do the backends is better. Implementation of operations look much clearer and cleaner, and we have more control over how to do implementation in each framework.

PR #739, opened 2021-05-04
Improve `compute_minimum_support_to_pad`

OverLordGoldDragon:
Code builds on #673, #727, #728, #734, arguably #738, and a not-made PR on `max_pad_factor`, and cannot run standalone; PR is to isolate code for review. It's fully functional in JTFS.

#673 introduces possibility of `psi` having greater temporal support than `phi`; this PR accounts for it, and for possibility of padding by more than `N`.

### DIFF
--- a/kymatio/scattering1d/utils.py
+++ b/kymatio/scattering1d/utils.py
@@ -176,16 +176,19 @@ def compute_minimum_support_to_pad(N, J, Q, T, criterion_amplitude=1e-3,
     else:
         psi2_halfwidth = -1
 
-    # take maximum
-    t_max = max(phi_halfwidth, psi1_halfwidth, psi2_halfwidth)
+    # set min to pad based on each; take a little extra on each to be safe
+    pads = [int(1.1 * hw) for hw in
+            (phi_halfwidth, psi1_halfwidth, psi2_halfwidth)]
 
-    # set min to pad based on maximum
-    min_to_pad = int(1.2 * t_max)  # take a little extra to be safe
+    # can pad half as much
     if pad_mode == 'zero':
-        min_to_pad //= 2
+        pads = [p//2 for p in pads]
+    pad_phi, pad_psi1, pad_psi2 = pads
+    # set main quantity as the max of all
+    min_to_pad = max(pads)
 
     # return results
-    return min_to_pad
+    return min_to_pad, pad_phi, pad_psi1, pad_psi2
 
 
 def precompute_size_scattering(J, Q, max_order=2, detail=False):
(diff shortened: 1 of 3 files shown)


OverLordGoldDragon:
Reverted commit; without it a filter may fail to decay to zero, but that seems fine as "left won't draw from right & vice versa" will still hold.

OverLordGoldDragon:
> Are dyadic increments enough here?

I don't follow. We only sample at dyadic, so a dyadic increment is the only possibility.

OverLordGoldDragon:
This one's for conceptual review rather than merging; will make latter upon approval.

OverLordGoldDragon:
Rather, with a few changes, this can merge after #673, #728

kymatio/scattering1d/utils.py
```diff
@@ -1,6 +1,8 @@
 import numpy as np
 import math
-from .filter_bank import scattering_filter_factory, calibrate_scattering_filters
+from .filter_bank import (calibrate_scattering_filters, compute_temporal_support,
+                          compute_minimum_required_length, gauss_1d, morlet_1d)
```
MuawizChaudhary, line 4: why parenthesis??? remove it

kymatio/scattering1d/utils.py
```diff
@@ -64,9 +66,10 @@ def compute_padding(J_pad, T):
         raise ValueError('Too large padding value, will lead to NaN errors')
     return pad_left, pad_right
 
-def compute_minimum_support_to_pad(T, J, Q, criterion_amplitude=1e-3,
+def compute_minimum_support_to_pad(N, J, Q, T, criterion_amplitude=1e-3,
```
MuawizChaudhary, line 69: why the switch?

kymatio/scattering1d/utils.py
```diff
@@ -114,20 +124,70 @@ def compute_minimum_support_to_pad(T, J, Q, criterion_amplitude=1e-3,
...
+        Name of padding used. If 'zero', will halve `min_to_pad`, else no effect.
 
     Returns
     -------
     min_to_pad: int
         minimal value to pad the signal on one size to avoid any
         boundary error.
     """
-    J_tentative = int(np.ceil(np.log2(T)))
-    _, _, _, t_max_phi = scattering_filter_factory(
```
MuawizChaudhary, line 125: good work

MuawizChaudhary (commented):
Looks mostly good to me.

kymatio/scattering1d/utils.py
MuawizChaudhary: ignore.

PR #740, opened 2021-05-06
Global average pooling

OverLordGoldDragon:
Building on #673, will detect `T == N` and take `mean()`, which is much faster than dot-multiply with `phi`, subsampling, `ifft`, and unpadding - while producing exact same result.

Didn't implement backend `mean`, proof of concept PR.

### DIFF
--- a/kymatio/scattering1d/frontend/numpy_frontend.py
+++ b/kymatio/scattering1d/frontend/numpy_frontend.py
@@ -51,7 +51,7 @@ class ScatteringNumPy1D(ScatteringNumPy, ScatteringBase1D):
             size_scattering = 0
 
         S = scattering1d(x, self.backend.pad, self.backend.unpad, self.backend, self.J, self.psi1_f, self.psi2_f,
-                         self.phi_f, max_order=self.max_order, average=self.average, global_average=self.global_average, pad_left=self.pad_left,
+                         self.phi_f, max_order=self.max_order, average=self.average, average_global=self.average_global, pad_left=self.pad_left,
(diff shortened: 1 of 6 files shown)


OverLordGoldDragon:
> This seems too specific for a backend function. 

What do you mean?

> Why not simply add mean if at all necessary 

> give keepdimS (numpy convention) as an argument?

I can do that, but I thought kymatio's conservative on adding kwargs to backend not used anywhere (e.g. `fft` lacked `axis`)

lostanlen:
LGTM. Did you observe a speedup?

OverLordGoldDragon:
Never tried, surely `mean(U1)` beats `unpad(ifft(subsample(U1 * phi)))`

kymatio/scattering1d/frontend/numpy_frontend.py
```diff
@@ -51,7 +51,7 @@ def scattering(self, x):
             size_scattering = 0
 
         S = scattering1d(x, self.backend.pad, self.backend.unpad, self.backend, self.J, self.psi1_f, self.psi2_f,
-                         self.phi_f, max_order=self.max_order, average=self.average, pad_left=self.pad_left,
```
MuawizChaudhary, line 54: why is this only changed for the numpy frontend? Why not torch and tf?

MuawizChaudhary (commented):
Can you test this actually gives speedup? How much?

lostanlen:
This feature is a priority IMO but i think it needs to be paired with a benchmark and a tutorial demonstrating the use

PR #742, opened 2021-05-09
Lower default `alpha`

OverLordGoldDragon:
Current `alpha` is too conservative, reducing number of second-order coefficients. It should either be exposed, or lowered to `4`.

Example with `Q=1`:

image

At green line (red is current behavior), the ratio of filter's maximum to the point at which it's subsampled is `46909`, which meets `criterion_amplitude` 46-fold. In `scattering.m`, this filter counts toward second order (confirmed by @lostanlen).

Below I show exhaustively that `alpha=4` is a safe value for two combinations of `Q, J` meant to test extremes; colors map subsampling factors. Code shows more plots.

<details>
  <summary><b>code</b></summary>

```python
import numpy as np
from kymatio import Scattering1D
from kymatio.visuals import filterbank_scattering, plot


def viz_bounds(psi_f, idx, sub_up, last_idx=None):
    p = psi_f[idx]
    sub0 = len(p[0]) // 2**(p['j'] + 1)
    sub1 = sub0 // 2 if sub_up else sub0
    pt = p[0][:last_idx]
    
    plot(pt, title="psi_f[%s]" % idx)
    plot([], vlines=(sub0, {'color': 'tab:red'}))
    plot([], vlines=(sub1, {'color': 'tab:green'}), show=1)
    
    add_txt = ("\np.max() / p[sub + 1] = {:.1f}".format(pt.max() / pt[sub1])
               if sub_up else "")
    print(("p=psi1_f[2]; sub = len(p) // 2**(p['j'] + 1)\n"
           "p.max() / p[sub]     = {:.1f}{}").format(pt.max() / pt[sub0], add_txt))

def viz_ratios(psi_f):
    ratios = []
    for p in psi_f:
        pt = p[0]
        sub = len(pt) // 2**(p['j'] + 1)
        ratios.append(pt.max() / pt[sub])
    plot(np.log10(ratios), show=1,
         title="log10(ratio) vs index | (max(psi) / psi[point_of_subsampling])")
    
#%%
N = 4096
for alpha in (5., 4.):
    kw = dict(shape=N, alpha=alpha, frontend='numpy')
    
    scattering = Scattering1D(J=8, Q=1, **kw)
    filterbank_scattering(scattering)
    viz_bounds(scattering.psi1_f, idx=2, sub_up=(alpha == 5))
    viz_ratios(scattering.psi1_f)
    
    scattering = Scattering1D(J=10, Q=16, **kw)
    filterbank_scattering(scattering)
    viz_bounds(scattering.psi1_f, idx=17, sub_up=(alpha == 5))
    viz_ratios(scattering.psi1_f)
    
    if alpha == 4:
        viz_bounds(scattering.psi1_f, idx=-1, sub_up=False, last_idx=32)
```
</details>

image

### DIFF
--- a/kymatio/scattering1d/filter_bank.py
+++ b/kymatio/scattering1d/filter_bank.py
@@ -310,7 +310,7 @@ def compute_temporal_support(h_f, criterion_amplitude=1e-3):
     return N
 
 
-def get_max_dyadic_subsampling(xi, sigma, alpha=5.):
+def get_max_dyadic_subsampling(xi, sigma, alpha=4.):
     """
     Computes the maximal dyadic subsampling which is possible for a Gabor
     filter of frequency xi and width sigma
@@ -333,7 +333,7 @@ def get_max_dyadic_subsampling(xi, sigma, alpha=5.):
         frequential width of the filter
     alpha : float, optional
         parameter controlling the error done in the aliasing.
-        The larger alpha, the smaller the error. Defaults to 5.
+        The larger alpha, the smaller the error. Defaults to 4.
 
     Returns
     -------
@@ -347,7 +347,7 @@ def get_max_dyadic_subsampling(xi, sigma, alpha=5.):
     return j
 
 
-def move_one_dyadic_step(cv, Q, alpha=5.):
+def move_one_dyadic_step(cv, Q, alpha=4.):
     """
     Computes the parameters of the next wavelet on the low frequency side,
     based on the parameters of the current wavelet.
@@ -374,7 +374,7 @@ def move_one_dyadic_step(cv, Q, alpha=5.):
         the frequency and width of the current wavelet and the next wavelet.
     alpha : float, optional
         tolerance parameter for the aliasing. The larger alpha,
-        the more conservative the algorithm is. Defaults to 5.
+        the more conservative the algorithm is. Defaults to 4.
 
     Returns
     -------
@@ -409,7 +409,7 @@ def compute_xi_max(Q):
     return xi_max
 
 
-def compute_params_filterbank(sigma_low, Q, r_psi=math.sqrt(0.5), alpha=5.):
+def compute_params_filterbank(sigma_low, Q, r_psi=math.sqrt(0.5), alpha=4.):
     """
     Computes the parameters of a Morlet wavelet filterbank.
 
@@ -440,7 +440,7 @@ def compute_params_filterbank(sigma_low, Q, r_psi=math.sqrt(0.5), alpha=5.):
     alpha : float, optional
         tolerance factor for the aliasing after subsampling.
         The larger alpha, the more conservative the value of maximal
-        subsampling is. Defaults to 5.
+        subsampling is. Defaults to 4.
 
     Returns
     -------
@@ -490,7 +490,7 @@ def compute_params_filterbank(sigma_low, Q, r_psi=math.sqrt(0.5), alpha=5.):
 
 
 def calibrate_scattering_filters(J, Q, r_psi=math.sqrt(0.5), sigma0=0.1,
-                                 alpha=5.):
+                                 alpha=4.):
     """
     Calibrates the parameters of the filters used at the 1st and 2nd orders
     of the scattering transform.
@@ -521,7 +521,7 @@ def calibrate_scattering_filters(J, Q, r_psi=math.sqrt(0.5), sigma0=0.1,
     alpha : float, optional
         tolerance factor for the aliasing after subsampling.
         The larger alpha, the more conservative the value of maximal
-        subsampling is. Defaults to 5.
+        subsampling is. Defaults to 4.
 
     Returns
     -------
@@ -552,7 +552,7 @@ def calibrate_scattering_filters(J, Q, r_psi=math.sqrt(0.5), sigma0=0.1,
 
 def scattering_filter_factory(J_support, J_scattering, Q, r_psi=math.sqrt(0.5),
                               criterion_amplitude=1e-3, normalize='l1',
-                              max_subsampling=None, sigma0=0.1, alpha=5.,
+                              max_subsampling=None, sigma0=0.1, alpha=4.,
                               P_max=5, eps=1e-7, **kwargs):
     """
     Builds in Fourier the Morlet filters used for the scattering transform.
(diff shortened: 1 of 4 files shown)


OverLordGoldDragon:
This can merge at any point, but breaks output tests (number of second-order coefficients increases).

OverLordGoldDragon:
I'll have to redo this analysis in terms of an L1 measure; I realized the sum after the tail is significant. Likewise for padding computation (that'd revert to the previous method but with a correction).

lostanlen:
This looks good but frankly i don't understand why we need both a parameter `alpha` and another one called `criterion_amplitude` which controls "how sigma maps to j for psi". (https://github.com/kymatio/kymatio/issues/789#issuecomment-1003783217)

`upper_bound = min(xi + alpha * sigma, 0.5)`

Also, this sort of design choice looks like it has one clear best value (`4`?) and no use case outside of that value; hence, i would be in favor in setting it to a constant, and documenting it as such.

"We define the cutoff frequency of a wavelet as its center frequency plus four standard deviations. This cutoff frequency defines the sampling rate of the sequence of scattering coefficients for the corresponding path."

What role does `criterion_amplitude` play in this?

OverLordGoldDragon:
> i don't understand why we need both a parameter alpha and another one called criterion_amplitude 

Actually, good catch: we don't. `criterion_amplitude` in original Kymatio is used only for warning about boundary effects, i.e. determining time-domain support. `alpha` is simply the frequency-domain equivalent - but it can, and _should_, be replaced by `criterion_amplitude` (not necessarily the same value but it can very well work), since currently `alpha` does

```
xi + sigma * alpha
```

to determine which octave the filter belongs to. This is faulty, however, since it relies on the continuous-time expression accurately mapping to the discrete-time sampling, which is often unreliable, and is certainly inaccurate for very low `xi`: computing directly against `criterion_amplitude` is superior.

> setting it to a constant

Yes it already is and I favor keeping it that way

lostanlen:
OK very good! Sounds like a plan: get rid of tunable alpha and rely on criterion_amplitude to determine the critical sampling of a given scattering path.
Regarding `criterion_amplitude`, i can imagine why it would be tunable: it needs to be lower in reconstruction than in classification, for example.
But wanting 3 orders of magnitude of decay (`1e-3`) in the time domain seems restrictive, and i see why you'd want to circumvent the warning, at least in some cases. 
I imagine that a good amount of this discussion could be anticipated from the numerical values of the inverse error function (erfinv) instead of by trial-and-error ?

Maybe take inspiration from Scattering2D and 3D for deciding this? Surely they don't pad as much as we do ...

OverLordGoldDragon:
> i see why you'd want to circumvent the warning

`warn=False` is solely for `filter_bank.compute_minimum_required_length`: this method deliberately pushes the limits to find where the break point is (basically, its very goal is to achieve the warning). Otherwise, we should always warn.

`1e-3` is a very safe value; per my "information preservation" formula, which uses a _square_ measure (which I confirmed to work much better than absolute), the cutoff is at contributions being `1e-6` of maximum. I'd certainly compare it against `alpha=4` before replacing, since `alpha=4` works well. (though, I think I arrived at `4` in the first place precisely via `1e-3`)

> it needs to be lower in reconstruction than in classification

Sure, but I'd favor settling for a single value that qualifies as "nearly lossless", and leave downsampling to downsampling-related processing steps.

> Maybe take inspiration from Scattering2D and 3D for deciding this?

... entirely possible they don't pad enough either. What I know is, perfectly satisfying `1e-3` requires x16 padding _if_ using `J = log2_N`, but x8 can be entirely satisfactory in many cases. This is controlled by `max_pad_factor`, and wherever slightest amounts of boundary effects matter, it helps the user to have the "perfection" option.

lostanlen:
>  I think I arrived at 4 in the first place precisely via 1e-3

Yes, that's also what i understood from your OP. It's weird to have to change a default value to satisfy another default value. I don't see the appeal behind two degrees of freedom here when there really should be one.

It's also worth reminding ourselves that there is nothing special about this value `1e-3`. Historically, it was chosen by Mathieu Andreux so as to seek a principled way to pad his time series. We should notworship that value, as it entails no guarantee on the transform.

I don't think Andreux ever brought `J` ever close, let alone equal, to `log2(N)` — nor did I. In most signal processing problems, we choose how much signal we acquire (`N`) based on how we want to represent it, not the other way around. So the answer to the warning `Signal support is too small to avoid border effects` shouldn't be "let's pad some more" but really "let's increase the signal support" or perhaps "let's lower the number of octaves and use `average_global`".

> What I know is, perfectly satisfying 1e-3 requires x16 padding if using J = log2_N, but x8 can be entirely satisfactory in many cases. 

Padding the signal 8-fold or 16-fold just to achieve a numerical criterion is almost never what the user wants in practice. I am worried that it is becoming the default behavior and that the user isn't even warned about it (unless they are Kymatio-savvy enough to know when to pass `warn=True`, in which case they don't need to be warned anymore).

MuawizChaudhary:
If I understand correctly, this changes the number of scattering coefficents. Currently we test if we match results from a previous scattering package, so this breaks the tests. For various, what appears to be theroetical, reasons, changing alpha is 'better'. 

I would like this benchmarked; does the accuracy of a linear regression trained on the new set of scattering coefficents improve? what about a neural network?

We should consider having a set of Kymatio "Benchmark" datasets. If someone makes such a large change as effecting the resulting number of scattering coefficents, or making the scattering optimizable, there should be a set of benchmarks we can run their scattering on to justify the changes. 

Is there a formula involving alpha or J, T, Q, for the resulting number of scattering coeffiecents?

MuawizChaudhary (changes requested):
If I understand correctly, this changes the number of scattering coefficents. Currently we test if we match results from a previous scattering package, so this breaks the tests. For various, what appears to be theroetical, reasons, changing alpha is 'better'.

I would like this benchmarked; does the accuracy of a linear regression trained on the new set of scattering coefficents improve? what about a neural network?

We should consider having a set of Kymatio "Benchmark" datasets. If someone makes such a large change as effecting the resulting number of scattering coefficents, or making the scattering optimizable, there should be a set of benchmarks we can run their scattering on to justify the changes.

Is there a formula involving alpha or J, T, Q, for the resulting number of scattering coeffiecents?

lostanlen:
> Is there a formula involving alpha or J, T, Q, for the resulting number of scattering coeffiecents?

Yes i think @OverLordGoldDragon has one or is working on one.
But the formula looks complicated, unfortunately. Way more complicated than in 2D ...

PR #743, opened 2021-05-11
MathJax

OverLordGoldDragon:
SVG rendering and extended math syntax; done via CDN, no install (unless offline), and works on Windows (unlike `imgmath`).


edouardoyallon (approved):
I'm pros merging.

PR #746, opened 2021-05-25
Deprecate `vectorize`

OverLordGoldDragon:
This fixes dev and passes all 1D tests. But it goes further:

 1. Removed `Modulus` implementations. The cited purpose is "stability" and "not returning NaNs for gradients", but `torch.abs(x)` accomplishes the same AFAIK, and `test_modulus()` passes.
 2. Dropped trailing dim everywhere (i.e. `view_as_real`) except `subsample_fourier` since complex `mean()` isn't implemented in 1.7. I'm not aware of its need beyond modulus; it only complicates code.
 3. Dropped `fft(normalized=False)`. If this was for performance by sparing a division, it makes intuitive sense but I saw no difference in my benchmarks. It makes outputs conflict with numpy's, makes the implementation numerically incorrect, and complicates correctness tests by requiring this difference accounted (and differently for first and second orders).
 4. Dropped `vectorize`, as was planned, and rerouted logic as `True -> out_type=="array" and average`, and `False -> out_type=="list" or not average`.
 5. Improved test debugging and local-friendliness via `run_without_pytest`. 

### Todo

 - [ ] Docs (will change after review)
 - [ ] 2D & 3D (I leave this to devs)

### DIFF
--- a/kymatio/backend/torch_skcuda_backend.py
+++ b/kymatio/backend/torch_skcuda_backend.py
@@ -1,5 +1,4 @@
 import torch
-from skcuda import cublas
 
 
 class TorchSkcudaBackend:
@@ -7,11 +6,11 @@ class TorchSkcudaBackend:
 
     @staticmethod
     def _is_complex(x):
-        return x.shape[-1] == 2
+        return torch.is_complex(x)
 
     @staticmethod
     def _is_real(x):
-        return x.shape[-1] == 1
+        return 'float' in str(x.dtype)
 
     @classmethod
     def cdgmm(cls, A, B):
@@ -50,17 +49,14 @@ class TorchSkcudaBackend:
 
         """
         if not cls._is_complex(A):
-            raise TypeError('The input should be complex (i.e. last dimension is 2).')
+            raise TypeError('The input should be complex (got %s).' % A.dtype)
 
-        if not cls._is_complex(B) and not cls._is_real(B):
+        if not (cls._is_complex(B) or cls._is_real(B)):
             raise TypeError('The filter should be complex or real, indicated by a '
                             'last dimension of size 2 or 1, respectively.')
 
-        if A.shape[-len(B.shape):-1] != B.shape[:-1]:
+        if A.shape[-B.ndim:-1] != B.shape[:-1]:
             raise RuntimeError('The filters are not compatible for multiplication.')
-
-        if A.dtype is not B.dtype:
-            raise TypeError('Input and filter must be of the same dtype.')
 
         if not A.is_cuda or not B.is_cuda:
             raise TypeError('Input and filter must be CUDA tensors.')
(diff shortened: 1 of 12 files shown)


lostanlen:
@janden or @eickenberg or @edouardoyallon, would you have the time to review this?

janden:
First, it will be easier in the long run to rebase dev onto master to solve the torch compatibility issues. I will look into this later today or tomorrow.

Second, given that these are quite separate fixes, it would be better to split them into smaller PRs.

OverLordGoldDragon:
> First, it will be easier in the long run to rebase dev onto master to solve the torch compatibility issues

Agreed the big commit should precede

> split them into smaller PRs.

1 and 2 are dependent. Separating anything would require separate modifications to tests. I think the overall size is reasonable - we can review one thing at a time.

janden:
I would still prefer keeping PRs smaller. It makes for shorter review cycles and prevents features being held up unnecessarily while others are being ironed out.

eickenberg:
I took a look. It uses complex numbers. So let's release the dev-master rebase as is before we get to these more fundamental changes.

kymatio/backend/torch_skcuda_backend.py
```diff
@@ -50,37 +49,19 @@ def cdgmm(cls, A, B):
...
-            m, n = B.nelement() // 2, A.nelement() // B.nelement()
-            lda = m
-            ldc = m
-            incx = 1
-            handle = torch.cuda.current_blas_handle()
-            stream = torch.cuda.current_stream()._as_parameter_
-            cublas.cublasSetStream(handle, stream)
-            cublas.cublasCdgmm(handle, 'l', m, n, A.data_ptr(), lda, B.data_ptr(), incx, C.data_ptr(), ldc)
-            return C
+        return A * B
```
eickenberg, line 67: this is only valid if pytorch actually calls `Cdgmm` which may not be the case (it would need to understand to replace 1d->2d broadcasting with this op. does it do that?)

OverLordGoldDragon:
@eickenberg Preferably number it <0.3.0 to reserve major version for the full merger.

kymatio/backend/torch_skcuda_backend.py
OverLordGoldDragon: Unsure what you mean; complex arithmetic follows same broadcasting rules as real, no? All tests pass.

janden:
Closing this since #749 supersedes.

OverLordGoldDragon:
Why close? 'smaller' is a subset of this, this'll come right after.

OverLordGoldDragon:
Or reopen after merging the other? Same idea as with JTFS

janden:
Sure, as long as the commit history is cleaned up.

MuawizChaudhary (changes requested):
The addition of .github/workflows, benchmarks, citation addition, documentation changes, example changes, tools changes/removal should each be a seperate, well needed, PR. 

Please seperate this PR out into 5-7 seperate PRs.

PR #751, opened 2021-06-03
Remove custom tensorflow build

dylex:
We've now added v100 GPUs to jenkins, so the custom tensorflow build is no longer necessary.  (Alternatively we could keep it and add support for the new GPUs so we can run on both, but probably there's not much point.)


janden:
Great! This will make things much easier. Thank you.

janden (approved):

PR #766, opened 2021-07-25
Add `width` meta

OverLordGoldDragon:
Useful in general, and used in a standard variant of JTFS; see [discussion](https://github.com/kymatio/kymatio/discussions/730#discussioncomment-1018400)


MuawizChaudhary (commented):
I think this should be merged after we've merged the previous PRs.

PR #770, opened 2021-08-24
Introduce `analytic`

OverLordGoldDragon:
`True` forces analyticity by zeroing negative frequencies; useful in JTFS, especially with `r_psi > sqrt(.5)`. Defaults to `False` for `Scattering1D`.

### DIFF
--- a/kymatio/scattering1d/frontend/base_frontend.py
+++ b/kymatio/scattering1d/frontend/base_frontend.py
@@ -11,7 +11,8 @@ compute_meta_scattering, precompute_size_scattering)
 
 class ScatteringBase1D(ScatteringBase):
     def __init__(self, J, shape, Q=1, max_order=2, average=True,
-            oversampling=0, vectorize=True, out_type='array', backend=None):
+            oversampling=0, vectorize=True, out_type='array', analytic=False,
+            backend=None):
         super(ScatteringBase1D, self).__init__()
         self.J = J
         self.shape = shape
@@ -21,6 +22,7 @@ class ScatteringBase1D(ScatteringBase):
         self.oversampling = oversampling
         self.vectorize = vectorize
         self.out_type = out_type
+        self.analytic = analytic
         self.backend = backend
 
     def build(self):
@@ -75,6 +77,16 @@ class ScatteringBase1D(ScatteringBase):
             criterion_amplitude=self.criterion_amplitude,
             r_psi=self.r_psi, sigma0=self.sigma0, alpha=self.alpha,
             P_max=self.P_max, eps=self.eps)
+
+        # analyticity
+        if self.analytic:
+          for psi_fs in (self.psi1_f, self.psi2_f):
+            for p in psi_fs:
+              for k in p:
+                if isinstance(k, int):
+                    M = len(p[k])
+                    p[k][M//2 + 1:] = 0  # zero negatives
+                    p[k][M//2] /= 2      # halve Nyquist
 
     def meta(self):
         """Get meta information on the transform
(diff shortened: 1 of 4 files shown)


lostanlen:
@janden thoughts?

lostanlen:
I don't think it's a good idea, nor do i think we should allow `r_psi>sqrt(2)/2` for `Q=1`. We'd be veering way too much outside of the Morlet design.

At this rate, better to design a new wavelet entirely.

OverLordGoldDragon:
> I don't think it's safe to take the analytic part

Why?

lostanlen:
Because it creates a discontinuity at Nyquist and thus destroys time localization (making it essentally sinc-like).
Hence the temporal envelope will no longer look like a Morlet at all.

Same problem with `r_psi>0.5`  under `Q=1`. This will cause the corrective term of Morlet at 0 Hz to grow in amplitude and become far from negligible. Hence the spectral envelope won't look Gaussian anymore, but asymmetric rather.

At this stage, we'd be losing so many mathematical properties of Gabor (time-frequency localization, chiefly) that we might as well design another wavelet from scratch. But that's not a short-term goal.

OverLordGoldDragon:
Surprisingly, there's little for worry.

Forced analytic Morlets are used in `ssqueezepy` for instantaneous AM & FM localization with success.

[Generalized Morse Wavelets](https://overlordgolddragon.github.io/generalized-morse-wavelets/) are _by design_ forced analytic. They are superior to Morlets per having more vanishing moments - however, I found that the two are nearly identical except for 1) very low frequencies; 2) high time localization (large `bw`).

> destroys time localization

Not so; the time resolution integral, with a squared weight, [still decays](https://github.com/jonathanlilly/jLab/issues/13#issuecomment-736834193) (if we handle the Nyquist bin properly). The effect is reduced much further after temporal averaging; if there's a difference vs. ideally time localized filters, we could hardly tell.

What's more important for scattering is analyticity. `|analytic(x)| = amplitude(x)` and `angle(analytic(x)) = phase(x)` break down the more "`analytic`" is inexact - which is bad news for second order time scattering, which is supposed to operate on `amplitude(x)`. It's even more important for frequential scattering in JTFS: low frequency `psi_fr` have the greatest energy, and any leaking into negatives significantly amplifies the opposite spin. In time domain, imperfect analyticity manifests as [breakdown of real-imag symmetries](https://github.com/kymatio/kymatio/discussions/752#discussioncomment-1124734). You can see the effect for yourself by passing `analytic=False` to `test_up_vs_down` in `test_jtfs.py`.

> Q=1

This is per the 2) and is a valid concern. I was thinking of making `analytic` a tuple like `Q` or `r_psi` for this; the ideal might be `(True, False)` for time scattering and `(True, False, True)` for JTFS, since there is no third order. But even with `Q_fr=1`, `True` wins.

So it's a tradeoff between time localization and analyticity. Since gains in latter far outweigh losses in former, I recommend `analytic=True` be the default, at least for JTFS.

MuawizChaudhary:
> I don't think it's safe to take the analytic part, nor do i think we should allow `r_psi>sqrt(2)/2` for `Q=1`. We'd be veering way too much outside of the Morlet design.
> 
> At this rate, better to design a new wavelet entirely.

With the changes introduced in the previous PRs, would we be able to introduce a generalized gabor wavelet scattering?

lostanlen:
I would be more in favor of designing a Generalized Morse Wavelet module (or other idea of the same kind) than loading Gabor with all sorts of opt-in keyword arguments. It will confuse the hell out of our users and make the Gaussianity of the Gabor less and less recognizable. As we say in French, `analytic` feels like a bandage on a wooden leg.

The same is true of `r_psi`, which has far-reaching and unintuitive consequences.

PR #764, opened 2021-07-25
Revamp `xi_max`

OverLordGoldDragon:
It's too low. Analyticity is worse with greater, but poor coverage beats no coverage.

image

(Note: LP sum images assume #765 merged)

PR #781, opened 2021-10-10
JTFS minimal

OverLordGoldDragon:


kymatio/numpy.py
```diff
@@ -1,4 +1,6 @@
-from .scattering1d.frontend.numpy_frontend import ScatteringNumPy1D as Scattering1D
+from .scattering1d.frontend.numpy_frontend import (
+    ScatteringNumPy1D as Scattering1D,
+    TimeFrequencyScatteringNumPy1D as TimeFrequencyScattering1D)
```
rastegah, line 3: why is `kymatio.numpy.TimeFrequencyScattering1D` exposed, but `kymatio.torch.TimeFrequencyScattering1D` is not?

OverLordGoldDragon: `jtfs-min` is [numpy only](https://user-images.githubusercontent.com/16495490/139905940-29c47d6c-4ca5-478b-bf35-c19f6f4c2867.png)

tests/scattering1d/test_jtfs.py
```diff
@@ -0,0 +1,372 @@
+# -*- coding: utf-8 -*-
+"""Joint Time-Frequency Scattering related tests."""
+import pytest
+import numpy as np
+from pathlib import Path
+from kymatio import Scattering1D, TimeFrequencyScattering1D
```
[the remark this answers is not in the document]
OverLordGoldDragon: As mentioned, `jtfs-min` is numpy only; only the main branch supports torch. Usage for `jtfs-min` is any of

```python
from kymatio import TimeFrequencyScattering1D
import kymatio; kymatio.TimeFrequencyScattering1D
from kymatio.numpy import TimeFrequencyScattering1D

jtfs = TimeFrequencyScattering1D(..., frontend='numpy')  # first two
jtfs = TimeFrequencyScattering1D(...)  # last
```
The default `torch` indeed won't work, but I kept it to minimize code changes. Also I believe `numpy` is planned to be the default in `0.3` or `0.4`.

kymatio/numpy.py
```diff
@@ -1,4 +1,6 @@
-from .scattering1d.frontend.numpy_frontend import ScatteringNumPy1D as Scattering1D
+from .scattering1d.frontend.numpy_frontend import (
```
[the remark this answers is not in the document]
OverLordGoldDragon: I assume the latest commit fixes this

kymatio/scattering1d/filter_bank.py
```diff
@@ -253,7 +261,7 @@ def compute_sigma_psi(xi, Q, r=math.sqrt(0.5)):
     return xi * term1 * term2
 
 
-def compute_temporal_support(h_f, criterion_amplitude=1e-3):
+def compute_temporal_support(h_f, criterion_amplitude=1e-3, warn=False):
```
rastegah, line 264: dare I say that `criterion_amplitude` is unlikely to change? hence, when will `warn` be necessary? for who is its exposure useful?

lostanlen: The behavior of `warn=True` may be obtained via a Python context manager:

```
with warnings.catch_warnings():
    warning.simplefilter('ignore')
```

so i don't think this `warn=False` new kwarg is necessary.

I don't fully understand the implications of `criterion_amplitude`.

kymatio/scattering1d/filter_bank.py
```diff
@@ -284,33 +292,92 @@ def compute_temporal_support(h_f, criterion_amplitude=1e-3):
...
+            N -= 1
     else:
-        # if there is none:
+        # if there are none
         N = half_support
-        # Raise a warning to say that there will be border effects
-        warnings.warn('Signal support is too small to avoid border effects')
+        if warn:
+            # Raise a warning to say that there will be border effects
+            warnings.warn('Signal support is too small to avoid border effects')
```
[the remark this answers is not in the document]
rastegah: let's remedy this so the filterbank can be ready to merge :)

kymatio/scattering1d/frontend/base_frontend.py
```diff
@@ -50,31 +62,96 @@ def build(self):
...
-        # to avoid padding more than N - 1 on the left and on the right,
-        # since otherwise torch sends nans
-        J_max_support = int(np.floor(np.log2(3 * self.N - 2)))
-        self.J_pad = min(int(np.ceil(np.log2(self.N + 2 * min_to_pad))),
-                         J_max_support)
+            normalize=self.normalize, pad_mode=self.pad_mode)
+        if self.average_global:
+            min_to_pad = max(pad_psi1, pad_psi2)  # ignore phi's padding
+
+        J_pad = math.ceil(math.log2(self.N + 2 * min_to_pad))
```
rastegah, line 117: can we move the instantiation of `J_pad` out of the frontend? Perhaps it is only necessary to compute it _within_ `scattering_filter_factory` so as to preserve the usability of this function, hiding the padding computation from the user

kymatio/scattering1d/frontend/base_frontend.py
```diff
@@ -175,6 +253,42 @@ def output_size(self, detail=False):
...
             concatenation of all scattering coefficients. Defaults to
             `'array'`.
+        pad_mode : str (default 'reflect') / function, optional
+            Name of padding scheme to use, one of (`x = [1, 2, 3]`):
+                - zero:    [0, 0, 1, 2, 3, 0, 0, 0]
+                - reflect: [3, 2, 1, 2, 3, 2, 1, 2]
+            Or, pad function with signature `pad_fn(x, pad_left, pad_right)`.
+            This sets `self.pad_mode='custom'` (the name of padding is used
+            for some internal logic).
+        max_pad_factor : int (default 2), optional
```
lostanlen, line 263: This looks terrifying. Why would you ever want to pad a signal of size `256` by _twice_ its length on both sides?

kymatio/scattering1d/frontend/base_frontend.py
```diff
@@ -368,4 +523,1496 @@ def _document(cls):
             n=cls._doc_array_n)
 
 
-__all__ = ['ScatteringBase1D']
+class TimeFrequencyScatteringBase1D():
+    def __init__(self, J_fr=None, Q_fr=2, F=None, implementation=None,
```
lostanlen, line 527: it's not clear to me what `implementation=None` does? why not `implementation=1` by default?

kymatio/scattering1d/frontend/base_frontend.py
```diff
@@ -368,4 +523,1496 @@ def _document(cls):
...
+          - If `'global'`, sets to maximum possible `F` based on `N_frs_max`.
+          - Used even with `average_fr=False` (see its docs); this is likewise
+            true of `T` for `phi_t * phi_f` and `phi_t * psi_f` pairs.
+
+    implementation : int / None
+        Preset configuration to use. Overrides the following parameters:
+
+            - `average_fr, aligned, out_3D, sampling_filters_fr`
+
+        Defaults to `None`, and any `None` argument above will default to
```
lostanlen, line 922: this sentence is unclear. Can anyone help?

kymatio/scattering1d/frontend/base_frontend.py
```diff
@@ -368,4 +523,1496 @@ def _document(cls):
...
+    pad_mode_fr : str (default 'conj-reflect-zero') / function
+        Name of frequential padding mode to use, one of: 'zero',
+        'conj-reflect-zero'.
+        Or, function with signature `pad_fn_fr(x, pad_fr, scf, B)`;
+        see `_right_pad` in
+        `kymatio.scattering1d.core.timefrequency_scattering1d`.
+
+        If using `pad_mode = 'reflect'` and `average = True`, reflected portions
+        will be automatically conjugated before frequential scattering to avoid
+        spin cancellation. For same reason there isn't `pad_mode_fr = 'reflect'`.
```
lostanlen, line 1114: Brilliant !!!!

kymatio/scattering1d/frontend/base_frontend.py
```diff
@@ -368,4 +523,1496 @@ def _document(cls):
...
+
+    analytic : bool (default True)
```
lostanlen, line 1116: `analytic==True` ? Time scattering doesn't do this.
I'd rather play with the xi_max and the r_psi to adjust this.
Can we revert to `analytic==False`?

kymatio/scattering1d/filter_bank.py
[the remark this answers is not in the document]
OverLordGoldDragon: Why? A warning is meant to be informative - if we throw false alarms, users will ignore them along legitimate ones.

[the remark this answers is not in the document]
lostanlen: If there's false alarms, it means the logic behind the warning should be fixed.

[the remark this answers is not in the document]
OverLordGoldDragon: There's nothing to fix, I'm just reusing a method that happens to throw warnings. A workaround is to have that method call a sub-method without warnings, but that's an extra function definition vs extra kwarg. These functions are purely internal to begin with.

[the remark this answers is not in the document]
OverLordGoldDragon: An alternative is to just catch the warning via `with`. That's more verbose if done multiple times, but I'm fine with it.

kymatio/numpy.py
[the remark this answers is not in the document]
rastegah: > resolved?

seems so

kymatio/scattering1d/frontend/base_frontend.py
rastegah: implementation `1` seems like the safest default option. I guess the `None` option lets the user build their own implementation?

kymatio/scattering1d/frontend/base_frontend.py
rastegah: if `implementation=None` and ``average_fr | aligned | out_3D | sampling_filters_fr == None` then `implementation=1` is used.

rastegah: If a user is going beyond implementations 2/3, could we assume they have already figured out the parameters to achieve their implementation? What if I want a `3D` output without `average_fr`? 

Is it necessary to have these options? Or would it be better to document possible options and use cases with examples?

kymatio/scattering1d/core/scattering1d.py
```diff
@@ -64,17 +70,17 @@ def scattering1d(x, pad, unpad, backend, J, psi1, psi2, phi, pad_left=0,
 
     # S is simply a dictionary if we do not perform the averaging...
     batch_size = x.shape[0]
-    kJ = max(J - oversampling, 0)
+    kJ = max(log2_T - oversampling, 0)
```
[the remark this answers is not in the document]
OverLordGoldDragon: Think we agreed to keep cleanups separate?

kymatio/scattering1d/filter_bank.py
```diff
@@ -284,33 +292,92 @@ def compute_temporal_support(h_f, criterion_amplitude=1e-3):
...
+            raise Exception("couldn't satisfy stop criterion before `N > 1e9`; "
+                            "check `fn`")
+        if 2 * p_halfwidth < N or (max_N is not None and N > max_N):
+            break
+        N *= 2
     return N
 
 
-def get_max_dyadic_subsampling(xi, sigma, alpha=5.):
+def get_max_dyadic_subsampling(xi, sigma, alpha=4.):
```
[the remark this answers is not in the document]
OverLordGoldDragon: `oversampling` isn't related to this in any way. And this is a private method, `alpha` is in the kwarg so we can set the value in one place and have it propagate internally rather changing it everywhere each time.

kymatio/scattering1d/frontend/base_frontend.py
OverLordGoldDragon: At this point there's lots such computations being done in the frontend. If that kind of refactoring is desired, it can be done separately, but I think keeping some in frontend is reasonable and adds clarity. It'd also require moving steps out of `build()` to after `create_filters()` for `Scattering1D`.

kymatio/scattering1d/filter_bank.py
```diff
@@ -712,22 +823,675 @@ def scattering_filter_factory(J_support, J_scattering, Q, r_psi=math.sqrt(0.5),
...
+def conj_fr(x):
+    """Conjugate in frequency domain by swapping all bins (except dc);
+    assumes frequency along first axis.
+    """
+    out = np.zeros_like(x)
+    out[0] = x[0]
+    out[1:] = x[:0:-1]
+    return out
+
+def compute_temporal_width(p_f, N=None, pts_per_scale=6, fast=True,
```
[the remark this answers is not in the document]
OverLordGoldDragon: > The last four keyword arguments are unused!

No, only `pts_per_scale` is unused, and that I left since it helps with prototyping if necessary - but I'm fine with hiding it.

> Do we really need two functions

Yes. Refer to our Discord discussions, also [here](https://github.com/kymatio/kymatio/discussions/730#discussioncomment-1018400)

kymatio/scattering1d/filter_bank.py
```diff
@@ -405,18 +472,18 @@ def compute_xi_max(Q):
     xi_max : float
         largest frequency of the wavelet frame.
     """
-    xi_max = max(1. / (1. + math.pow(2., 3. / Q)), 0.35)
+    xi_max = max(1. / (1. + math.pow(2., 1. / Q)), 0.4)
```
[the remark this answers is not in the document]
OverLordGoldDragon: Why what? Do you refer to extending the docs beyond original? The rationale is [here](https://github.com/kymatio/kymatio/pull/764)

kymatio/scattering1d/filter_bank.py
```diff
@@ -405,18 +472,18 @@ def compute_xi_max(Q):
...
     return xi_max
 
 
-def compute_params_filterbank(sigma_low, Q, r_psi=math.sqrt(0.5), alpha=5.):
+def compute_params_filterbank(sigma_min, Q, r_psi=math.sqrt(0.5), alpha=4.,
```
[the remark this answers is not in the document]
OverLordGoldDragon: Not exposed to user

kymatio/scattering1d/filter_bank.py
```diff
@@ -566,21 +670,32 @@ def scattering_filter_factory(J_support, J_scattering, Q, r_psi=math.sqrt(0.5),
         Each value for k is an array (or tensor) of size 2**(J_support - k)
         containing the Fourier transform of the filter after subsampling by
         2**k
+    * 'width': temporal width
```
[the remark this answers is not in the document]
OverLordGoldDragon: I'll do it. I kept it as brief as original but never liked original to begin with

kymatio/scattering1d/frontend/base_frontend.py
OverLordGoldDragon: `None` will set any params unspecified by the user to that of `implementation=1`.

`not None` will _override_ any params specified by the user by that of `3` (or whatever).

It's the only way to distinguish user-supplied args from defaulted args, besides defaulting everything else to `None` in the function declaration, which we don't want.

kymatio/scattering1d/filter_bank.py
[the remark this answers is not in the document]
lostanlen: It's still not clear to me what the chain of dependencies is here, between user-defined T and filter design. It's also not clear what sigma0 does in this context.

At the end of the day, what the user wants is to design a representation which guarantees a certain amount of invariance T. Your function does the opposite: measure the invariance of a filter that's already designed.

If the invariance is not feasible because the filter can't be designed (signal too short), we shouldn't hesitate to raise a helpful error message.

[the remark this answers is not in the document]
lostanlen: (although, to your credit, this is not possible in frequency scattering, in which case we should indeed pad the frequency domain to allow a filter of width F being constructible)

lostanlen:
@OverLordGoldDragon so, what i don't get is that this PR is supposedly "minimal" in the sense that it has no Torch nor TensorFlow backends, only NumPy; but at the same time it introduces many breaking changes to Scattering1D code, including new default values for certain parameters.

Also, it doesn't pass the unit tests. Is it because of the new alpha value? Does that mean we need to redo regression tests? I'd rather avoid that in the JTFS PR, which is already very heavy on its own. The unit tests must pass.

Lastly, what is the timeline for merging the Torch and TensorFlow backends in your opinion? Making JTFS differentiable and GPU-friendly is very important, of course

kymatio/scattering1d/frontend/base_frontend.py
lostanlen: I propose this:
(beginner) by default, `implementation=1`
(intermediate) the user can switch to a different implementation by setting `implementation=2`, `3`, etc.
(advanced) the user may pass `implementation` as a dictionary containing `average_fr` and friends.

The main benefit of this is that we don't have parameters overriding other parameters under the hood. Every user-defined choice has an effect.

The second benefit is that it allows a learning curve for JTFS. First without `implementation` (implicitly =1), then with integer-valued presets, then with dict-valued custom settings (useful in debugging and research!).

The third benefit is that my proposal is already partially achieved by @OverLordGoldDragon via `self._implementation_args`, so i don't expect there's a lot to change here.

OverLordGoldDragon:
The changes are either necessary or are improvements. They're included in "minimal" because they are difficult to disentangle without breaking everything. It's possible to minify further, but I've ran out of time; if anyone's willing to go through the trouble, I can review it, but I don't think it's worth the effort. Also I doubt it's "many" breaking changes to Scattering1D.

> regression tests

What?

> timeline for merging the Torch and TensorFlow backends

Fairly quick. They're mainly Numpy-equivalents with their own syntax, so once Numpy's done, we're good. There's a few design choices to be discussed, and improvements to consider for awkward TF code, but as far as backends go, I think the only major piece is `agnostic_backend`.

> doesn't pass the unit tests

Unsure why. There was a trivial error I fixed, but currently getting

image

I'll check back later.

kymatio/scattering1d/filter_bank.py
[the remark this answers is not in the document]
OverLordGoldDragon: User wants `T`. How do we know we're giving it to them? Continuous-time expressions guarantee nothing: we must measure to validate. The `width` meta is used directly in JTFS with `sampling_psi_fr='exclude'`.

`sigma0` is how `T` maps to final `sigma` for `phi`. It's always there, even if it's not there (in which case it's just `1`: `sigma=1/T`).

kymatio/scattering1d/frontend/base_frontend.py
OverLordGoldDragon: It's _max_ pad factor. And it's not absolute input length, but length of filters relative to input. I chose `2` because anything less will _severely_ degrade filters in the worst case of `J=log2_N`: we get pure sines.

Also, note that the whole x8 x16 padding thing isn't as scary as it seems: it _only_ applies for `J=log2_N`. That isn't a typical `J` at all, AFAIK. If `J=log2_N - 4`, we don't need to pad at all.

kymatio/scattering1d/frontend/base_frontend.py
OverLordGoldDragon: > Is it necessary to have these options? Or would it be better to document possible options and use cases with examples?

I don't understand, could you elaborate? We have kwargs for full control, and `implementation` for presets. We need not scrap one for other.

> What if I want a 3D output without average_fr?

`out_3D=True; average_fr=False`? @rastegah

kymatio/scattering1d/frontend/base_frontend.py
OverLordGoldDragon: That's a nice way to clean up the namespace and support more args without worry for clutter, yes. Some problems:

 1. Currently two very basic parameters, `average_fr` and `out_type`, vary by `implementation`: do we really want to hide these?
 2. How would docs change? Some params have long descriptions and I'd really rather not have them all indented and packed under the grand umbrella `implementation`. Perhaps a note somewhere, "these kwargs are accessible via `implementation`"?

> don't have parameters overriding other parameters under the hood.

Agreed that's a benefit. While it's documented, it'd be safer to throw a warning; problem is, we'd create duplication as we'd have to store defaults both in `def` and elsewhere, to be able to compare.

kymatio/scattering1d/core/timefrequency_scattering1d.py
```diff
@@ -0,0 +1,1314 @@
...
+    Frequential scattering
+    ======================
+
+    Variables
+    ---------
+    Explanation of variable naming and roles. For `scf`'s attributes see
+    full Attributes docs in `TimeFrequencyScattering1D`.
+
+    subsample_equiv_due_to_pad
+        Equivalent amount of subsampling due to padding, relative to
```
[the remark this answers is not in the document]
OverLordGoldDragon: Eh, I always found this naming clumsy to first impressions. The docs read:

> Distinguishes differences in input lengths (to `_frequency_scattering()` and `_joint_lowpass()`) due to padding and conv subsampling. This is needed to tell whether input is *trimmed* (pad difference) or *subsampled*.

The idea is, input lengths may differ for two reasons, and for some operations, we need to know exactly for which one. This should be clearer after familiarizing with code. 

For example, `lowpass_subsample_fr` aims for _total_ subsampling to be a certain amount (e.g. `log2_F`). Suppose: input length `32`, target total subsampling `4`. Now, the input could come from: 1) pad to `64`, subsample by `2`; 2) pad to `128`, subsample by `4`. In case 1, we want to subsample further by `2`; in case 2, we're done.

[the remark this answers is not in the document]
OverLordGoldDragon: I'm open to renaming

kymatio/scattering1d/frontend/base_frontend.py
rastegah: an `implementation` argument adds overhead. reading about it is quite confusing. By the time i've figured out what all the presets do, I probably already understand each of the arguments well enough to create my own preset. I don't find it so helpful. 

>     What if I want a 3D output without average_fr?

I say this because its an example of a possible preset that is not available. There are plenty of possible presets. I guess that a given user may only ever opt for two of them. 

Overriding parameters could have consequences due to incorrect interpretation by the user. If I set `implementation=1`, `average_fr=True`, `aligned=True`, `out_3D=True`, what do I get? How can I be sure that I get what I ask for?

Does this _preset_ functionality add confusion? Can you use it before understanding the role of the overridden parameters?

What I am saying is, rather than overriding parameters, documentation would better serve to demonstrate how to use the frontend to achieve an _implementation_ and give examples of _when_ you may want to use a particular implementation.

kymatio/scattering1d/frontend/base_frontend.py
lostanlen: then at least let's add a comment saying what `J_pad` does?

kymatio/scattering1d/filter_bank.py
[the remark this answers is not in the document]
lostanlen: From what you're saying, i understand that `sigma0` should be computed at runtime from the specification of `T` and `phi`. Changing `sigma0` defeats the purpose of adjusting `phi` according to an objective definition of invariance `T=phi.width`. It should not be a keyword argument anymore.

> The width meta is used directly in JTFS with sampling_psi_fr='exclude'

what's the criterion for `exclude`? i imagine something involving `psi2` and `psi_fr`?

kymatio/scattering1d/frontend/base_frontend.py
lostanlen: Thanks for explaining. That makes sense.

You're doing a nice work making sure that everything functions as intended even in the most extreme cases (`J` approaching `log2_N`).

In practice, we will indeed have something like `J = log2_N - 4`, as you've pointed out.

I'll try to remember to explain this in the tutorial

kymatio/scattering1d/frontend/base_frontend.py
lostanlen: 1. yes, frankly i have no problem with hiding `average_fr` and `out_type` behind implementation
2. yes, good idea

From this discussion it seems like we should either put `implementation` to full use (remove other kwargs) or remove `implementation` itself, as proposed by @rastegah. Leaving both will lead to a never-ending stream of silent bugs in user code.

"There should be one-- and preferably only one --obvious way to do it." — Zen of Python


> While it's documented, it'd be safer to throw a warning

Warnings are too easily ignored and should only be used as a last resort. Error messages are almost always preferable.

kymatio/scattering1d/frontend/base_frontend.py
lostanlen: @rastegah well, precisey, it's impossible to do 3D output without average_fr because the frequency indices `n1`  would not be evenly sampled across quefrencies `j_fr`, so we couldn't stack them. Having to explain that in the documentation, and to catch those edge cases in the code, would be difficult, due to the combinatorial explosion of low-level logical settings.

The nice thing about `implementation` is that it sets a nice framework of what's possible vs. what isn't, in the form of a collection of officially supported presets.


> Overriding parameters could have consequences due to incorrect interpretation by the user

I agree that overriding parameters would be a disaster and should not be allowed.


>  If I set implementation=1, average_fr=True, aligned=True, out_3D=True, what do I get? 

Good example. I can't resolve the logic in my head at the glance of an eye, which should be enough to tell us that implementation should not coexist with average_fr and friends. It's one XOR the other 😅 



> What I am saying is, rather than overriding parameters, documentation would better serve to demonstrate how to use the frontend to achieve an implementation and give examples of when you may want to use a particular implementation.

Absolutely. This JTFS needs to be documented via a tutorial and illustrating examples, not just docstrings. @OverLordGoldDragon has already done a fantastic work of illustrating examples, and we have people around to help (Changhong, Han, you), so i'm optimistic about that.

kymatio/scattering1d/filter_bank.py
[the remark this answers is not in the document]
lostanlen: @OverLordGoldDragon i still don't get why the `max` with 0.4 though. It seems like the value is `1 / (1 + 2**(1/Q))` if Q>1 and `0.4` otherwise, whereas the formula would have given `0.333`.

so the logic is:
* if Q>1, `2**(1/Q) xi = 1 - xi`
* if Q=1, `2**(1/Q) xi = 1 - xi/2`

Why does `Q=1` gets a special treatment?

kymatio/scattering1d/filter_bank.py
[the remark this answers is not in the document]
lostanlen: i see. Long-term this kwarg can be completely removed and become a global constant or something like that. It's upsetting that it appears in more than one place, but i can live with that.

lostanlen:
@OverLordGoldDragon excellent news! The last things on the to-do list i'm seeing are:

- [ ]  remove user-specific sigma0 from the definition of phi. phi is purely defined from T now.

- [ ]  sort out the discussion on new `xi_max`. it's not clear to me why we're giving `Q=1` a special treatment. It is a heuristic?

- [ ] stick to `analytic=False` by default. I am not fully aware of the implications of `analytic=True` on time-domain localization.

- [ ] sort out the discussion on `implementation`. make sure keyword arguments don't override each other.

- [ ] get the unit tests in order

- [ ] merge!


> Fairly quick. They're mainly Numpy-equivalents with their own syntax, so once Numpy's done, we're good. There's a few design choices to be discussed, and improvements to consider for awkward TF code, but as far as backends go, I think the only major piece is agnostic_backend.

OK this is great. We can team up with @rastegah et al. for this.

kymatio/scattering1d/frontend/base_frontend.py
OverLordGoldDragon: An argument docstring is meant to document the argument, not explain the entire algorithm. Of course reading `implementation` alone will leave one clueless.

I agree for a need for an intro, with 

>  examples of when you may want to use a particular implementation

and it could be referred to in `implementation`.

The idea behind current `implementation` is simple: if you use it, you will get parameters specified in it no matter what. If want full control, then set it to `None` and pass your own stuff. Regardless I agree there's awkwardness in design, one solved by @lostanlen's suggestion to move it all into `dict` while preserving `int` option - main downside being that arguments are no longer exposed directly in the function's namespace, and one must read docs.

I favor moving the args into `implementation` per the `dict` + `int` design.

kymatio/scattering1d/frontend/base_frontend.py
OverLordGoldDragon: Why specifically for `J_pad`? It's covered in attribute docstrings along all other attributes

kymatio/scattering1d/frontend/base_frontend.py
lostanlen: @rastegah would you be on board with this? `implementation` is either `int` or `dict`

kymatio/scattering1d/frontend/base_frontend.py
rastegah: i've realized that the `J_pad` logic in the frontend was not enttirely introduced in this commit. 

although it makes it harder to use `scattering_filter_factory` (for demo and visualization purposes), fixing it can be left for another time ...

kymatio/scattering1d/frontend/base_frontend.py
rastegah: > @rastegah would you be on board with this? `implementation` is either `int` or `dict`

this seems like the best solution in the current state. that way, `implementation` itself must be overriden.

let's make sure that the overriden parameters are not `kwargs`

rastegah:
> @OverLordGoldDragon excellent news! The last things on the to-do list i'm seeing are:
> 
>     * [ ]   remove user-specific sigma0 from the definition of phi. phi is purely defined from T now.
> 
>     * [ ]   sort out the discussion on new `xi_max`. it's not clear to me why we're giving `Q=1` a special treatment. It is a heuristic?
> 
>     * [ ]  stick to `analytic=False` by default. I am not fully aware of the implications of `analytic=True` on time-domain localization.
> 
>     * [ ]  sort out the discussion on `implementation`. make sure keyword arguments don't override each other.
> 
>     * [ ]  get the unit tests in order
> 
>     * [ ]  merge!
> 
> 
> > Fairly quick. They're mainly Numpy-equivalents with their own syntax, so once Numpy's done, we're good. There's a few design choices to be discussed, and improvements to consider for awkward TF code, but as far as backends go, I think the only major piece is agnostic_backend.
> 
> OK this is great. We can team up with @rastegah et al. for this.

I think we have concluded on a solution to `implementation` and `J_pad`. let's address these points and merge. then we can handle the `torch` and `tf` frontends.

great work @OverLordGoldDragon !

kymatio/scattering1d/utils.py
```diff
@@ -114,23 +127,73 @@ def compute_minimum_support_to_pad(T, J, Q, criterion_amplitude=1e-3,
...
     """
-    J_tentative = int(np.ceil(np.log2(T)))
-    _, _, _, t_max_phi = scattering_filter_factory(
-        J_tentative, J, Q, normalize=normalize, to_torch=False,
-        max_subsampling=0, criterion_amplitude=criterion_amplitude,
-        r_psi=r_psi, sigma0=sigma0, alpha=alpha, P_max=P_max, eps=eps)
-    min_to_pad = 3 * t_max_phi
-    return min_to_pad
+    # compute params for calibrating, & calibrate
+    J_tentative = int(np.ceil(np.log2(N)))
```
rastegah, line 140: `J_tentative` unused

kymatio/scattering1d/filter_bank.py
```diff
@@ -284,33 +292,92 @@ def compute_temporal_support(h_f, criterion_amplitude=1e-3):
...
         N = half_support
-        # Raise a warning to say that there will be border effects
-        warnings.warn('Signal support is too small to avoid border effects')
+        if warn:
+            # Raise a warning to say that there will be border effects
+            warnings.warn('Signal support is too small to avoid border effects')
+    return N
+
+
+def compute_minimum_required_length(fn, N_init, max_N=None,
```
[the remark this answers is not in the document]
OverLordGoldDragon: They serve to compute `min_to_pad`, i.e. minimum amount of required padding to avoid boundary effects. 

> just do phi_halfwidth = compute_temporal_support(N)

doesn't achieve this

> just exclude those paths

That's a choice we can make but is a separate consideration. The purpose of this compute block is to tell us the padding we'll need _given_ a configuration.

kymatio/scattering1d/filter_bank.py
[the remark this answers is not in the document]
OverLordGoldDragon: Because `0.333` is too low, we want coverage near Nyquist. Full curve for reference

kymatio/scattering1d/filter_bank.py
```diff
@@ -426,53 +493,66 @@ def compute_params_filterbank(sigma_low, Q, r_psi=math.sqrt(0.5), alpha=5.):
...
-        (the larger r_psi, the larger the overlap between adjacent wavelets).
+        (the larger r_psi, the larger the overlap between adjacent wavelets),
+        and stability against time-warp deformations (larger r_psi improves it).
         Defaults to sqrt(0.5).
     alpha : float, optional
         tolerance factor for the aliasing after subsampling.
         The larger alpha, the more conservative the value of maximal
-        subsampling is. Defaults to 5.
+        subsampling is. Defaults to 4.
+    xi_min : float, optional
```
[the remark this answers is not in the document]
OverLordGoldDragon: It's just an additional heuristic to ensure a well-behaved filterbank, like tiling the lowest portion with STFT. There's no need to burden the user with having to lower `J`, and requiring greater input length is far too intrusive ("just get more data"?)

> shouldn't pad more

That's impossible with `J=log2_N`. We either pad more or pay the price in boundary effects and ill-behaved wavelets.

kymatio/scattering1d/frontend/base_frontend.py
OverLordGoldDragon: Not a fan of moving compute to a function unless necessary - in this case it's just one line.

> harder to use scattering_filter_factory

How so? It's just `scattering_filter_factory(log2(N), J, Q, T)` - we pass `J_pad` in frontend because that's our "N" for filters. It's also an internal method anyway, and filter length would instead be controlled by `max_pad_factor`

Regardless yes, "optional" refactors best wait

kymatio/scattering1d/filter_bank.py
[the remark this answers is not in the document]
OverLordGoldDragon:
```python
if scf.sampling_psi_fr == 'exclude':
    width = psi1_f[n1_fr]['width'][subsample_equiv_due_to_pad]
    if width > scf.N_frs[n2]:
    	continue
```
If `psi_fr` is too wide relative to input -> skip. You have something similar in `scattering.m`

kymatio/scattering1d/filter_bank.py
```diff
@@ -48,16 +48,20 @@ def adaptive_choice_P(sigma, eps=1e-7):
...
     """
     Computes a periodization of a filter provided in the Fourier domain.
-
     Parameters
     ----------
     h_f : array_like
         complex numpy array of shape (N*n_periods,)
     n_periods: int, optional
         Number of periods which should be used to periodize
+    aggregation: str['sum', 'mean'], optional
```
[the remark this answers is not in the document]
OverLordGoldDragon: Currently it's unused and can be removed, yes - can reintroduce if there's ever a use case. Still would need to document why we're summing

Also I was never a fan of this name. "Periodizing", in full, is sampling the continous-time function over multiple periods, which is done in `morlet_1d` and `gabor_1d`, _and_ subsampling. This function _only_ does latter, and I originally found it very confusing. It should either be called `subsample_filter_fourier` or `fold_filter_fourier`, and I'd recommend referencing [this post](https://dsp.stackexchange.com/q/73011/50076)

kymatio/scattering1d/frontend/base_frontend.py
OverLordGoldDragon: > it's just one line

though fair enough if moving all that other stuff, but would need to `return` these

kymatio/scattering1d/filter_bank.py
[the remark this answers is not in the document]
OverLordGoldDragon: Add to todo?

OverLordGoldDragon:
mkay powered through some posts

Dunno what's up with unit tests, especially 2d and 3d which work on `dev` but don't here, despite this branch deriving from `dev` and not modifying 2d or 3d. No time to dig, but if they call for changes to 1d code, I can review them

Good luck!

tests/scattering1d/test_correctness.py
```diff
@@ -0,0 +1,222 @@
...
+import pytest
+import numpy as np
+import scipy.signal
+from kymatio import Scattering1D
+from kymatio.scattering1d.backend.agnostic_backend import (
+    pad, stride_axis, unpad_dyadic)
+from kymatio.toolkit import rel_l2
+
+# set True to execute all test functions without pytest
+run_without_pytest = 0
```
rastegah, line 12: this is dangerous and won't make it to main

tests/scattering1d/test_jtfs.py
```diff
@@ -0,0 +1,372 @@
...
+def assert_pad_difference(jtfs, test_params_str):
+    assert not all(
+        J_pad_fr == jtfs.J_pad_frs_max
+        for J_pad_fr in jtfs.J_pad_frs if J_pad_fr != -1
+        ), "\n{}\nJ_pad_fr={}\nN_frs={}".format(
+            test_params_str, jtfs.J_pad_frs, jtfs.N_frs)
+
+
+if __name__ == '__main__':
+    if run_without_pytest:
```
rastegah, line 360: what if you accidentally omit a test in this block? that means no coverage

This does not have to be commited since it's a developer feature. 

It is not much effort to add for your personal convenience.

tests/scattering1d/test_numpy_scattering1d.py
```diff
@@ -32,4 +32,5 @@ def test_Scattering1D(self, backend):
         scattering = Scattering1D(J, T, Q, backend=backend, frontend='numpy')
 
         Sx = scattering(x)
-        assert np.allclose(Sx, Sx0)
+        print("WARNING: test skipped")  # TODO
```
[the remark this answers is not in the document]
rastegah: any update here?

rastegah (changes requested):

tests/scattering1d/test_jtfs.py
OverLordGoldDragon: > what if you accidentally omit a test in this block?

`run_without_pytest == False` when committed, always. This can be refined further to override in case someone commits `True`, by [detecting](https://stackoverflow.com/q/25188119/10133797) the runtime environment, or by setting one global value in `tests/__init__.py`, which I'm willing to do - but saw no need for just two files.

Also Github notified me of none of this, helps to @ @rastegah

kymatio/scattering1d/utils.py
```diff
@@ -114,23 +127,73 @@ def compute_minimum_support_to_pad(T, J, Q, criterion_amplitude=1e-3,
...
+    J_support = J_tentative
```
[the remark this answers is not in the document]
OverLordGoldDragon: [dupe](https://github.com/kymatio/kymatio/pull/781#discussion_r791879776)

tests/scattering1d/test_numpy_scattering1d.py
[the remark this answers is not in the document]
OverLordGoldDragon: Can change to whatever #673 does, not much else to be done here without the test data code

kymatio/scattering1d/core/timefrequency_scattering1d.py
```diff
@@ -0,0 +1,1314 @@
...
+        pad_ref = (scf.J_pad_frs_min if aligned else
+                   scf.J_pad_frs_max)
+        subsample_equiv_due_to_pad_ref = scf.J_pad_frs_max_init - pad_ref
+        # ensure stride is zero if `not average and aligned`
+        average_fr = bool(scf.average_fr or not aligned)
+        stride_ref = _get_stride(
+            None, pad_ref, subsample_equiv_due_to_pad_ref, scf, average_fr)
+        stride_ref = max(stride_ref - oversampling_fr, 0)
+        ind_start_fr = scf.ind_start_fr_max[stride_ref]
+        ind_end_fr   = scf.ind_end_fr_max[  stride_ref]
```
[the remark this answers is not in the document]
MuawizChaudhary: Not a fan.

kymatio/scattering1d/core/timefrequency_scattering1d.py
```diff
@@ -0,0 +1,1314 @@
...
+        else:
+            total_conv_stride_over_U1 = phi_fr['j'][k]
+
+    n1_fr_subsample = min(j1_fr, total_conv_stride_over_U1)
+    lowpass_subsample_fr = total_conv_stride_over_U1 - n1_fr_subsample
+    total_downsample_fr = n1_fr_subsample + lowpass_subsample_fr + k
+    ```
+    """
+    # pack for later
+    B = backend
```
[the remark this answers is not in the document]
MuawizChaudhary: @lostanlen  @OverLordGoldDragon  I believe we should reconsider. We should be keeping in line with how we have programmed the rest of the kymatio software; we cache the functions. [Furthermore, attribute access is significantly slower than chaching the attribute. ](https://betterprogramming.pub/optimize-your-python-codebases-by-2-5x-with-direct-attribute-access-410df06ec46e). [Here is a stack overflow post going into more detail. ](https://stackoverflow.com/questions/28597014/python-why-is-accessing-instance-attribute-slower-than-local)

kymatio/scattering1d/core/timefrequency_scattering1d.py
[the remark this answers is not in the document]
lostanlen: > For example, lowpass_subsample_fr aims for total subsampling to be a certain amount (e.g. log2_F). Suppose: input length 32, target total subsampling 4. Now, the input could come from: 1) pad to 64, subsample by 2; 2) pad to 128, subsample by 4. In case 1, we want to subsample further by 2; in case 2, we're done.

What you describe sounds like `k1_J` in the 1D code. This logic is not affected by padding/trimming.

I don't see how padding interacts with subsampling (unless you're using array shapes to guess these numbers?)

kymatio/scattering1d/core/timefrequency_scattering1d.py
```diff
@@ -0,0 +1,1314 @@
...
+        See above. Alt name: `total_conv_stride_fr`; `over_U1` seeks
+        to emphasize that it's the absolute stride over first order coefficients
+        that matters rather than equivalent/relative quantities w.r.t. padded etc
+
+    n1_fr_subsample
+        Amount of conv subsampling done in `_frequency_scattering()`.
+
+    total_downsample_fr
+        Total amount of subsampling, conv and equivalent, relative to
+        `J_pad_frs_max_init`. Conceptual quantity.
```
lostanlen, line 47: i'm sorry but this documentation is unreadable. You write a manual to define "conceptual quantities" but you assume that we know the definition of `J_pad_frs_max_init`

kymatio/scattering1d/core/timefrequency_scattering1d.py
```diff
@@ -0,0 +1,1314 @@
...
+
+    "conv subsampling"
+        Refers to `subsample_fourier()` after convolution (as opposed to
+        equivalent subsampling due to padding). i.e. "actual" subsampling.
+        Refers to `total_conv_stride_over_U1`.
+
+    total_conv_stride_over_U1
+        See above. Alt name: `total_conv_stride_fr`; `over_U1` seeks
+        to emphasize that it's the absolute stride over first order coefficients
+        that matters rather than equivalent/relative quantities w.r.t. padded etc
```
lostanlen, line 40: > the absolute stride over first order coefficients that matters rather than equivalent/relative quantities w.r.t. padded

you're making it sound like stride depends upon padding? is this another `aligned=False` trick?

Issue #801, opened 2022-03-24
AttributeError: 'Tensor' object has no attribute 'ndim'


ryankshah:
I was following the classification tutorial for 1D wavelet scattering and I receive the error:

```python
Traceback (most recent call last):
  File "filter_signals_fft.py", line 145, in <module>
    x = Scattering1D(J, Q=Q)(x_in)
  File "C:\Users\xwb18152\AppData\Roaming\Python\Python38\site-packages\keras\utils\traceback_utils.py", line 67, in error_handler
    raise e.with_traceback(filtered_tb) from None
  File "C:\Users\xwb18152\AppData\Roaming\Python\Python38\site-packages\tensorflow\python\autograph\impl\api.py", line 692, in wrapper
    raise e.ag_error_metadata.to_exception(e)
AttributeError: Exception encountered when calling layer "scattering1d" (type Scattering1D).        

in user code:

    File "C:\Users\xwb18152\AppData\Roaming\Python\Python38\site-packages\kymatio\frontend\keras_frontend.py", line 17, in call  *
        return self.scattering(x)
    File "C:\Users\xwb18152\AppData\Roaming\Python\Python38\site-packages\kymatio\frontend\keras_frontend.py", line 14, in scattering  *
        return self.S.scattering(x)
    File "C:\Users\xwb18152\AppData\Roaming\Python\Python38\site-packages\kymatio\scattering1d\frontend\tensorflow_frontend.py", line 53, in scattering  *
        S = scattering1d(x, self.pad_fn, self.backend.unpad, self.backend, self.J, self.log2_T, self.psi1_f, self.psi2_f,
    File "C:\Users\xwb18152\AppData\Roaming\Python\Python38\site-packages\kymatio\scattering1d\core\scattering1d.py", line 76, in scattering1d  *
        U_0 = pad_fn(x)
    File "C:\Users\xwb18152\AppData\Roaming\Python\Python38\site-packages\kymatio\scattering1d\frontend\base_frontend.py", line 80, in pad_fn  *
        self.pad_mode)
    File "C:\Users\xwb18152\AppData\Roaming\Python\Python38\site-packages\kymatio\scattering1d\backend\tensorflow_backend.py", line 71, in pad  *
        return agnostic.pad(x, pad_left, pad_right, pad_mode, axis=axis)
    File "C:\Users\xwb18152\AppData\Roaming\Python\Python38\site-packages\kymatio\scattering1d\backend\agnostic_backend.py", line 36, in pad  *
[...]

Issue #775, opened 2021-08-31
Keras Scattering1D example does not work with TF>=2.5


entn-at:
When using Scattering1D with the Keras frontend (for example, in the spoken digit classification example,  `classif_keras.py`) of
 Kymatio 0.2.1 (installed from PyPi via pip) with TF 2.5 or 2.6, I'm getting the following error:
```ValueError: Error processing property 'phi_f' of DictWrapper(...```
TF2.4 works flawlessly. I'm not sure what change is causing this issue.

Minimal example for reproduction:
```
python3 -m venv ./venv
source venv/bin/activate
pip install tensorflow==2.6.0 kymatio=0.2.1
wget https://github.com/kymatio/kymatio/raw/master/examples/1d/classif_keras.py
python classif_keras.py
```

The full error log of running the `classif_keras.py` example script is below:
```
$ python classif_keras.py
2021-08-31 13:55:02.788089: W tensorflow/stream_executor/platform/default/dso_loader.cc:64] Could not load dynamic library 'libcudart.so.11.0'; dlerror: libcudart.so.11.0: cannot open shared object file: No such file or directory
2021-08-31 13:55:02.788144: I tensorflow/stream_executor/cuda/cudart_stub.cc:29] Ignore above cudart dlerror if you do not have a GPU set up on your machine.
2021-08-31 13:55:15.781264: W tensorflow/stream_executor/platform/default/dso_loader.cc:64] Could not load dynamic library 'libcuda.so.1'; dlerror: libcuda.so.1: cannot open shared object file: No such file or directory
2021-08-31 13:55:15.781343: W tensorflow/stream_executor/cuda/cuda_driver.cc:269] failed call to cuInit: UNKNOWN ERROR (303)
2021-08-31 13:55:15.781364: I tensorflow/stream_executor/cuda/cuda_diagnostics.cc:156] kernel driver does not appear to be running on this host (eduxps): /proc/driver/nvidia/version does not exist
2021-08-31 13:55:15.781694: I tensorflow/core/platform/cpu_feature_guard.cc:142] This TensorFlow binary is optimized with oneAPI Deep Neural Network Library (oneDNN) to use the following CPU instructions in performance-critical operations:  AVX2 AVX512F FMA
To enable them in other operations, rebuild TensorFlow with the appropriate compiler flags.
Model: "model"
_________________________________________________________________
Layer (type)                 Output Shape              Param #
=================================================================
input_1 (InputLayer)         [(None, 8192)]            0
[...]

PR #803, opened 2022-03-28
Fix issue #801 and Merge #775

ryankshah:
Somehow sorting the issue for scattering1d by modifying scattering3D backend (ndim resolution) and merge for #775. However, the code runs significantly slower and does not match run times as presented in documented examples.


MuawizChaudhary:
Closed by #882. Thanks for reporting the issue to us.

PR #814, opened 2022-05-20
Removed redundant default value for alpha

MarcusHaggbom:
Backwards compatible
No new functionality
Note! Test function was modified to use a value of alpha=5.

Removed default value for alpha parameter everywhere in filter_bank.py except factory function (scattering_filter_factory). This way the value of alpha is set once, avoiding a possible override of this value by the default.

### DIFF
--- a/kymatio/scattering1d/filter_bank.py
+++ b/kymatio/scattering1d/filter_bank.py
@@ -310,7 +310,7 @@ def compute_temporal_support(h_f, criterion_amplitude=1e-3):
     return N
 
 
-def get_max_dyadic_subsampling(xi, sigma, alpha=5.):
+def get_max_dyadic_subsampling(xi, sigma, alpha):
     """
     Computes the maximal dyadic subsampling which is possible for a Gabor
     filter of frequency xi and width sigma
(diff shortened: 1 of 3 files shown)


lostanlen:
@janden how can i restart this CI so that we retry the build on Jenkins?
and how do i run "self-hosted CI" ?

janden:
I have rebased this on top of the new dev, so the CI should be working again.

janden:
@MarcusHaggbom Remind me again what `alpha` does and why it's redundant?

MarcusHaggbom:
> @MarcusHaggbom Remind me again what `alpha` does and why it's redundant?

Alpha works as a parameter on how much aliasing error is tolerated when subsampling, and hence affects the number of "allowed" dyadic subsamplings for a wavelet.
The redundancy comes from assigning a default value to alpha in multiple places; when querying for filters given a certain value for alpha, this value should be used throughout the module. By requiring it to be passed explicitly in later functions, we mitigate the risk of (in future versions) forgetting to pass the intended value, which otherwise could cause a silent error. I.e.:
```
def main(alpha=1):
    _sub1(alpha)

def _sub1(alpha):
    _sub2()  # Forgetting to pass alpha

def _sub2(alpha=1):
    print(alpha)

main(0)
```

lostanlen:
ready to merge

lostanlen:
@MuawizChaudhary merge ?

Issue #760, opened 2021-07-21
docs Scikit-learn examples not working


chepo92:
From https://www.kymat.io/codereference.html#module-kymatio.sklearn
I run ` from kymatio.sklearn import Scattering1D, Scattering2D, HarmonicScattering3D`
And then run code example, eg 
```
# Set the parameters of the scattering transform.
J = 3
M, N, O = 32, 32, 32

# Generate a sample signal.
x = np.random.randn(np.prod(M, N, O))

# Define a HarmonicScattering3D object.
S = HarmonicScattering3D(J, (M, N, O))

# Calculate the scattering transform.
Sx = S.scattering(x)

# Equivalently, use the alias.
Sx = S.predict(x)
```
none of the examples work, errors throwed: 
3D:
```
TypeError                                 Traceback (most recent call last)

<ipython-input-65-7d3625924daf> in <module>()
      4 
      5 # Generate a sample signal.
----> 6 x = np.random.randn(np.prod(M, N, O))
      7 
      8 # Define a HarmonicScattering3D object.

<__array_function__ internals> in prod(*args, **kwargs)

1 frames

/usr/local/lib/python3.7/dist-packages/numpy/core/fromnumeric.py in _wrapreduction(obj, ufunc, method, axis, dtype, out, **kwargs)
     85                 return reduction(axis=axis, out=out, **passkwargs)
     86 
---> 87     return ufunc.reduce(obj, axis, dtype, out, **passkwargs)
     88 
     89 

TypeError: Cannot interpret '32' as a data type
```
1D:
```
TypeError                                 Traceback (most recent call last)

<ipython-input-66-79e7ca25bbde> in <module>()
     14 
     15 # Equivalently, use the alias.
---> 16 Sx = S.predict(x)

/usr/local/lib/python3.7/dist-packages/kymatio/frontend/sklearn_frontend.py in predict(self, x)
      9     def predict(self, x):
     10         n_samples = x.shape[0]
---> 11         x = x.reshape((-1,) + self.shape)
     12         Sx = self.scattering(x)
     13         Sx = Sx.reshape(n_samples, -1)

TypeError: can only concatenate tuple (not "int") to tuple

```

2D: 

```
AxisError                                 Traceback (most recent call last)

<ipython-input-64-d0dc2efe80b0> in <module>()
      4 
      5 # Generate a sample signal.
----> 6 x = np.random.randn(np.prod(M, N))
      7 
      8 # Define a Scattering2D object.

[...]

PR #823, opened 2022-05-31
DOC correction

MuawizChaudhary:
This pull request corrects an incorrect docstring in the sklearn documentation. I remove the use of `np.prod` in the creation of the input, and additionally add a batch dimension. 

While the documentation now correctly shows how to apply it to an input of a certain shape, the sklearn documentation example is obviously not that great.

I don't know how to build the documentation. can someone send me a tutorial on how to build it? 

This solves issue #760

### DIFF
--- a/kymatio/frontend/keras_frontend.py
+++ b/kymatio/frontend/keras_frontend.py
@@ -21,7 +21,7 @@ class ScatteringKeras(Layer):
 
     _doc_alias_name = 'call'
 
-    _doc_alias_call = ''
+    _doc_alias_call = '({x})'
 
     _doc_frontend_paragraph = \
         """
--- a/kymatio/frontend/numpy_frontend.py
+++ b/kymatio/frontend/numpy_frontend.py
@@ -14,7 +14,7 @@ class ScatteringNumPy:
 
     _doc_alias_name = '__call__'
 
-    _doc_alias_call = ''
+    _doc_alias_call = '({x})'
 
     _doc_frontend_paragraph = ''
 
--- a/kymatio/frontend/sklearn_frontend.py
+++ b/kymatio/frontend/sklearn_frontend.py
@@ -21,7 +21,7 @@ class ScatteringTransformerMixin(BaseEstimator, TransformerMixin):
 
     _doc_alias_name = 'predict'
 
-    _doc_alias_call = '.predict'
+    _doc_alias_call = '.predict({x}.flatten())'
 
     _doc_frontend_paragraph = \
         """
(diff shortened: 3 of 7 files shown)


janden:
Doesn't scikit-learn assume that all inputs are one-dimensional vectors (not arrays). I think that's why we have `np.prod` here.

> I don't know how to build the documentation. can someone send me a tutorial on how to build it?

Try `cd doc && make html`.

MuawizChaudhary:
> Doesn't scikit-learn assume that all inputs are one-dimensional vectors (not arrays). I think that's why we have `np.prod` here.
> 
> > I don't know how to build the documentation. can someone send me a tutorial on how to build it?
> 
> Try `cd doc && make html`.

It assumes that there is a batch dimension too. the original example wasn't taking this into account

janden:
Ok so I think the issue with the example is that we shouldn't be calling `S.scattering`, just `S.predict`. The latter expects the vector input while the former expects a 1D/2D/3D array. Will see if there's a good way to fix this.

MuawizChaudhary:
> Ok so I think the issue with the example is that we shouldn't be calling `S.scattering`, just `S.predict`. The latter expects the vector input while the former expects a 1D/2D/3D array. Will see if there's a good way to fix this.

Yeah, [you](https://colab.research.google.com/drive/1QtXZBJ-_Omso8mTbRzlifK0ziCzBengK#scrollTo=Kg3rLC3qumoV) are [right](https://github.com/kymatio/kymatio/blob/dev/examples/2d/plot_sklearn.py#L63). Personally, I think it's fine, our scattering in the sklearn framework [implements the predict function](https://github.com/kymatio/kymatio/blob/master/kymatio/frontend/sklearn_frontend.py#L9) which is what users should be using when they're doing sklearn stuff. Perhaps the `S(x)` call should call `predict` instead of `scattering`.

MuawizChaudhary:
> So shouldn't it be `(1, np.prod(({shape}))`, then?

Should we also change the `1` to a `B` for batch size?

janden:
> Perhaps the `S(x)` call should call `predict` instead of `scattering`.

I'm not following here. I think it would be better to keep the same semantics for `__call__` (i.e., a non-flattened array) and reserve the sklearn-specific semantics for `predict` (i.e., a flattened array). The thing is this would require changing the example structure quite a bit.

> Should we also change the `1` to a `B` for batch size?

Not sure. We don't do it in any of the other frontends, do we? In fact, we don't have a batch dimension at all (which is why we didn't here to begin with, I assume).

janden:
Think I've managed to fix this. We now specify the flattened version when calling `predict` and otherwise use the standard (array) version when calling `scattering`. I also fixed the underlying bug that occurs when calling `predict` without a batch dimension.

Issue #817, opened 2022-05-30
Examples fail


edouardoyallon:
Hi, I'm just noticing that many examples of the repo (e.g. https://www.kymat.io/gallery_2d/plot_invert_scattering_torch.html#sphx-glr-gallery-2d-plot-invert-scattering-torch-py which doesn't recover the original image or the scattering disk which makes no sens given the image&scattering features https://www.kymat.io/gallery_2d/plot_scattering_disk.html#sphx-glr-gallery-2d-plot-scattering-disk-py ) fail to achieve what they claim to do - I think it's a bit crucial to fix, as these are literally the showcase of the package, which is thus not really convincing, or even sounds wrong.

MuawizChaudhary:
@edouardoyallon @anakin-datawalker 

I'm not sure what is the fix to to scattering disk. do you have an image of what is supposed to be shown there/whats the correct/expected output? 

It looks to be pretty similar to Joan Bruna's figure from their seminal Invariant Scattering Networks paper. 

![image](https://user-images.githubusercontent.com/39755015/171216834-d8666d57-284a-449e-b300-c2283c5614d0.png)

![image](https://user-images.githubusercontent.com/39755015/171217732-f43cf1fc-b4c1-4102-b50f-b93a2ac3560c.png)

MuawizChaudhary:
@edouardoyallon  @anakin-datawalker  This seems hard. It seems like the ordering might be different than what's shown in Bruna's paper. The oscillations are a bit confusing. Is the ordering of our coefficents different than Bruna's original presentation? 

[Here is my notebook.](https://colab.research.google.com/drive/1jacon162Xtg0PKapC07AGAPbCMupaXG7?usp=sharing) To get the angles to better match, I simply negate the angle. 

Here I plot the first order disk. The first image is my regeneration of the disk for first order, second image is Bruna's disk from the paper. I was able to get most of the angles to match, but there are differences. 
![image](https://user-images.githubusercontent.com/39755015/171493828-9615619b-62f1-48a9-a1a7-31a6c0ffbb15.png)
![image](https://user-images.githubusercontent.com/39755015/171493855-494284a1-2334-47b4-8caa-5d4e293de66e.png)

Here I plot the second order disk. The first image is my regeneration of the disk for second order, second image is Bruna's disk from the paper. I'm wondering if the spacing is too close. 

[...]

PR #829, opened 2022-06-02
DOC Fix 3d example failing import

MuawizChaudhary:
The 3D example on dev fails an import due to the fact that we now use a class structure for the backend. This PR fixes this by importing instead the backend and calling compute integrals. 

[Currently in the process of running the full example on a google colab.](https://colab.research.google.com/drive/1FzK481GSptZijQrmjRuJV9GNJcdtgyku?usp=sharing) 

This goes a small way of solving #817 .


edouardoyallon (approved):
looks good

janden (approved):
Thanks!

MuawizChaudhary:
Mind if we merge?

PR #830, opened 2022-06-02
MAINT removed useless apply function

MuawizChaudhary:
Back in the good old days, before I existed, [Kymatio reimplemented various calls](https://github.com/kymatio/kymatio/commit/12b082fca0e3fdc88bee3220c9e867a29fee5d11#diff-144c72d86c0312177446796e4be3836116b58a5e9169cdea6254a24ab3243134R236) that a `nn.module` had. These included `.cpu`,`cuda`, and were often implemented using a `_apply` call.

A reviewer of our JMLR submission kindly noted that we could just inherit from a `nn.module` instead of reimplementing all those calls. 

This `_apply_filters` tucked and hidden away in the 3d utils code is the last remaining vestige of this old way of doing things.

### DIFF
--- a/kymatio/scattering3d/utils.py
+++ b/kymatio/scattering3d/utils.py
@@ -93,17 +93,3 @@ def sqrt(x):
     np.seterr(**old_settings)
 
     return y
-
-def _apply_filters(filters, fn):
-    """
-        Parameters
-        ----------
-        filters: a filter bank
-        fn: a function to apply on the parameters
-        Returns
-        -------
-        filters: the filters modified
-    """
-    for k in range(len(filters)):
-        filters[k] = fn(filters[k])
-    return filters


lostanlen:
```
Can't use 'tar -xzf' extract archive file: /home/runner/work/_actions/_temp_735093b6-e20a-47f8-9c61-589d4c653a4c/fb47e774-6995-4eb3-8ca0-770e1e7792dc.tar.gz. return code: 2.
```

This `build-pip` CI looks very brittle

edouardoyallon:
Be extra careful. The reason it was done this way is because we wanted an access to the filters via psi[j] and that the filters were stored as a parameter. Unfortunatly, when using .cuda(), the pointer was not going through the correct direction.

MuawizChaudhary:
> Be extra careful. The reason it was done this way is because we wanted an access to the filters via psi[j] and that the filters were stored as a parameter. Unfortunatly, when using .cuda(), the pointer was not going through the correct direction.

hm, wasnt this because we weren't inheriting nn.module? 

A git grep for "_apply_filters" shows that this function is not used anywhere.

edouardoyallon:
no clue, i'm just saying it was there for this reason

MuawizChaudhary:
Can we merge?

lostanlen (approved):

lostanlen:
i'll merge in 24 hours unless there are objections

PR #834, opened 2022-06-04
DOC Invert scattering increase quality of image

MuawizChaudhary:
This pull request increases the quality of the image produced by the sphinx documentation.

some results and wall clock timings using the `time` console command; 
Prior: 
```Epoch 14, loss: 0.0010976835619658232
PSNR: 14.85dB for order 1 and J=2
real	0m10.457s
user	0m16.764s
sys	0m0.836s
```

Figure_2_bad

Now: 
```
Epoch 49, loss: 1.9380240701138973e-05
PSNR: 22.97dB for order 1 and J=2
real	0m14.745s
user	0m33.890s
sys	0m0.772s
```
Figure_racoon_good

This goes a small way of solving https://github.com/kymatio/kymatio/issues/817 .


janden:
I don't quite understand the point of this example. At `J = 2`, we have virtually no averaging and the transform is approximately invertible. What is it that we want to show and why?

MuawizChaudhary:
@edouardoyallon did not like how pixelated the image was.

janden:
> @edouardoyallon did not like how pixelated the image was.
> 
> I'm inclined to agree that it isn't a good look for the package if our image reconstruction is pixelated with noise.

I agree that the current reconstruction is crap. My question was rather why we should have this example at all. It doesn't make any sense to me. Scattering transforms are not (and should not be) used to encode images at high accuracy.

MuawizChaudhary:
I agree with you with that statement. That being said, the example replicates the [result of section 3.4](https://arxiv.org/pdf/1809.06367.pdf) in Scattering Networks for Hybrid Representation
Learning. Its of some scientific validity, so I'm pro keeping it in the package. 

Thoughts @edouardoyallon @eugenium ?

eugenium:
approximating the inverse was a use case that came up a few times (with several people asking how to do this with pyscatwave and later kymatio) e.g. for doing Scat GAN's and the like, it's in several papers. I vote for keeping it.

janden:
Ok then it would be good for this to be written in the example. On the face of it, it sounds like a strange thing to do so it deserves some explanation.

janden (approved):

PR #837, opened 2022-06-05
remove `size_scattering` kwarg from `scattering1d`

lostanlen:
Fixes #836

One kwarg less from `scattering1d` in base_frontend.py`
Fewer LOCs, fewer repetitions

PR #839, opened 2022-06-05
`self.output_size()` (always `detail`-ed), introduce `self.num_coefficients`

### LINKED ISSUE #838: `precompute_size_scattering` should always return a tuple (`detail=False`)
Came up in conversation with @MuawizChaudhary 

The `detail` kwarg in `precompute_size_scattering` isn't really justified:
* it is `True` by default
* it creates type instability (`int` or `tuple`)
* getting the `False` version is trivial: it suffices to call `sum` on the output
* our object-oriented method `self.output_size`, has `detail=False` by default. I suggest removing that kwarg and simply summing across orders after calling `precompute_size_scattering`

Hence i propose to remove it. The only occasion where we need `self.output_size()` in the codebase is in the Keras frontend. 
Perhaps we can set up a helpful error message if people ever call the non-default: `self.output_size(detail=True)`, telling them to call `self.meta()` instead?

lostanlen:
Fixes #838 

* `precompute_size_scattering` now always returns a tuple
* same for `self.output_size()` in `Scattering1D`
* introduce read-only property `self.num_coefficients` to replace the old `self.output_size(detail=False)`
* fewer LOCs for `precompute_size_scattering`

This PR will conflict with #837 in the `if self.vectorize:` statements. #837  should take precedence

### DIFF
--- a/kymatio/scattering1d/frontend/base_frontend.py
+++ b/kymatio/scattering1d/frontend/base_frontend.py
@@ -86,7 +86,6 @@ class ScatteringBase1D(ScatteringBase):
             r_psi=self.r_psi, sigma0=self.sigma0, alpha=self.alpha,
             P_max=self.P_max, eps=self.eps)
 
-    @property
     def meta(self):
         """Get meta information on the transform
 
--- a/kymatio/scattering1d/frontend/base_frontend.py
+++ b/kymatio/scattering1d/frontend/base_frontend.py
@@ -86,6 +86,7 @@ class ScatteringBase1D(ScatteringBase):
             r_psi=self.r_psi, sigma0=self.sigma0, alpha=self.alpha,
             P_max=self.P_max, eps=self.eps)
 
+    @property
     def meta(self):
         """Get meta information on the transform
 
@@ -99,7 +100,7 @@ class ScatteringBase1D(ScatteringBase):
         """
         return compute_meta_scattering(self.J, self.Q, self.T, max_order=self.max_order)
 
-    def output_size(self, detail=False):
+    def output_size(self):
         """Get size of the scattering transform
 
         Calls the static method `precompute_size_scattering()` with the
@@ -113,12 +114,16 @@ class ScatteringBase1D(ScatteringBase):
 
         Returns
         ------
-        size : int or tuple
-            See the documentation for `precompute_size_scattering()`.
+        size : tuple
+            Returns a tuple of size `1+max_order` containing the number of
+            coefficients in each order.
         """
+        return precompute_size_scattering(self.J, self.Q, self.T, self.max_order)
 
-        return precompute_size_scattering(
-            self.J, self.Q, self.T, max_order=self.max_order, detail=detail)
+    @property
+    def num_coefficients(self):
+        """Get total number of scattering coefficients across all orders"""
+        return sum(self.output_size())
 
     _doc_shape = 'N'
 
--- a/kymatio/scattering1d/frontend/base_frontend.py
+++ b/kymatio/scattering1d/frontend/base_frontend.py
@@ -105,12 +105,6 @@ class ScatteringBase1D(ScatteringBase):
         Calls the static method `precompute_size_scattering()` with the
         parameters of the transform object.
 
-        Parameters
-        ----------
-        detail : boolean, optional
-            Specifies whether to provide a detailed size (number of coefficient
-            per order) or an aggregate size (total number of coefficients).
-
         Returns
         ------
         size : tuple
--- a/kymatio/scattering1d/frontend/numpy_frontend.py
+++ b/kymatio/scattering1d/frontend/numpy_frontend.py
@@ -41,14 +41,6 @@ class ScatteringNumPy1D(ScatteringNumPy, ScatteringBase1D):
 
         x = x.reshape((-1, 1) + signal_shape)
 
-        # get the arguments before calling the scattering
-        # treat the arguments
-        if self.vectorize:
-            size_scattering = precompute_size_scattering(
-                self.J, self.Q, self.T, max_order=self.max_order)
-        else:
-            size_scattering = 0
-
         S = scattering1d(x, self.backend.pad, self.backend.unpad, self.backend, self.log2_T, self.psi1_f, self.psi2_f,
                          self.phi_f, max_order=self.max_order, average=self.average, pad_left=self.pad_left,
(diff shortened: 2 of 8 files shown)


kymatio/scattering1d/frontend/base_frontend.py
```diff
@@ -113,12 +113,16 @@ def output_size(self, detail=False):
...
-            See the documentation for `precompute_size_scattering()`.
+        size : tuple
+            Returns a tuple of size `1+max_order` containing the number of
+            coefficients in each order.
         """
+        return precompute_size_scattering(self.J, self.Q, self.T, self.max_order)
 
-        return precompute_size_scattering(
-            self.J, self.Q, self.T, max_order=self.max_order, detail=detail)
+    @property
```
MuawizChaudhary, line 122: Wonderful

kymatio/scattering1d/frontend/numpy_frontend.py
```diff
@@ -46,7 +46,7 @@ def scattering(self, x):
         # treat the arguments
         if self.vectorize:
             size_scattering = precompute_size_scattering(
-                self.J, self.Q, self.T, max_order=self.max_order, detail=True)
+                self.J, self.Q, self.T, max_order=self.max_order)
```
MuawizChaudhary, line 49: Could we not just do `size_scattering = self.num_coefficents`?

tests/scattering1d/test_torch_scattering1d.py
```diff
@@ -266,25 +266,21 @@ def test_precompute_size_scattering(device, backend, random_state=42):
...
+            for k in s_dico.keys():
+                if k is ():
+                    num_orders[0] += 1
                 else:
-                    assert len(s_dico) == size
+                    if len(k) == 1:  # order1
+                        num_orders[1] += 1
+                    elif len(k) == 2:
+                        num_orders[2] += 1
+            todo = 2 if max_order == 2 else 1
```
MuawizChaudhary, line 280: ?

tests/scattering1d/test_utils_scattering1d.py
```diff
@@ -55,9 +55,11 @@ def test_border_indices(random_state=42):
 
 # Check that the default frontend is numpy and that errors are correctly launched.
 def test_scattering1d_frontend():
-    scattering = Scattering1D(2, shape=(10, ))
+    scattering = Scattering1D(2, shape=(16, ))
```
MuawizChaudhary, line 58: why?

MuawizChaudhary (changes requested):

tests/scattering1d/test_utils_scattering1d.py
lostanlen: To avoid this warning

```
/Users/user/kymatio/kymatio/scattering1d/filter_bank.py:309: UserWarning: Signal support is too small to avoid border effects
  warnings.warn('Signal support is too small to avoid border effects')
```

kymatio/scattering1d/frontend/numpy_frontend.py
lostanlen: this code block will disappear with #837. i wouldn't worry about it

tests/scattering1d/test_torch_scattering1d.py
```diff
@@ -263,28 +264,9 @@ def test_precompute_size_scattering(device, backend, random_state=42):
...
-                                num_orders[2] += 1
-                    todo = 2 if max_order == 2 else 1
-                    for i in range(todo):
-                        assert num_orders[i] == size[i]
-                        # check that the orders are completely equal
-                else:
-                    assert len(s_dico) == size
+        ctr = Counter(map(len, scattering(x)))
+        assert scattering.output_size() == (ctr[0], ctr[1], ctr[2])
+        assert scattering.num_coefficients == sum(ctr.values())
```
MuawizChaudhary, line 269: what do you think @janden

kymatio/scattering1d/frontend/numpy_frontend.py
```diff
@@ -41,6 +41,14 @@ def scattering(self, x):
...
         x = x.reshape((-1, 1) + signal_shape)
 
+        # get the arguments before calling the scattering
+        # treat the arguments
+        if self.vectorize:
+            size_scattering = precompute_size_scattering(
+                self.J, self.Q, self.T, max_order=self.max_order)
+        else:
+            size_scattering = 0
+
```
MuawizChaudhary, line 51: shouldnt this be removed?

kymatio/scattering1d/frontend/tensorflow_frontend.py
```diff
@@ -44,6 +44,14 @@ def scattering(self, x):
...
         x = tf.reshape(x, tf.concat(((-1, 1), signal_shape), 0))
 
+        # get the arguments before calling the scattering
+        # treat the arguments
+        if self.vectorize:
+            size_scattering = precompute_size_scattering(
+                self.J, self.Q, self.T, max_order=self.max_order)
+        else:
+            size_scattering = 0
+
```
MuawizChaudhary, line 54: ditto

kymatio/scattering1d/frontend/torch_frontend.py
```diff
@@ -98,6 +98,15 @@ def scattering(self, x):
...
 
+        # get the arguments before calling the scattering
+        # treat the arguments
+        if self.vectorize:
+            size_scattering = precompute_size_scattering(
+                self.J, self.Q, self.T, max_order=self.max_order)
+        else:
+            size_scattering = 0
+
+
```
MuawizChaudhary, line 109: ditto

MuawizChaudhary (changes requested):

lostanlen:
@janden your `output` is requested 😊

MuawizChaudhary (approved):
Most of my comments have been approved. @janden it would be nice to have your approval. Else I'll merge in a day ;)

lostanlen:
actually, don't merge this. i'm realizing that we're switching from detail=False by default to detail=True by default. So we definitely need a deprecation cycle. I'll work on it

MuawizChaudhary:
Wait what? I thought we discussed this and that this doesn't need a deprecation cycle? We are doing a sum on the output.

lostanlen:
@MuawizChaudhary ah yes, sorry. i got confused between current default and intended default. So the implementation we want is `never-detail`, not `always-detail` right? But then why did you ask for a `num_coefficients` property?

PR #844, opened 2022-06-06
Remove `vectorize`, introduce `out_type="dict"`

### LINKED ISSUE #843: `vectorize` should be removed
In Scattering1D
`vectorize=True` should become `out_type="array"`
`vectorize=False` should become `out_type="list"`

It's already being deprecated for a while now, so this removal can be included for v0.3

lostanlen:
Fixes #843
I'm proposing a new value of `out_type`, namely, `dict`, to support the previous setting of `out_type="array"` and `not vectorize` (we will need to document the new `out_type` behavior a little bit better than it currently is. your feedback is valued)

Net loss in LOC count
Updated docs in base frontend accordingly

### DIFF
--- a/kymatio/scattering1d/frontend/base_frontend.py
+++ b/kymatio/scattering1d/frontend/base_frontend.py
@@ -11,7 +11,7 @@ compute_meta_scattering, precompute_size_scattering)
 
 class ScatteringBase1D(ScatteringBase):
     def __init__(self, J, shape, Q=1, T=None, max_order=2, average=True,
-            oversampling=0, vectorize=True, out_type='array', backend=None):
+            oversampling=0, out_type='array', backend=None):
         super(ScatteringBase1D, self).__init__()
         self.J = J
         self.shape = shape
@@ -20,7 +20,6 @@ class ScatteringBase1D(ScatteringBase):
         self.max_order = max_order
         self.average = average
         self.oversampling = oversampling
-        self.vectorize = vectorize
         self.out_type = out_type
         self.backend = backend
 
@@ -165,20 +164,12 @@ class ScatteringBase1D(ScatteringBase):
     r"""average : boolean
             Controls whether the output should be averaged (the standard
             scattering transform) or not (resulting in wavelet modulus
-            coefficients). Note that to obtain unaveraged output, the
-            `vectorize` flag must be set to `False` or `out_type` must be set
-            to `'list'`.
+            coefficients). If `average` is set to `False`, `out_type` must be
+            set to `'list'`.
      """
 
-    _doc_param_vectorize = \
-    r"""vectorize : boolean, optional
-            Determines wheter to return a vectorized scattering transform
-            (that is, a large array containing the output) or a dictionary
-            (where each entry corresponds to a separate scattering
-            coefficient). This parameter may be modified after object
-            creation. Deprecated in favor of `out_type` (see below). Defaults
-            to True.
-        out_type : str, optional
+    _doc_param_out_type = \
+    r"""out_type : str, optional
             The format of the output of a scattering transform. If set to
             `'list'`, then the output is a list containing each individual
             scattering coefficient with meta information. Otherwise, if set to
@@ -187,13 +178,8 @@ class ScatteringBase1D(ScatteringBase):
             `'array'`.
         """
 
-    _doc_attr_vectorize = \
-    r"""vectorize : boolean
-            Controls whether the output should be vectorized into a single
-            Tensor or collected into a dictionary. Deprecated in favor of
-            `out_type`. For more details, see the documentation for
-            `scattering`.
-        out_type : str
+    _doc_attr_out_type = \
+    r"""out_type : str
             Specifices the output format of the transform, which is currently
             one of `'array'` or `'list`'. If `'array'`, the output is a large
             array containing the scattering coefficients. If `'list`', the
@@ -289,7 +275,7 @@ class ScatteringBase1D(ScatteringBase):
             calculation. If this is not desirable, `oversampling` can be set
             to a large value to prevent too much subsampling. This parameter
             may be modified after object creation. Defaults to `0`.
-        {param_vectorize}
+        {param_out_type}
         Attributes
         ----------
         J : int
@@ -306,7 +292,7 @@ class ScatteringBase1D(ScatteringBase):
         {attr_average}oversampling : int
             The number of powers of two to oversample the output compared to
             the default subsampling rate determined from the filters.
-        {attr_vectorize}"""
+        {attr_out_type}"""
 
     _doc_scattering = \
     """Apply the scattering transform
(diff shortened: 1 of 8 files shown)


kymatio/scattering1d/frontend/base_frontend.py
```diff
@@ -165,20 +164,12 @@ def output_size(self, detail=False):
     r"""average : boolean
             Controls whether the output should be averaged (the standard
             scattering transform) or not (resulting in wavelet modulus
-            coefficients). Note that to obtain unaveraged output, the
-            `vectorize` flag must be set to `False` or `out_type` must be set
-            to `'list'`.
+            coefficients). If `average` is set to `False`, `out_type` must be
+            set to `'list'`.
```
lostanlen, line 168: or 'dict'

kymatio/scattering1d/frontend/base_frontend.py
```diff
@@ -347,10 +323,8 @@ def output_size(self, detail=False):
        Returns
        -------
        S : tensor or dictionary
-           If `out_type` is `'array'` and the `vectorize` flag is `True`, the
-           output is a{n} `{array}` containing the scattering coefficients,
-           while if `vectorize` is `False`, it is a dictionary indexed by
-           tuples of filter indices. If `out_type` is `'list'`, the output is
+           If `out_type` is `'array'`, the output is a{n} `{array}` containing
+           the scattering coefficients. If `out_type` is `'list'`, the output is
```
lostanlen, line 327: document 'dict'

kymatio/scattering1d/frontend/numpy_frontend.py
```diff
@@ -22,20 +22,6 @@ def scattering(self, x):
...
                     len(x.shape)))
 
-        if not self.out_type in ('array', 'list'):
-            raise RuntimeError("The out_type must be one of 'array' or 'list'.")
-
-        if not self.average and self.out_type == 'array' and self.vectorize:
-            raise ValueError("Options average=False, out_type='array' and "
-                             "vectorize=True are mutually incompatible. "
-                             "Please set out_type to 'list' or vectorize to "
-                             "False.")
```
MuawizChaudhary, line 32: I think we wanna keep this non?

kymatio/scattering1d/frontend/tensorflow_frontend.py
```diff
@@ -24,21 +23,6 @@ def scattering(self, x):
...
                     len(x.shape)))
 
-        if not self.out_type in ('array', 'list'):
-            raise RuntimeError("The out_type must be one of 'array' or 'list'.")
-
-        if not self.average and self.out_type == 'array' and self.vectorize:
-            raise ValueError("Options average=False, out_type='array' and "
-                             "vectorize=True are mutually incompatible. "
-                             "Please set out_type to 'list' or vectorize to "
-                             "False.")
```
MuawizChaudhary, line 34: Ditto

kymatio/scattering1d/frontend/torch_frontend.py
```diff
@@ -76,37 +76,19 @@ def scattering(self, x):
...
                     len(x.shape)))
 
-        if not self.out_type in ('array', 'list'):
-            raise RuntimeError("The out_type must be one of 'array' or 'list'.")
-
-        if not self.average and self.out_type == 'array' and self.vectorize:
-            raise ValueError("Options average=False, out_type='array' and "
-                             "vectorize=True are mutually incompatible. "
-                             "Please set out_type to 'list' or vectorize to "
-                             "False.")
```
MuawizChaudhary, line 86: ditto

MuawizChaudhary (changes requested):
Looks mostly good to me.

kymatio/scattering1d/frontend/numpy_frontend.py
lostanlen: these errors now live in `core/scattering1d.py`, to avoid repeating the text across frontends:

```python
    if out_type not in ['array', 'dict', 'list']:
        raise RuntimeError("'out_type' must either be 'array', 'dict', or 'list'.")

    if out_type=="array" and not average:
        raise RuntimeError("out_type=='array' and average are mutually incompatible.")
```

MuawizChaudhary: we should move these lines to the base frontend object then

lostanlen: but which method of `ScatteringBase1D`? it does not have a `scattering` method. Should i introduce a `check_out_type` method or something like that? Or call the method `scattering` and call `super().scattering(self, input) ?

MuawizChaudhary: Yeah, introduce `check_out_type` for now. what do you think @janden ?

lostanlen: i like this `check_out_type` idea

lostanlen: implemented in ff8344a

kymatio/scattering1d/frontend/base_frontend.py
```diff
@@ -11,7 +11,7 @@
 
 class ScatteringBase1D(ScatteringBase):
     def __init__(self, J, shape, Q=1, T=None, max_order=2, average=True,
-            oversampling=0, vectorize=True, out_type='array', backend=None):
```
MuawizChaudhary, line 14: Why is the `out_type` here `array`, but default above is `dict`?

kymatio/scattering1d/frontend/base_frontend.py
```diff
@@ -306,34 +292,24 @@ def output_size(self, detail=False):
...
-       containing `C` keys, each a tuple whose length corresponds to the
-       scattering order and whose elements are the sequence of filter indices
-       used.
-
-       Note that the `vectorize` flag has been deprecated in favor of the
-       `out_type` parameter. If this is set to `'array'` (the default), the
-       `vectorize` flag is still respected, but if not, `out_type` takes
-       precedence. The two current output types are `'array'` and `'list'`.
-       The former gives the type of output described above. If set to
-       `'list'`, however, the output is a list of dictionaries, each
```
MuawizChaudhary, line 332: We should set `out_type` default to `array` like we said we would.

MuawizChaudhary (changes requested):

kymatio/scattering1d/frontend/base_frontend.py
lostanlen: for backwards compatibility. The current default is `out_type="array"` but `vectorize=False`, meaning, we don't return an array but a dict !!

We do
```python
if out_type == 'array' and not vectorize:
        out_S = {x['n']: x['coef'] for x in out_S}
```

which is extremely confusing and why i propose to remove the `vectorize` kwarg.

You might argue in favor of changing the default `out_type` to `"array"`, and i'd agree with you, but that would technically be a breaking change (even though all PT/TF/NP frontends override the `vectorize=False` default of `base_frontend` to make it `vectorize="true"`, and thus return an array)

I did not take the liberty to make this breaking change because i wanted to keep the PR minimal. Hope this clarifies

lostanlen: Also note that `out_type="array"` as a default has some drawbacks, not the least of which being that it assumes `average` to be `True`.
passing `average=False` to this function and letting `out_type` be an array would cause an error at runtime, which we catch in the non-base frontends.

Perhaps the safest solution is no default value at all?

PR #845, opened 2022-06-06
MAINT, STY Various ASV maintenance, style changes. 

MuawizChaudhary:
This Pull Request add's some style changes and removes some unneeded cuda synchronization. 

I wanted to add some documentation, but turns out y'all already added documentation during the 2021 sprint. 

I think all we need to do in terms of ASV related stuff is figure out how to integrate it into our continuous integration stack ;P

### DIFF
--- a/benchmarks/benchmarks/torch_benchmark/scattering1d.py
+++ b/benchmarks/benchmarks/torch_benchmark/scattering1d.py
@@ -83,11 +83,7 @@ class BenchmarkScattering1D:
         # can take some time
 
     def time_constructor(self, sc_params,  backend, device):
-        if device == 'cuda':
-            torch.cuda.synchronize()
         Scattering1D(backend=backend, J=sc_params["J"], shape=sc_params["shape"], Q=sc_params["Q"])
-        if device == 'cuda':
-            torch.cuda.synchronize()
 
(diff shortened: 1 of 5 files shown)


lostanlen (approved):

lostanlen:
see with @janden / @eickenberg about Jenkins deployment?

benchmarks/benchmarks/torch_benchmark/scattering1d.py
```diff
@@ -83,16 +83,13 @@ def setup(self, sc_params,  backend, device):
         # can take some time
 
     def time_constructor(self, sc_params,  backend, device):
-        if device == 'cuda':
-            torch.cuda.synchronize()
```
janden, line 87: Why don't we want to `synchronize`? Won't the timings be underestimated now?

janden (changes requested):

benchmarks/benchmarks/torch_benchmark/scattering1d.py
MuawizChaudhary: Non. This is the constructor for the Scattering. It doesn't involve cuda at all.

janden (approved):

PR #850, opened 2022-06-07
MAINT remove `calibrate_scattering_filters`

### LINKED ISSUE #849: `calibrate_scattering_filters` is unnecessary
`calibrate_scattering_filters` is probably my least favorite function from `filter_bank.py`. It it called only ~once~ thrice, and has no units test. It has six inputs, seven (!) outputs, and only five lines of code. (and 50 lines of docs!)

In the spirit of #513, i suggest we remove it and have  `scattering_filter_factory` call `compute_filter_params` directly.

lostanlen:
Fix #849 

Another PR with more deletions than additions.

This PR adds some new input arguments to `precompute_size_scattering` and `compute_meta_scattering` but i have a plan on how to remove them later while maintaining backwards compatibility. (#851)


tests/scattering1d/test_torch_scattering1d.py
```diff
@@ -428,3 +428,15 @@ def test_T(device, backend):
...
+@pytest.mark.parametrize("backend", backends)
+def test_Q(device, backend):
+    J = 3
+    length = 1024
+    shape = (length,)
+
+    with pytest.raises(ValueError) as ve:
+        _ = Scattering1D(
+            J, shape, Q=0.9, backend=backend, frontend='torch').to(device)
+    assert "Q should always be >= 1" in ve.value.args[0]
```
MuawizChaudhary, line 442: Why is this test here?

lostanlen: because there is no more `calibrate_scattering_filters`  and thus no `test_calibrate_scattering_filters` either. Instead, i am catching the case `Q<1` in the `base_frontend`

I am happy to reorganize the test suite files if you judge it necessary

PR #855, opened 2022-06-08
Simplify `scattering1d`

### LINKED ISSUE #852: Unnecessary arguments in `core/scattering1d/scattering1d`
Not a big priority, but an easy simplification. The current prototype of `scattering` is: (phew!)

```python
scattering1d(x, pad, unpad, backend, log2_T, psi1, psi2, phi, pad_left=0,
        max_order=2, average=True, vectorize=False, out_type='array')
```

This didn't come to my attention while reviewing #673 , but the 5th input parameter, `log2_T`, is unnecessary: it is already available as `phi["j"]`. So we could just remove it, and put `log2_T = phi["j"]` at the top of the function.

Thoughts?

lostanlen:
Fixes #852 and does more :)

This PR removes five input arguments from the core `scattering1d` function while respecting backwards compatibility. 
It will add a few lines of code inside `scattering1d` but will make calls for the 1D core way more lightweight and readable.

Here's how i did it:

```python
pad = backend.pad
unpad = backend.unpad

# compute the padding quantities:
N = x.shape[-1]
N_pad = phi[0].shape[-1]
pad_left, pad_right = compute_padding(N, N_pad)

# compute start and end indices
log2_T = phi["j"]
J = max(max(map(itemgetter("j"), psi1)), max(map(itemgetter("j"), psi2)))
ind_start, ind_end = compute_border_indices(log2_T, J, pad_left, pad_left+N)
```

I only had to modify `compute_padding` so that it takes `N, N_pad` as argument instead of `J_pad,N`. But it's equivalent since`N_pad=2**J_pad`.

This first commit is a fully working example, ready for review. After it's accepted, i'll take care of cleaning up extra keyword arguments from core and call sites (TF/PT/NP frontends) and will confirm the new prototype of `scattering1d` and `compute_padding`

### DIFF
--- a/kymatio/scattering1d/frontend/base_frontend.py
+++ b/kymatio/scattering1d/frontend/base_frontend.py
@@ -4,8 +4,8 @@ import numbers
 
 import numpy as np
 
-from ..filter_bank import scattering_filter_factory
-from ..utils import (compute_border_indices, compute_padding, compute_minimum_support_to_pad,
+from ..filter_bank import compute_temporal_support, gauss_1d, scattering_filter_factory
+from ..utils import (compute_border_indices, compute_padding,
 compute_meta_scattering, precompute_size_scattering)
 
 
@@ -38,7 +38,6 @@ class ScatteringBase1D(ScatteringBase):
         self.alpha = 5.
         self.P_max = 5
         self.eps = 1e-7
-        self.criterion_amplitude = 1e-3
         self.normalize = 'l1'
 
         # check the shape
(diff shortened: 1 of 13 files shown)


kymatio/scattering1d/frontend/base_frontend.py
```diff
@@ -71,11 +71,6 @@ def build(self):
         J_max_support = int(np.floor(np.log2(3 * self.N - 2)))
         self.J_pad = min(int(np.ceil(np.log2(self.N + 2 * min_to_pad))),
                          J_max_support)
-        # compute the padding quantities:
-        self.pad_left, self.pad_right = compute_padding(self.J_pad, self.N)
-        # compute start and end indices
-        self.ind_start, self.ind_end = compute_border_indices(
-            self.log2_T, self.J, self.pad_left, self.pad_left + self.N)
```
MuawizChaudhary, line 78: I prefer this version of doing the padding calculation. I don't think this calculation should be put into core.

MuawizChaudhary (changes requested):
Not sure I support this PR. @janden @eickenberg thoughts?

kymatio/scattering1d/frontend/base_frontend.py
lostanlen: why?

lostanlen:
i'm struggling with name resolutions and module imports but this is clearly doable ([a9d2db6](https://github.com/kymatio/kymatio/pull/855/commits/a9d2db67b2590c223e96b101e379b967d8336630))

The question is should we do it.

Issue #735, opened 2021-04-27
Towards modularizing the codebase in a semantically meaningful way


eickenberg:
Hi everybody,

It's been about a year now that I have been simmering thoughts about how to organize the kymatio code base away from the current hard-baked design choices and towards a modular architecture that maintains the same outputs and interfaces in the default settings, but that allows researchers to import parts of the code base instead of requiring to rewrite parts when they want to do anything new.
In addition, we are actually hiding away several types of wavelet transforms and fourier convolutions whose results we cannot access other than by their smoothed moduli. It should be possible for users to access these intermediate states in an easy way.

Any modifications to the code base must respect some guiding principles.
1. They should implement new functionality for all backends
2. They should maintain or improve computation speeds
3. They should maintain or improve memory requirements
4. They should at least tend towards unifying the codebase across dimensionalities of signal (though this is always hard and case-dependent. There is a reason we set up the codebase such that the different dimensionalities of signal can be treated with different code)

One important thing to maintain through any modifications is the possibility of depth-first traversal of the scattering tree. It is at the leaves of this tree that the outputs live and the outputs have lower memory requirements than the intermediate states. Storing all intermediate states due to breadth-first search may become prohibitive in certain settings, especially for 3D signals at high resolution. This depth-first-requirement makes it less simple to specify wavelets and scattering layers as a stack of feed-forward modules in the spirit of `torch.nn.Sequential`.

One way of addressing this issue would be to create modules with hooks/callbacks in which you hook a second-order scattering layer into a wavelet transform object.

Fortunately, python comes to our rescue with a more elegant solution: iterator/generator pipelines

Modularizing Kymatio with generator pipelines
-------------------------------------------------------------

I see the following semantic split of kymatio functionalities:
1. (Convolutions)
2. Wavelet Transforms
3. Scattering Transforms

Scattering uses wavelets and we have implemented several types of them. Making a wavelet object can expose them to the user. Wavelet transforms are implemented as convolutions with very specific filters. Sometimes it is useful to use a generic Fourier convolution to implement a wavelet transform, but sometimes there are other ways of computing wavelet transforms more efficiently.

[...]

PR #860, opened 2022-06-09
avoid passing `log2_T`, `pad`, `unpad` to 1D core

### LINKED ISSUE #852: Unnecessary arguments in `core/scattering1d/scattering1d`

lostanlen:
Fixes #852

a lighter and better focused version of #855

unfortunately the same thing cannot be done in 2D right now because the logic is quite different: `self.pad` is a backend function which is being defined at `build` time and takes `self.shape` as input. It also depends upon a parameter `pre_pad`. It's almost like an object inside the `Scattering2D` object.

Next step for me is to remove `pad_left` and friends and perform padding/unpadding outside of scattering. We've discussed a plan with @MuawizChaudhary for that. (in the spirit of #735 )

### DIFF
--- a/kymatio/scattering1d/core/scattering1d.py
+++ b/kymatio/scattering1d/core/scattering1d.py
@@ -1,5 +1,5 @@
 
-def scattering1d(x, pad, unpad, backend, log2_T, psi1, psi2, phi, pad_left=0,
+def scattering1d(x, backend, psi1, psi2, phi, pad_left=0,
         pad_right=0, ind_start=None, ind_end=None, oversampling=0,
         max_order=2, average=True, vectorize=False, out_type='array'):
     """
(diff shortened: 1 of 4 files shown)


MuawizChaudhary (approved):
Merging.

PR #861, opened 2022-06-09
Remove vectorize + write frontend-agnostic runtime checks

lostanlen:
Fix #843 
(a second attempt after some productive discussions during review of #844)

This is a planned change for v0.3

The default behavior is still `out_type="array"` (`vectorize=True`)
We introduce `out_type="dict"` to replicate the behavior of `out_type="array", vectorize=False`

A net reduction in code size thanks to utility methods `_check_runtime_args()` and `_check_input()` in `ScatteringBase1D`

Written on top of #860. I suggest we look at #860 first
Tests pass locally

Note that vectorize is still documented in the base frontend. I'll take care of updating these docs after code review.

PR #862, opened 2022-06-09
Remove `out_type` from 1D core, perform formatting in the frontends

lostanlen:
Address (but does not fully fix)  #852 

Third PR in a streak (after #860 and #861), attempting to simplify core 1D scattering and decouple its functionalities.
This time the offender is `out_type` and the post-processing that comes after the scattering transform per se.

It turns out that there isn't much of a reason to write a backend-agnostic implementation of this formatting stage, considering that it's already divided into `"array"`, `"dict"` and `"list"` cases.

Besides, this PR brings us closer than we've ever been to a `Scattering1D` generator, which would `yield` coefficients instead of `return`-ing them. Hence a _big_ simplification in scope once we're able to decouple `scattering1d` from padding, averaging, and unpadding routines.

### DIFF
--- a/kymatio/scattering1d/frontend/numpy_frontend.py
+++ b/kymatio/scattering1d/frontend/numpy_frontend.py
@@ -10,7 +10,7 @@ class ScatteringNumPy1D(ScatteringNumPy, ScatteringBase1D):
             oversampling=0, out_type='array', backend='numpy'):
         ScatteringNumPy.__init__(self)
         ScatteringBase1D.__init__(self, J, shape, Q, T, max_order, average,
-                oversampling, vectorize, out_type, backend)
+                oversampling, out_type, backend)
         ScatteringBase1D._instantiate_backend(self, 'kymatio.scattering1d.backend.')
         ScatteringBase1D.build(self)
         ScatteringBase1D.create_filters(self)
--- a/kymatio/scattering1d/frontend/numpy_frontend.py
+++ b/kymatio/scattering1d/frontend/numpy_frontend.py
@@ -41,7 +41,7 @@ class ScatteringNumPy1D(ScatteringNumPy, ScatteringBase1D):
 
         x = x.reshape((-1, 1) + signal_shape)
 
-        S = scattering1d(x, self.backend.pad, self.backend.unpad, self.backend, self.log2_T, self.psi1_f, self.psi2_f,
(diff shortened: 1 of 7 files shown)


MuawizChaudhary:
I get the impression that we have to rebase this on top of #861 ?

lostanlen:
> I get the impression that we have to rebase this on top of #861 ?

Yes. #861 takes precedence

kymatio/scattering1d/frontend/numpy_frontend.py
```diff
@@ -27,26 +27,26 @@ def scattering(self, x):
...
             for k, v in S.items():
                 # NOTE: Have to get the shape for each one since we may have
                 # average == False.
                 scattering_shape = v.shape[-2:]
                 new_shape = batch_shape + scattering_shape
-
                 S[k] = v.reshape(new_shape)
         elif self.out_type == 'list':
             for x in S:
+                x.pop('n')
```
MuawizChaudhary, line 47: Why do we pop in the first place?

MuawizChaudhary (approved):
I approve these changes. @lostanlen I'll merge once you've addressed my comment.

kymatio/scattering1d/frontend/numpy_frontend.py
lostanlen: hm ... you're right, we don't have to. I think that we did this to match the returned formats of core 2D. Note that the dictionaries in the `out_S` list do not have an `n` field.
https://github.com/kymatio/kymatio/blob/master/kymatio/scattering2d/core/scattering2d.py

I can't see another justification for what we're doing now in 1D, i.e. to redact metadata that could potentially be useful for the user. `n`can be seen as a "pointer" (or tuple of pointers) to the filterbank `psi1_f` (or tuple of filterbanks `(psi1_f, psi2_f)`)

i'd be in favor of not popping it away. that change is backwards compatible and very easy to maintain (we'd be _removing_ a for loop!) This change adds some new functionality; but it's one that the users already had access too in a different way, namely, `S.keys()`under `out_type="dict"`

TL;DR i can delete that line if that makes you happy!

MuawizChaudhary: Lets do it in a separate repo. For now can you create an issue?

MuawizChaudhary:
I'm squashing and merging. @lostanlen pls create an issue as discussed above.

PR #865, opened 2022-06-10
MAINT redefine cdgmm3d to call cdgmm in torch skcuda backend

MuawizChaudhary:
This pull request redefines cdgmm3d to call cdgmm in torch skcuda backend. 

I'm also using this as a proxy for testing the torch_skcuda 3d backend.

### DIFF
--- a/kymatio/scattering3d/backend/torch_skcuda_backend.py
+++ b/kymatio/scattering3d/backend/torch_skcuda_backend.py
@@ -9,80 +9,8 @@ from .torch_backend import TorchBackend3D
 
 
 class TorchSkcudaBackend3D(TorchSkcudaBackend, TorchBackend3D):
-    @classmethod
-    def cdgmm3d(cls, A, B):
-        """Complex pointwise multiplication.
-
-            Complex pointwise multiplication between (batched) tensor A and tensor B.
-
-            Parameters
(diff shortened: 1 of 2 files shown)


lostanlen:
is this tied to a specific issue? what's the plan here?

MuawizChaudhary:
my reasoning: why are we repeating code that we already have?

lostanlen:
oh wow, we have duplicate code here? then yes sounds like an easy win

lostanlen:
is the change from a `RuntimeError` to a `TypeError` harmless here?

MuawizChaudhary:
define harmless? 

I changed it because it seemed like we forgot to switch the 3d cgmm over to throwing a [Type Error when we changed the 1d and 2d code to be unified](https://github.com/kymatio/kymatio/commit/69f132af14b1108ba70ae3ac9c945e4b0bb913b0)

MuawizChaudhary:
> > my reasoning: why are we repeating code that we already have?
> 
> Maybe there was a reason for having a separate `cdgmm3d`? On the other hand, the tests pass…

The code is the same except for the error checking.

janden:
Ok sounds good then.

janden (approved):

Issue #870, opened 2022-06-11
`scattering(self, x)` could (should?) be frontend-agnostic


lostanlen:
Right now, our methods `scattering(self, x)` are frontend-specific, not frontend-agnostic: one is defined in each frontend. Of course, the code in these frontends look very similar: they do argument check, input check, input reshape, scattering, and output reshape.

We have a `self.backend` object, so there is no reason why we couldn't perform all these operations in `base_frontend.py`, with appropriate `shape`, `reshape_input`, and `reshape_output` routines being methods of `self.backend`.

This would help us maintain the code because there would be a single point of entry for the `if out_type="array"` clause and its friends. And i believe it is compliant with #735 philosophy, since it decouples shape pre/post-processing from core scattering.

I have written a short PR where i do this in 1D, to showcase the idea. I surmise that the same would be possible in 2D and 3D.

janden:
I think this makes sense. I remember looking over the frontend code and noticing that there is a lot of duplication here that can be reduced…

PR #871, opened 2022-06-11
`reshape_input`, `reshape_output`

### LINKED ISSUE #895: what should be the returned shape of `Scattering1D` paths under `out_type='dict'` ?
Spotted while reviewing #871 with @janden

Right now, `out_type='dict'`behaves strangely. 

```python
J, Q = 2, 1
length = 1024
S = Scattering1D(J, shape, Q)
x = torch.zeros((5, 10,) + shape)

Sx = S(x)
print(Sx.shape)

S.out_type = "dict"
Sx = S(x)
print(list(map(lambda v: v.shape, Sx.values())))
```

Result: 
```
torch.Size([5, 10, 4, 256])

[torch.Size([5, 10, 1, 256]),
 torch.Size([5, 10, 1, 256]),
 torch.Size([5, 10, 1, 256]),
 torch.Size([5, 10, 1, 256])]
```

In other words, `out_type="dict"` is inserting a singleton dimension in penultimate position. This penultimate position serves to concatenate coefficients under `out_type="array"` (it becomes `4` in the example above).

I am in favor of removing this extra singleton dimension, and making values in `out_type="dict"` more like `'coef'`'s of `out_type="list"`. Compare with 

```python
S.out_type = "list"
Sx = S(x)
list(map(lambda v: v['coef'].shape, Sx))
```

Result:

```
[torch.Size([5, 10, 256]),
 torch.Size([5, 10, 256]),
 torch.Size([5, 10, 256]),
 torch.Size([5, 10, 256])]
```

This would be a breaking change, which is why i didn't make it single-handedly in my open PR #871.
I'm curious to hear everyone's thought here? bug or feature?

lostanlen:
A proposed fix for #870 

(EDIT: also fixes #895 )

includes a working demo in `Scattering1D` module
the same idea could work in other dimensions with the same methods `reshape_input` and `reshape_output` (no need to specialize them per 1D/2D/3D scattering!)

to help reviewers, i haven't moved the code from specialized frontends to the base frontend, only created call sites for `reshape_input` and `reshape_output` in 1D. But that's the idea of course. (could be done in a separate PR after review and resolution of merge conflicts)


kymatio/scattering1d/frontend/torch_frontend.py
```diff
@@ -67,36 +67,26 @@ def load_filters(self):
                     n += 1
 
     def scattering(self, x):
+        self.load_filters()
```
MuawizChaudhary, line 70: why called before checks?

MuawizChaudhary (commented):
I feel like its better to have these operations in the frontend rather than implemented in the backend. perhaps each classes base frontend class?

kymatio/scattering1d/frontend/torch_frontend.py
lostanlen: good question! because the goal is for this method to become:

```
def scattering(self, x):
    self.load_filters()
    super(ScatteringBase1D, self).scattering(x)
```

lostanlen:
yes, we could do that, but consider that:

(1) padding already belongs to backends, so i don't see why other array ops like reshapes should be in frontends

(2) if we're putting it in the frontend, users will see it. and i don't think they'll have much use of methods called `scattering.reshape_input` and `scattering.reshape_output` . So we might turn them into private methods `scattering._reshape_input` and `scattering._reshape_output`. But i believe that it's preferable to have public methods in the backend module rather than private methods in the frontend object.

lostanlen:
i'm having some difficulty rebasing this PR. it's not entirely clear to me what are the specifications of array shapes depending on whether we have out_type="array", "dict", or "list".

It seems like we insert a `1` between batch shape and signal shape that we end up removing sometimes, but not always? can anyone help?

lostanlen:
alright, now attempting to unify the frontends.

the latest commit ([9637abb](https://github.com/kymatio/kymatio/pull/871/commits/9637abb65214936c48372d57ce019c0f1ef0b270)) is where the code diff is going to be easiest to understand

MuawizChaudhary (approved):
Pro merging

lostanlen:
@MuawizChaudhary i'm assuming that this one too is being postponed to post-v0.3 ?

kymatio/scattering1d/frontend/numpy_frontend.py
```diff
@@ -18,38 +18,26 @@ def __init__(self, J, shape, Q=1, T=None, max_order=2, average=True,
...
-                new_shape = batch_shape + scattering_shape
-                x['coef'] = x['coef'].reshape(new_shape)
-
-        return S
+        n_kept_dims = 1 + (self.out_type=="dict")
+        for n, path in enumerate(S):
+            S[n]['coef'] = self.backend.reshape_output(
+                path['coef'], batch_shape, n_kept_dims=n_kept_dims)
+
+        if self.out_type=='array':
```
cyrusvahidi, line 35: `self.out_type == 'array'` please. same for others

kymatio/scattering1d/frontend/torch_frontend.py
```diff
@@ -67,41 +67,28 @@ def load_filters(self):
...
-                new_shape = batch_shape + scattering_shape
-                x['coef'] = x['coef'].reshape(new_shape)
-
-        return S
+        n_kept_dims = 1 + (self.out_type=="dict")
+        for n, path in enumerate(S):
+            S[n]['coef'] = self.backend.reshape_output(
+                path['coef'], batch_shape, n_kept_dims=n_kept_dims)
+
+        if self.out_type=='array':
```
cyrusvahidi, line 86: same here `self.out_type == 'array'`

cyrusvahidi (changes requested):
small edits, then LGTM

kymatio/scattering1d/frontend/torch_frontend.py
lostanlen: 182ae3b

(search-and-replaced 12 times)

MuawizChaudhary:
> @MuawizChaudhary i'm assuming that this one too is being postponed to post-v0.3 ?

I think that's the right decision here.

cyrusvahidi (approved):

kymatio/backend/tensorflow_backend.py
```diff
@@ -6,12 +6,26 @@ class TensorFlowBackend(NumpyBackend):
...
+        return tf.stack(arrays, axis=dim)
 
     @staticmethod
     def modulus(x):
         norm = tf.abs(x)
 
         return norm
 
+    @staticmethod
+    def reshape_input(x, signal_shape, n_inserted_dims=0):
```
janden, line 19: Not following this `n_inserted_dims`, `n_kept_dims` business. Seems like the former is always `1` and the second is sometimes `1`, sometimes `2`. Should the latter case be handled by a `stack`–`concatenate` distinction?

janden (dismissed):
Looks good but a question on these `dims` arguments.

kymatio/backend/tensorflow_backend.py
lostanlen: i have raised an issue https://github.com/kymatio/kymatio/issues/895

> Should the latter case be handled by a stack–concatenate distinction?

I did this to avoid introducing two kymatio primitives (stack and concatenate) when `reshape_output` does the job up to now. Although i agree that "stack" and "concatenate" are more recognizable in terms of what they do.
I am happy to make that distinction in the backends if you think it's helpful to understand the logic here.

janden: > The former is always one for 1D. It would be two for 2D and three for 3D. Since i'm implementing this in a dimension-agnostic way, i need to pass `n_inserted_dims` as an input argument.

We're adding two singleton dimensions in 2D and three in 3D? Are you sure? If I understand correctly, this added dimension corresponds to the channel and is one for all the dimensions.

> For me, the extra `1` in `out_type="dict"` has no justification and we should present dict values as `(batch_size, time)`, just like with `"list"`.

I agree. We can discuss further on the issue.

lostanlen: > We're adding two singleton dimensions in 2D and three in 3D? Are you sure? 

My bad. We're adding one singleton dimension no matter what the input dimension is.
2D code

```python
        batch_shape = input.shape[:-2]
        signal_shape = input.shape[-2:]
        input = input.reshape((-1,) + signal_shape)
        S = scattering( ...
        scattering_shape = S.shape[-3:]
        new_shape = batch_shape + scattering_shape
        S = S.reshape(new_shape)
```


So, `n_inserted_dims` doesn't need to be modifiable. I'll update accordingly.

lostanlen: fixed in 4a274be

MuawizChaudhary (dismissed):
I think we should leave this as a draft for now.

MuawizChaudhary:
I'd be open after 0.3

lostanlen:
@MuawizChaudhary but why ... ? this is backwards-compatible, does not introduce new feature, and is LOC-neutral. It can only make our life easier downstream ...

lostanlen:
Rebased. While writing this i found out that our `concatenate`primitive has inconsistent default values for `dims` across backends, and that this kwarg should probably be called `axis` instead of `dims`, in the spirit of being compliant with the Array API.

kymatio/scattering1d/frontend/numpy_frontend.py
```diff
@@ -36,7 +36,7 @@ def scattering(self, x):
         elif self.out_type == 'dict':
             return {path['n']: path['coef'] for path in S}
         elif self.out_type == 'list':
-            return list(map(lambda path: path.pop('n')), S)
+            return S
```
MuawizChaudhary, line 39: i like this change.

lostanlen:
@cyrusvahidi this is next in line. please merge when you're ready

cyrusvahidi (approved):
bon i've checked out the tests. LGTM

Issue #872, opened 2022-06-11
Shorten duration of `test_gradients`


MuawizChaudhary:
`test_gradients` takes a long time to finish running. I propose we look into making the test take a smaller amount of time to finish running, or switch over to using a modulus that doesn't implement a custom backwards.

janden:
Would be nice. Maybe there are some settings there to tweak.

But yes, if we can use a stock modulus, that would be better… Maybe now that Torch has complex numbers that should be possible?

MuawizChaudhary:
@janden are we wanting to keep supporting torch 1.7? One thing we could do in that case to make the sqrt operation differentiable is adding a small epsilon.

janden:
> @janden are we wanting to keep supporting torch 1.7?

I'm not sure. That's a separate discussion.

> One thing we could do in that case to make the sqrt operation differentiable is adding a small epsilon.

That would break all the tests, so it'd be nice to avoid that…

janden:
Another option here is to test our modulus separately (using `test_gradients`) and only verify that we can run `backward` on the output of scattering (i.e., not test the gradient of the full transform).

MuawizChaudhary:
Given that we merged #873 I believe this discussion is no longer needed. We no longer use the long version of `test_gradients`.

PR #873, opened 2022-06-12
MAINT Speed up `test_gradients`

MuawizChaudhary:
This pull request speeds up `test_gradients` by enabling `fast_mode`. Previously it took 109 seconds to run, now it takes 17 seconds. 

> fast_mode ([bool](https://docs.python.org/3/library/functions.html#bool), optional) – Fast mode for gradcheck and gradgradcheck is currently only implemented for R to R functions. If none of the inputs and outputs are complex a faster implementation of gradcheck that no longer computes the entire jacobian is run; otherwise, we fall back to the slow implementation.

Note that since our scattering takes in as input a real valued signal, and outputs a real valued signal, torch allows us to enable `fast_mode` to avoid computing additional entries of a jacobian. 

[See more here.](https://pytorch.org/docs/master/generated/torch.autograd.gradcheck.html#torch.autograd.gradcheck)

This PR resolves issue #872

### DIFF
--- a/tests/scattering2d/test_torch_scattering2d.py
+++ b/tests/scattering2d/test_torch_scattering2d.py
@@ -245,4 +245,4 @@ class TestScatteringTorch2D:
             scattering = Scattering2D(2, shape=(8, 8), backend=backend,
                                       frontend='torch').double().to(device)
             x = torch.rand(2, 1, 8, 8).double().to(device).requires_grad_()
-            gradcheck(scattering, x, nondet_tol=1e-5)
+            gradcheck(scattering, x, nondet_tol=1e-5, fast_mode=True)


MuawizChaudhary:
lol `fast_mode` is implemented in pytorch 1.9 but not before. Maybe this is an un-needed change? 

I do believe that OLGD seemed to mention something about the tests taking a long time to finish running. perhaps this grad check business was slowing him down? Perhaps we should move this gradcheck to a seperate script+folder?

janden:
> lol `fast_mode` is implemented in pytorch 1.9 but not before. Maybe this is an un-needed change?

Ok and since we want to be testing 1.8 we'll have to run the slow check anyhow… One option is to just run the gradcheck on more recent torch versions. Another is to stop supporting old torch versions, but that's a separate question.

lostanlen:
@janden in the wake of https://github.com/kymatio/kymatio/pull/892, we no longer test torch 1.8
so i believe we could reopen this and merge it for v0.4

edouardoyallon (approved):

Issue #824, opened 2022-05-31
Utilize .to(device) semantics in examples


MuawizChaudhary:
We have many checks for `"cuda"`, and pass to gpu using `.cuda()` in our examples, but we could use `.to(device)` semantics and not have to worry about all that. 

This issue is related to the overarching project of fixing our examples. Please see issues, #817 , #812 and pull request #815 for related details.

janden:
Sounds great. Do it!

PR #826, opened 2022-06-02
DOC fix reconstruct_torch.py example by calling detach

MuawizChaudhary:
This pull request fixes an example from erroring out due to not detaching tensors from the computational graph before converting to numpy. 

This goes a small way of solving #817

PR #874, opened 2022-06-12
DOC Enable `.to(device)` semantics for pytorch examples

MuawizChaudhary:
This PR enables `to(device)` semantics for the pytorch examples. 

This resolves issue #824 .

### DIFF
--- a/examples/1d/reconstruct_torch.py
+++ b/examples/1d/reconstruct_torch.py
@@ -16,6 +16,7 @@ from kymatio.torch import Scattering1D
 
 from torch.autograd import backward
 import matplotlib.pyplot as plt
+device = torch.device("cuda" if torch.cuda.is_available else "cpu")
 
 ###############################################################################
 # Write a function that can generate a harmonic signal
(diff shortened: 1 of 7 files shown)


MuawizChaudhary:
We should really merge #829 , #826  before this.

examples/1d/reconstruct_torch.py
```diff
@@ -58,16 +59,16 @@ def generate_harmonic_signal(T, num_intervals=4, gamma=0.9, random_state=42):
 # Let’s take a look at what such a signal could look like.
 
 T = 2 ** 13
-x = torch.from_numpy(generate_harmonic_signal(T))
+x = torch.from_numpy(generate_harmonic_signal(T)).to(device)
```
janden, line 62: Maybe just move this call to just before the `scattering` call (so have `x = x.to(device)` on the line before)?

examples/2d/cifar_resnet_torch.py
```diff
@@ -158,20 +158,18 @@ def test(model, device, test_loader, scattering):
...
 
     model = Scattering2dResNet(K, args.width).to(device)
 
     # DataLoaders
+    num_workers = 4
     if use_cuda:
-        num_workers = 4
         pin_memory = True
     else:
-        num_workers = None
```
janden, line 174: ?

janden (changes requested):
Wonderful!

examples/2d/cifar_resnet_torch.py
MuawizChaudhary: Why should number of cpu workers change depending on if we're on cpu or not?

MuawizChaudhary:
looking at the commit history... did i rebase all wrong Q_Q

examples/2d/cifar_resnet_torch.py
janden: Agreed, but this was maybe there for a reason. Does it affect performance?

MuawizChaudhary: `num_workers = None` appears to give me an [error](https://colab.research.google.com/drive/10_bgL7GRPvrpLiMBxd__aK1mtdj3U4wr#scrollTo=0Pp33Tr7qZIM). 

I don't think there are many reasons to set `num_works = 0 ` when most modernday cpus are multicore. @eugenium care to chime in?

janden: Ok.

janden (approved):

janden:
Nm looks like it's the TF FFT implementation which is still flaky.

PR #878, opened 2022-06-14
`precompute_size_scattering` now returns a tuple

### LINKED ISSUE #838: `precompute_size_scattering` should always return a tuple (`detail=False`)

lostanlen:
no more `detail` kwarg
fixes #838

came up in conversation with @janden and @MuawizChaudhary : https://github.com/kymatio/kymatio/issues/838#issuecomment-1151084298

### DIFF
--- a/kymatio/scattering1d/utils.py
+++ b/kymatio/scattering1d/utils.py
@@ -70,8 +70,7 @@ def compute_padding(J_pad, N):
     return pad_left, pad_right
 
 
-def precompute_size_scattering(
-        J, Q, T, max_order, r_psi, sigma0, alpha, detail=False):
+def precompute_size_scattering(J, Q, T, max_order, r_psi, sigma0, alpha):
     """Get size of the scattering transform
 
     The number of scattering coefficients depends on the filter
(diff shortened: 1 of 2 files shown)


MuawizChaudhary (approved):

kymatio/scattering1d/utils.py
```diff
@@ -103,38 +102,27 @@ def precompute_size_scattering(
...
-        else:
-            return size_order0, size_order1
-    else:
-        if max_order == 2:
-            return size_order0 + size_order1 + size_order2
-        else:
-            return size_order0 + size_order1
+
+    if max_order == 2:
+        sizes.append(size_order2)
```
cyrusvahidi, line 124: great

cyrusvahidi (approved):

PR #884, opened 2022-06-15
1D deprecation warning for average

cyrusvahidi:
Altering the semantics of `T` parameter. It will effectively replace `average` in 1D in preparation for JTFS release in 0.4. `T` will also be used for frequency averaging in the frequential filterbank.

* `T=None` will perform default mode averaging, i.e. `T=2^J`
* `T=0` corresponds to `average=False`
* Any other integer `T` will stride over the specified support

### DIFF
--- a/kymatio/scattering1d/frontend/base_frontend.py
+++ b/kymatio/scattering1d/frontend/base_frontend.py
@@ -24,6 +24,11 @@ class ScatteringBase1D(ScatteringBase):
         self.out_type = out_type
         self.backend = backend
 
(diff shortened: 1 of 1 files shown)


lostanlen:
upon reflection, i no longer think this is beneficial. `T` is applied at filterbank construction time whereas `average` is applied at runtime. Presumably, `average` could be switched on and off without having to redesign a new `phi`. Also, if `T=0`, we'll still have to define `log2_T` as though `T` was `None`, which makes it sound like it's unintended.

lostanlen:
Actually, that might be OK

PR #863, opened 2022-06-09
MAINT deprecate `self.N` and `self.J_pad` in 1D

lostanlen:
Fixes #857 

An important deprecation to set before we release v0.3 is `J_pad`. Indeed, there won't be a `J_pad_fr` in JTFS, so it'd best to plan for symmetry in design.

```python
_N_padded, _N_padded_fr
J, J_fr
Q, Q_fr
T, F
log2_T, log2_F
```

I have moved the shape check to its own static method. We'll be able to call this method a second time inside `self._check_input(x)` (recently merged!)

PR #886, opened 2022-06-15
Deprecate `self.N` in 2D

lostanlen:
2D equivalent of #863 
triaged for v0.3

### DIFF
--- a/kymatio/scattering2d/frontend/base_frontend.py
+++ b/kymatio/scattering2d/frontend/base_frontend.py
@@ -43,7 +43,7 @@ class ScatteringBase2D(ScatteringBase):
     def N(self):
         warn("The attribute N is deprecated and will be removed in v0.4. "
         "Replace by shape[1].", DeprecationWarning)
-        return int(self.shape[0])
+        return int(self.shape[1])
 
     _doc_shape = 'M, N'
 
--- a/kymatio/scattering2d/frontend/base_frontend.py
+++ b/kymatio/scattering2d/frontend/base_frontend.py
@@ -17,23 +17,33 @@ class ScatteringBase2D(ScatteringBase):
         self.out_type = out_type
 
     def build(self):
-        self.M, self.N = self.shape
+        M, N = self.shape
 
-        if 2 ** self.J > self.M or 2 ** self.J > self.N:
+        if 2 ** self.J > M or 2 ** self.J > N:
             raise RuntimeError('The smallest dimension should be larger than 2^J.')
-        self.M_padded, self.N_padded = compute_padding(self.M, self.N, self.J)
+        self._M_padded, self._N_padded = compute_padding(M, N, self.J)
         # pads equally on a given side if the amount of padding to add is an even number of pixels, otherwise it adds an extra pixel
         if not self.pre_pad:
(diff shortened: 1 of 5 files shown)


MuawizChaudhary (approved):
Thanks for this.

lostanlen:
@janden merge at your convenience

kymatio/scattering2d/frontend/base_frontend.py
```diff
@@ -17,24 +17,34 @@ def __init__(self, J, shape, L=8, max_order=2, pre_pad=False,
         self.out_type = out_type
 
     def build(self):
-        self.M, self.N = self.shape
+        M, N = self.shape
 
-        if 2 ** self.J > self.M or 2 ** self.J > self.N:
+        if 2 ** self.J > M or 2 ** self.J > N:
             raise RuntimeError('The smallest dimension should be larger than 2^J.')
-        self.M_padded, self.N_padded = compute_padding(self.M, self.N, self.J)
```
MuawizChaudhary, line 24: Actually, I'm unsure why we do this change here. In the parametric scattering code we access `self.M_padded`, `self.N_padded`... does this need to have the underscore added to it?

MuawizChaudhary:
I think in this instance its worth it having a deprecation warning.

kymatio/scattering2d/frontend/base_frontend.py
lostanlen: the idea is that the user shouldn't attempt to change these values in the `Scattering2D` object after it's being constructed. So we put a leading underscore to them to signify that they are "privarte" (although in Python nothing is truly private and there's nothing preventing the user to get or set leading-underscore values)

You might say that the same could be said about `J`, `L`, etc. which really should be stored as `_J`, `_L`, etc. and accessed as properties, but one API change at a time.

kymatio/scattering2d/frontend/base_frontend.py
```diff
@@ -17,24 +17,34 @@ def __init__(self, J, shape, L=8, max_order=2, pre_pad=False,
...
+        self._M_padded, self._N_padded = compute_padding(M, N, self.J)
```
MuawizChaudhary, line 24: It might be good for this to give us a tuple containing the padded shape rather than having two seperate variable for ot.

kymatio/scattering2d/frontend/base_frontend.py
```diff
@@ -17,24 +17,34 @@ def __init__(self, J, shape, L=8, max_order=2, pre_pad=False,
...
             self.pad = lambda x: x
 
         self.unpad = self.backend.unpad
 
     def create_filters(self):
-        filters = filter_bank(self.M_padded, self.N_padded, self.J, self.L)
+        filters = filter_bank(self._M_padded, self._N_padded, self.J, self.L)
         self.phi, self.psi = filters['phi'], filters['psi']
 
+    def M(self):
```
cyrusvahidi, line 38: we don't want these as properties?

lostanlen: you're right: we do. updating

cyrusvahidi (approved):
good

MuawizChaudhary (approved):

PR #876, opened 2022-06-12
move core 1D asserts to `_check_filterbanks` in `ScatteringBase`

lostanlen:
Fixes #869 by @MuawizChaudhary 

don't mind the title of the branch: `oversampling` is not becoming immutable.
also don't mind the title of the commit: `_check_filterbanks` belongs in `ScatteringBase`, not `ScatteringBase1D`

PR #887, opened 2022-06-15
check `oversampling` in 1D

lostanlen:
in `check_runtime_args`

and write `test_check_runtime_args`

### DIFF
--- a/kymatio/scattering1d/frontend/base_frontend.py
+++ b/kymatio/scattering1d/frontend/base_frontend.py
@@ -118,11 +118,16 @@ class ScatteringBase1D(ScatteringBase):
 
     def _check_runtime_args(self):
         if not self.out_type in ('array', 'dict', 'list'):
-            raise RuntimeError("The out_type must be one of 'array', 'dict', or 'list'.")
+            raise ValueError("The out_type must be one of 'array', 'dict'"
+                             ", or 'list'. Got: {}".format(self.out_type))
 
         if not self.average and self.out_type == 'array':
             raise ValueError("Cannot convert to out_type='array' with "
                              "average=False. Please set out_type to 'dict' or 'list'.")
+
+        if self.oversampling<0:
+            raise ValueError("oversampling must be nonnegative. Got: {}".format(
+                self.oversampling))
 
     def _check_input(self, x):
         # basic checking, should be improved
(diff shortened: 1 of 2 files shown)


MuawizChaudhary (approved):
Should oversampling be set by user?

lostanlen:
Yes. although i'll admit i don't love the current semantics. `alpha` also achieves a form of oversampling when getting the j's from xi's and sigma's. It's also a but surprising that the stride (`if average`) is not `T` but `T/2**oversampling`

But these are discussions for another day. here the goal is to make sure that the value of oversampling is OK at runtime.

lostanlen:
i've written this to complement #876. The new asserts rely on the assumption that oversampling>=0, but this wasnt being checked at runtime until now

kymatio/scattering1d/frontend/base_frontend.py
```diff
@@ -134,12 +134,17 @@ def output_size(self, detail=False):
...
         if not self.out_type in ('array', 'dict', 'list'):
-            raise RuntimeError("The out_type must be one of 'array', 'dict', or 'list'.")
+            raise ValueError("The out_type must be one of 'array', 'dict'"
+                             ", or 'list'. Got: {}".format(self.out_type))
 
         if not self.average and self.out_type == 'array':
             raise ValueError("Cannot convert to out_type='array' with "
                              "average=False. Please set out_type to 'dict' or 'list'.")
 
+        if self.oversampling<0:
```
cyrusvahidi, line 144: change to `self.oversampling < 0`

tests/scattering1d/test_torch_scattering1d.py
```diff
@@ -439,13 +439,12 @@ def test_Q(device, backend):
         _ = Scattering1D(
             J, shape, Q=0.9, backend=backend, frontend='torch')
     assert "Q should always be >= 1" in ve.value.args[0]
-
+
```
cyrusvahidi, line 442: remove?

lostanlen: my bad. my editor (Atom) added this automatically
6f3715b

MuawizChaudhary (approved):
Will merge

MuawizChaudhary (approved):

Issue #885, opened 2022-06-15
MAINT deprecate `self.average` 1D


cyrusvahidi:
in preparation for the JTFS and 0.4 we could deprecate `average` in `Scattering1D`.

While reusing the 1D frontend for frequency scattering it came to our attention that we could utilize `T` to define the semantics of frequential averaging. This would evade introduction of an additional attribute `F`. Moreover, `T` could potentially cover the semantics of averaging, allowing for removal of `average`:

* `T=None` will perform default mode averaging, i.e. `T=2^J`
* `T=0` will correspond to `average=False`
* Any other integer `T` will stride over the specified support

lostanlen:
> we could utilize T to define the semantics of frequential averaging. This would evade introduction of an additional attribute F

I think `F` would still be nice. The other option would be to make `T` a tuple, but that will collide with #854 

For JTFS we'd have

```python
def __init__(self, J, shape, Q, J_fr, Q_fr, T=None, F=None, oversampling=0, oversampling_fr=0, out_type='array', backend=None)
```

(order of arguments to be debated)

lostanlen:
If we do this, `average` should temporarily default to `None`.

Then we'd have something like this in `__init__`

```python
if average is not None:
    warn('`average`  is deprecated v0.3 and will be removed in v0.4. Replace `average=False` by `T=0` and set `T>1` or leave `T=None` for `average=True` (default)', DeprecationWarning)
self.average = average
```

cyrusvahidi:
> > we could utilize T to define the semantics of frequential averaging. This would evade introduction of an additional attribute F
> 
> I think `F` would still be nice. The other option would be to make `T` a tuple, but that will collide with #854
> 
> For JTFS we'd have
> 
> ```python
> def __init__(self, *, J, shape, Q, J_fr, Q_fr, T=None, F=None, oversampling=0, oversampling_fr=0, out_type='array', backend=None)
> ```

I think it's necessary to have an `F` parameter in the JTFS frontend, but this can be passed to the Scattering1D frontend for frequential averaging as `T`?

lostanlen:
> I think it's necessary to have an `F` parameter in the JTFS frontend, but this can be passed to the Scattering1D frontend for frequential averaging as `T`?

`F` will be passed as `T` to the frequential `scattering_filter_factory`, yes.
But there won't be a `Scattering1D` object for frequential scattering, because @eickenberg doesn't want objects inside other objects.
[...]

PR #897, opened 2022-06-21
MAINT deprecate `average` in `Scattering1D`

cyrusvahidi:
Addressing #885. Deprecation of `Scattering1D` `average` for v0.3, in preparation for removal in v0.4

Alters the semantics of `T`:

- `T=None` performs default mode averaging, i.e. `T=2**J`
- `T=0` is no averaging
- Any other integer`T` will stride over the specified support

### DIFF
--- a/kymatio/scattering1d/frontend/base_frontend.py
+++ b/kymatio/scattering1d/frontend/base_frontend.py
@@ -10,15 +10,14 @@ compute_meta_scattering, precompute_size_scattering)
 
 
 class ScatteringBase1D(ScatteringBase):
-    def __init__(self, J, shape, Q=1, T=None, max_order=2, average=True,
-            oversampling=0, out_type='array', backend=None):
+    def __init__(self, J, shape, Q=1, T=None, max_order=2, oversampling=0, 
+                 out_type='array', backend=None):
         super(ScatteringBase1D, self).__init__()
         self.J = J
         self.shape = shape
         self.Q = Q
         self.T = T
         self.max_order = max_order
-        self.average = average
         self.oversampling = oversampling
         self.out_type = out_type
         self.backend = backend
@@ -65,11 +64,17 @@ class ScatteringBase1D(ScatteringBase):
 
         # check T or set default
         if self.T is None:
-            self.T = 2**(self.J)
+            self.average = True
+            self.T = 2 ** self.J
         elif self.T > N_input:
             raise ValueError("The temporal support T of the low-pass filter "
                              "cannot exceed input length (got {} > {})".format(
                                  self.T, N_input))
+        elif type(self.T) is not int or self.T < 0:
+            raise ValueError("T must be a nonnegative integer (got {})".format(
+                                self.T))
+        else: 
+            self.average = T != 0
         self.log2_T = math.floor(math.log2(self.T))
 
         # Compute the minimum support to pad (ideally)
--- a/kymatio/scattering1d/frontend/base_frontend.py
+++ b/kymatio/scattering1d/frontend/base_frontend.py
@@ -11,7 +11,7 @@ compute_meta_scattering, precompute_size_scattering)
 
 class ScatteringBase1D(ScatteringBase):
     def __init__(self, J, shape, Q=1, T=None, max_order=2, oversampling=0, 
-                 out_type='array', backend=None):
+                 out_type='array', backend=None, average=None):
         super(ScatteringBase1D, self).__init__()
         self.J = J
         self.shape = shape
@@ -21,6 +21,11 @@ class ScatteringBase1D(ScatteringBase):
         self.oversampling = oversampling
         self.out_type = out_type
         self.backend = backend
+
+        if average is not None:
+            warn("`average` is deprecated v0.3 and will be removed in v0.4."
+                 "Replace `average=False` by `T=0` and set `T>1` or leave" 
+                 "`T=None` for `average=True` (default)", DeprecationWarning)
 
     def build(self):
         """Set up padding and filters
@@ -223,9 +228,9 @@ class ScatteringBase1D(ScatteringBase):
             averaged output corresponds to the standard scattering transform,
             while the un-averaged output skips the last convolution by
             :math:`\phi_J(t)`.  This parameter may be modified after object
-            creation. Defaults to `True`. Deprecated in favour of `T` and will 
-            be removed in v0.4. Replace `average=False` by `T=0` and set `T>1` 
-            or leave `T=None` for `average=True` (default).
+            creation. Defaults to `True`. Deprecated in v0.3 in favour of `T` 
+            and will  be removed in v0.4. Replace `average=False` by `T=0` and 
+            set `T>1` or leave `T=None` for `average=True` (default).
         """
 
     _doc_attr_average = \
--- a/kymatio/scattering1d/frontend/base_frontend.py
+++ b/kymatio/scattering1d/frontend/base_frontend.py
@@ -10,8 +10,8 @@ compute_meta_scattering, precompute_size_scattering)
 
 
 class ScatteringBase1D(ScatteringBase):
-    def __init__(self, J, shape, Q=1, T=None, max_order=2, oversampling=0, 
-                 out_type='array', backend=None, average=None):
+    def __init__(self, J, shape, Q=1, T=None, max_order=2, average=None, 
+                 oversampling=0, out_type='array', backend=None):
         super(ScatteringBase1D, self).__init__()
         self.J = J
         self.shape = shape
@@ -75,11 +75,11 @@ class ScatteringBase1D(ScatteringBase):
             raise ValueError("The temporal support T of the low-pass filter "
                              "cannot exceed input length (got {} > {})".format(
                                  self.T, N_input))
-        elif type(self.T) is not int or self.T < 0:
+        elif self.T < 0 or (self.T > 0 and self.T < 1):
             raise ValueError("T must be a nonnegative integer (got {})".format(
                                 self.T))
         else: 
-            self.average = T != 0
+            self.average = self.T != 0
         self.log2_T = math.floor(math.log2(self.T))
 
         # Compute the minimum support to pad (ideally)
--- a/kymatio/scattering1d/frontend/base_frontend.py
+++ b/kymatio/scattering1d/frontend/base_frontend.py
@@ -173,6 +173,15 @@ class ScatteringBase1D(ScatteringBase):
         "or access shape[0] for the unpadded length (previously N).", DeprecationWarning)
         return int(self.shape[0])
 
+    @property
+    def average(self):
+        warn("The average option is deprecated and will be "
+             "removed in v0.4. Please set "
+             "T=None for equivalent functionality to average=False.",
+             "T appropriately for equivalent functionality.",
+             DeprecationWarning)
+        return self.T != 0
+
     _doc_shape = 'N'
 
     _doc_instantiation_shape = {True: 'S = Scattering1D(J, N, Q)',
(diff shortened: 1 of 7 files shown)


cyrusvahidi:
> > * `T=None` performs default mode averaging, i.e. `T=2**J`
> > * `T=0` is no averaging
> > * Any other integer `T` will stride over the specified support
> 
> Why not using `T=2**J` as the default rather than `T=None` (as following)? Having both `T=None` and `T=0` may cause confusion for the case of no averaging?
> 
>     * `T=0`: no averaging
> 
>     * `T=2**J`: default averaging
> 
>     * `T=other positive integer`: other averaging

good idea but you cannot write `def __init__(J, shape, T=2**J ... )`. that would be referencing a variable`J` that is undefined.

changhongw:
> > > * `T=None` performs default mode averaging, i.e. `T=2**J`
> > > * `T=0` is no averaging
> > > * Any other integer `T` will stride over the specified support
> > 
> > 
> > Why not using `T=2**J` as the default rather than `T=None` (as following)? Having both `T=None` and `T=0` may cause confusion for the case of no averaging?
> > ```
> > * `T=0`: no averaging
> > 
> > * `T=2**J`: default averaging
> > 
> > * `T=other positive integer`: other averaging
> > ```
> 
> good idea but you cannot write `def __init__(J, shape, T=2**J ... )`. that would be referencing a variable`J` that is undefined.

Indeed.

kymatio/scattering1d/frontend/base_frontend.py
```diff
@@ -10,19 +10,23 @@
...
         self.max_order = max_order
-        self.average = average
         self.oversampling = oversampling
         self.out_type = out_type
         self.backend = backend
 
+        if average is not None:
+            warn("`average` is deprecated v0.3 and will be removed in v0.4."
+                 "Replace `average=False` by `T=0` and set `T>1` or leave" 
+                 "`T=None` for `average=True` (default)", DeprecationWarning)
```
changhongw, line 28: "Replace `average=False` by `T=0`; for `average=True` , either use the default averaging by `T=None` or customize averaging by setting T to a positive integer" => something like this to make it clearer?

cyrusvahidi:
OK the `average` kwarg is back and is an attribute. `T=0` is only accepted if `average=False`

kymatio/scattering1d/frontend/base_frontend.py
```diff
@@ -23,6 +23,14 @@ def __init__(self, J, shape, Q=1, T=None, max_order=2, average=True,
         self.out_type = out_type
         self.backend = backend
 
+        if not average:
+            warn("The average option is deprecated and will be "
+                 "removed in v0.4."
+                 "For average=True, set T=None for default averaging,"
+                 "or T>1 (positive integer) for custom averaging."
+                 "T=0 with average=False will not average.",
+                 DeprecationWarning)
```
MuawizChaudhary, line 32: `if average is not None`

lostanlen:
@cyrusvahidi can't we default `average` to `None` so that it becomes `False` internally if and only if `T=0`?
With your latest change, the setting `T=0` has no justification because it must be paired with passing `average=False`

the issue raised by @janden was different:
>  if user passes average=False and leaves T=None, we raise a warning and perform averaging. Instead, we should not average

kymatio/scattering1d/frontend/base_frontend.py
MuawizChaudhary: if the user ignores specifying an average that means they've switched over. If `average` is False, the user specified this then.

tests/scattering1d/test_torch_scattering1d.py
```diff
@@ -122,11 +122,12 @@ def test_computation_Ux(backend, device, random_state=42):
     rng = np.random.RandomState(random_state)
     J = 6
     Q = 8
-    T = 2**12
-    scattering = Scattering1D(J, T, Q, average=False,
-                              max_order=1, out_type="dict", frontend='torch', backend=backend).to(device)
+    shape = 2**J
+    scattering = Scattering1D(J, shape, Q, average=False,
+                              max_order=1, out_type="dict", frontend='torch', 
+                              backend=backend).to(device)
```
MuawizChaudhary, line 128: Why this change?

MuawizChaudhary (changes requested):
just one set of changes

cyrusvahidi:
> 

this seems safe to me and also a better way to send the deprecate message. as long as we check `not average and T == 0`. do you agree?

tests/scattering1d/test_torch_scattering1d.py
lostanlen: i approve this. `T` wasn't a good notation here. It's really the shape (length of signal) we're passing as second argument, not the `T` (amount of averaging).

lostanlen:
> this seems safe to me and also a better way to send the deprecate message. as long as we check not average and T == 0. do you agree?

it's certainly safe now but i'd argue it's _too_ safe in the sense that (if i understand well " T=0 is only accepted if average=False") the average argument will remain compulsory (=False) if one wants to disable averaging.

cyrusvahidi:
> > this seems safe to me and also a better way to send the deprecate message. as long as we check not average and T == 0. do you agree?
> 
> it's certainly safe now but i'd argue it's _too_ safe in the sense that (if i understand well " T=0 is only accepted if average=False") the average argument will remain compulsory (=False) if one wants to disable averaging.

yes, too safe. So only error if `T=0` and `average=True`. `average=False` will override any `T`.

MuawizChaudhary (approved):
@lostanlen what do ya think?

lostanlen:
"average=False will override any T" is still not right. Here's what i have in mind, presented as a truth table


| ⬇️ `average` / `T` ➡️ | `None`             | `0`       | `>1`                   | 
|-------|------------------|---------|----------------------|
| `None`  | avg. with `T=2**J` | no avg. | avg. with user-set `T` | 
| `True`  | avg. with `T=2**J` | ERROR   | avg. with user-set `T` | 
| `False` | no avg.          | no avg. | ERROR                | 

* The new default will be `average=None`, `T=None`, and will be backwards compatible.
* Already-in-place settings `average=False` and `T>1` will keep working as intended. The former will raise a `DeprecationWarning`
* Passing `T=0` and letting `average=None` will be the new way to specify `average=False`
* None of the two parameters will override the other. `T=0, average=True` will error, and same for `T>1, average=False`.


We had four possible combinations before. We have nine now, of which two are errors and a four are deprecated.
In the future (v0.4), we will remove `average` and only have three branches: `T=None`, `T=0`, and `T>1`. 

This is a good first step towards supporting `T="global"` in v0.4

lostanlen (changes requested):

kymatio/scattering1d/frontend/base_frontend.py
```diff
@@ -65,11 +69,19 @@ def build(self):
...
         if self.T is None:
-            self.T = 2**(self.J)
+            self._average = True
+            self.T = 2 ** self.J
         elif self.T > N_input:
             raise ValueError("The temporal support T of the low-pass filter "
                              "cannot exceed input length (got {} > {})".format(
                                  self.T, N_input))
+        elif self.T < 0:
+            raise ValueError("T must be a nonnegative integer (got {})".format(
```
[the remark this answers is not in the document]
MuawizChaudhary: Is float T even implemented?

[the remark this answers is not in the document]
lostanlen: yes, (although untested). `T` is used in two places in `scattering_filter_factory`

```
sigma_low = sigma0 / T
```
(float/int or float/float, assigns to float)
so `sigma_low` is float


```
log2_T = math.floor(math.log2(T))
```
(log2(int or float) assigns to float, math.floor of float assigns to int)
so `log2_T` is int

no lizard.

kymatio/scattering1d/frontend/base_frontend.py
```diff
@@ -65,11 +69,19 @@ def build(self):
 
         # check T or set default
         if self.T is None:
-            self.T = 2**(self.J)
+            self._average = True
```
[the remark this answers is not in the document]
cyrusvahidi: i've added some additional logic for this. 

* If `T` is `None`, `average` is set by the user or defaults to `True`
* if `T` is `0`, averaging is only performed if the user explicitly sets `average` to `True`. If `average` is `None` then no averaging is performed.

kymatio/scattering1d/frontend/base_frontend.py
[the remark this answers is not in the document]
cyrusvahidi: > "average=False will override any T" is still not right. Here's what i have in mind, presented as a truth table
> ⬇️ `average` / `T` ➡️ 	`None` 	`0` (int) 	`>=1` (int or float)
> `None` 	avg. with `T=2**J` 	no avg. 	avg. with user-set `T`
> `True` 	avg. with `T=2**J` (`DeprecationWarning`) 	`ValueError` 	avg. with user-set `T` (`DeprecationWarning`)
> `False` 	no avg. (`DeprecationWarning`) 	`ValueError` 	`ValueError`
> 
>     * The new default will be `average=None`, `T=None`, and will be backwards compatible with the old default, without warning nor error.
> 
>     * Passing `T=0` and letting `average=None` will be the new way to specify `average=False`.
> 
>     * Already-in-place settings `average=False` and `T>=1` will keep working as intended. The former will raise a `DeprecationWarning`, instructing the user to replace `average=False` by `T=0` before v0.4.
> 
>     * Passing `average=True` together with `T>=1` or `T=None` will raise a `DeprecationWarning`, instructing the user to remove the `average` argument befor v0.4.
> 
>     * Passing `T=0` together with setting `average=False` will raise a `ValueError`, because this is redundant. The `ValueError` will instruct the user to remove the `average` argument immediately.
> 
>     * If both parameters are set (to non-None values), neither of the two parameters will override the other. `T=0, average=True` will error, and same for `T>=1, average=False`.
> 
> 
> We have four possible combinations as of v0.2. With this PR, we will have nine, of which three are errors and three are deprecated. In the future (v0.4), we will remove `average` and only have (9-3-3) = 3 branches: `T=None`, `T=0`, and `T>=1`.
> 
> I haven't included this in the truth table, but cases `T<0` and `0<T<1` should also lead to `ValueError`'s.
> 
> This is a good first step towards supporting `T="global"` in v0.4

excellent - i've resolved these

kymatio/scattering1d/frontend/base_frontend.py
```diff
@@ -24,8 +24,7 @@ def __init__(self, J, shape, Q=1, T=None, max_order=2, average=True,
         self.backend = backend
 
         if average is not None:
-            warn("The average option is deprecated and will be "
-                 "removed in v0.4."
+            warn("The average option is deprecated and will be removed in v0.4."
                  "For average=True, set T=None for default averaging,"
```
lostanlen, line 28: no comma after averaging

lostanlen (changes requested):

kymatio/scattering1d/frontend/base_frontend.py
```diff
@@ -79,18 +78,21 @@ def build(self):
...
             else:
                 raise ValueError("average must not be True if T=0 " 
                                  "(got {})".format(self.average)) 
+        elif self.T < 1:
+            raise ValueError("T must be nonnegative (got {})".format(
+                             self.T))
         else:
             self.average = True if self.average is None else self.average 
+            if not self.average: 
+                raise ValueError("T must be None or 0 when average=False"
```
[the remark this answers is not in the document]
cyrusvahidi: in this case if `average` is None then indeed no lizard. if `False then there's a value error. Updated the message.

lostanlen (approved):
looks perfect now !!

PR #900, opened 2022-06-23
FIX include 'n' field in 1D/2D list + remove singleton dimension in 1D dict

### LINKED ISSUE #875: Why wouldn't returned dicts with `out_type="list"` not have an `n` field?
Astutely brought up by reviewer @MuawizChaudhary during CR of #862 
https://github.com/kymatio/kymatio/pull/862#discussion_r895089212

In 1D, the `scattering` method ends on a rather strange note:
```python
...
elif out_type == 'list':
    for x in out_S:
        x.pop('n')
return out_S
```

in other words, we're going out of our way to conceal the `n` field to the `out_type="list"` user.
Personally, i don't see a justification for not exposing `n`. It is quite easy to understand and re-use: it can be seen as a "pointer" (or tuple of pointers) to the filterbank `psi1_f` (or tuple of filterbanks `(psi1_f, psi2_f)`)
And `n` is already exposed to the user anyway, under the form of `S.keys()` with `out_type="dict"` (#861). So i don't think we're running into a big risk of introducing a new functionality that will be difficult to maintain in the future.

Hence i propose to experiment with removing this `x.pop('n')` statement. And if the tests pass, make it part of `v0.3`. There's nothing in it that couldn't be achieved with `out_dict="dict"` and functional programming / some advanced looping mechanisms (like `for k,v in S.items()`, but not all our users are familiar with those Pythonic idioms, so giving them access to `n` as the value of a simple data type (dict) contained in a list is a good way to let them write custom code that inspects `psi` metadata n relation with `S` activations.

This change would remove code in 1D and add very little code in 2D/3D.

Thoughts?

### LINKED ISSUE #895: what should be the returned shape of `Scattering1D` paths under `out_type='dict'` ?

lostanlen:
Fix #875, Fix #895
as discussed in today's Kymatio meeting

NB: these changes must be in the release notes.

### DIFF
--- a/kymatio/scattering1d/frontend/numpy_frontend.py
+++ b/kymatio/scattering1d/frontend/numpy_frontend.py
@@ -44,7 +44,6 @@ class ScatteringNumPy1D(ScatteringNumPy, ScatteringBase1D):
                 S[k] = v.reshape(new_shape)
         elif self.out_type == 'list':
             for x in S:
-                x.pop('n')
                 scattering_shape = x['coef'].shape[-1:]
                 new_shape = batch_shape + scattering_shape
                 x['coef'] = x['coef'].reshape(new_shape)
--- a/kymatio/scattering1d/frontend/numpy_frontend.py
+++ b/kymatio/scattering1d/frontend/numpy_frontend.py
@@ -39,7 +39,7 @@ class ScatteringNumPy1D(ScatteringNumPy, ScatteringBase1D):
             for k, v in S.items():
                 # NOTE: Have to get the shape for each one since we may have
                 # average == False.
-                scattering_shape = v.shape[-2:]
+                scattering_shape = v.shape[-1:]
                 new_shape = batch_shape + scattering_shape
                 S[k] = v.reshape(new_shape)
         elif self.out_type == 'list':
--- a/kymatio/scattering1d/frontend/tensorflow_frontend.py
+++ b/kymatio/scattering1d/frontend/tensorflow_frontend.py
@@ -45,7 +45,6 @@ class ScatteringTensorFlow1D(ScatteringTensorFlow, ScatteringBase1D):
                 S[k] = tf.reshape(v, new_shape)
         elif self.out_type == 'list':
             for x in S:
-                x.pop('n')
                 scattering_shape = tf.shape(x['coef'])[-1:]
                 new_shape = tf.concat((batch_shape, scattering_shape), 0)
                 x['coef'] = tf.reshape(x['coef'], new_shape)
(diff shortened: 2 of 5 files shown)


kymatio/scattering1d/frontend/numpy_frontend.py
```diff
@@ -39,12 +39,11 @@ def scattering(self, x):
             for k, v in S.items():
                 # NOTE: Have to get the shape for each one since we may have
                 # average == False.
-                scattering_shape = v.shape[-2:]
```
MuawizChaudhary, line 42: why this change?

MuawizChaudhary: Does it remove the 1 dimension?

MuawizChaudhary (approved):

kymatio/scattering1d/frontend/numpy_frontend.py
lostanlen: yes

MuawizChaudhary:
@changhongw If you approve pls press the big green button!

changhongw:
Thanks for noticing this problem!

PR #877, opened 2022-06-13
Frontend-agnostic 1D scattering

lostanlen:
segue of #871
if merged, will close #870

negative LOC count
this is an important step towards an automated `self.meta()` algorithm

to manage multiple inheritance, i had to move the `raise NotImplementedError` in kymatio's top-level class to 2D and 3D base frontends. Otherwise the backend method takes precedence because it's first in the inheritance pair. I hope that's OK. Feel free to suggest alternatives

PR #908, opened 2022-08-11
Unified scattering1d frontend

lostanlen:
bis of #877, but to main this time

@MuawizChaudhary it's the problem i was telling you about. The PR to main has many more commits than the one i wrote. It also have conflicts


MuawizChaudhary:
wtf is going on here

PR #880, opened 2022-06-14
`core/scattering1d` now returns a generator

lostanlen:
fourth item in a streak, after #871, #877, #879. This concludes the decoupling between core scattering versus pre-processing routines (type checks, reshaping, padding) as well as post-processing routines (reshaping again, padding, and formatting to `out_type`).

The LOC impact is more or less neutral, but the number of input arguments to `core/scattering1d` is significantly reduced:
```python
def scattering1d(U_0, backend, psi1, psi2, phi, oversampling, max_order, average=True):
```

Compare with current master branch:
```python
def scattering1d(x, pad, unpad, backend, J, psi1, psi2, phi, pad_left=0,
        pad_right=0, ind_start=None, ind_end=None, oversampling=0,
        max_order=2, average=True, size_scattering=(0, 0, 0),
        vectorize=False, out_type='array'):
```

This change is backwards compatible so it may happen in v0.3 
The main appeal behind this change is that it allows to envision making a "dry run" of the scattering transform in which `x` is replaced by `None` and the `backend` is replaced by 

```python
class DryBackend:
    __getattr__ = lambda self, attr: (lambda *args: None)
```

This dry run will simplify `self.meta()` and `self.out_size()` methods.

Small new features:
* paths now contain a field `order`, just like the `self.meta()` dictionary
* it would now be very easy to implement `out_type='generator'` for reduced memory footprint. Drink from the `ScatteringBase1D` faucet

PR #879, opened 2022-06-14
Move `pad` from `core` to `base_frontend`

lostanlen:
segue of #877, and moves further in the direction of #855 which closed #852
remove pad from core scattering1d

backwards compatible. opens the door towards custom padding in 1d (#859)
no additional tests
contributes to #735

PR #913, opened 2022-08-11
ENH `core/scattering1d` now returns a generator

lostanlen:
dup of #880 except geared at main branch

Fourth item in a streak, after #871, #877, #879. This concludes the decoupling between core scattering versus pre-processing routines (type checks, reshaping, padding) as well as post-processing routines (reshaping again, padding, and formatting to `out_type`).

The LOC impact is more or less neutral, but the number of input arguments to `core/scattering1d` is significantly reduced:
```python
def scattering1d(U_0, backend, psi1, psi2, phi, oversampling, max_order, average=True):
```

Compare with current master branch:
```python
def scattering1d(x, pad, unpad, backend, J, psi1, psi2, phi, pad_left=0,
        max_order=2, average=True, size_scattering=(0, 0, 0),
        vectorize=False, out_type='array'):
```

This change is backwards compatible so it may happen in v0.3 
The main appeal behind this change is that it allows to envision making a "dry run" of the scattering transform in which `x` is replaced by `None` and the `backend` is replaced by 

```python


class DryBackend:
    __getattr__ = lambda self, attr: (lambda *args: None)
```

This dry run will simplify `self.meta()` and `self.out_size()` methods.

Small new features:
* paths now contain a field `order`, just like the `self.meta()` dictionary
* it would now be very easy to implement `out_type='generator'` for reduced memory footprint. Drink from the `ScatteringBase1D` faucet

### DIFF
--- a/kymatio/scattering1d/core/scattering1d.py
+++ b/kymatio/scattering1d/core/scattering1d.py
@@ -1,12 +1,15 @@
 
-def scattering1d(U_0, backend, psi1, psi2, phi, oversampling, max_order, average):
+def scattering1d(U_0, backend, psi1, psi2, phi, oversampling, max_order, average=True):
     """
     Main function implementing the 1-D scattering transform.
 
     Parameters
     ----------
     U_0 : Tensor
-        a torch Tensor of size `(B, 1, N)` where `N` is the temporal size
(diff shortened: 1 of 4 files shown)


MuawizChaudhary (approved):
It'd be nice to run a time test and see how much the speed changes. @lostanlen please put on your TODO.

PR #914, opened 2022-08-12
ENH automatic self.meta() and output_size in 1D

lostanlen:
dup of https://github.com/kymatio/kymatio/pull/893 except geared at main branch

Continuation of the Scattering1D simplification/decoupling streak: #871, #877, #879, #880.

The `DryBackend()` is now in full swing:

```python
        class DryBackend:
            __getattr__ = lambda self, attr: (lambda *args: None)

        S = scattering1d(None, DryBackend(), self.psi1_f, self.psi2_f,
            self.phi_f, self.oversampling, self.max_order)
        S = sorted(list(S), key=lambda path: (len(path['n']), path['n']))
```

No more calling the `scattering_filter_factory` to get self.meta() !
And no more mimicry of depth-first search in a `utils.py` function. We're now relying on core scattering to perform the depth-first search.

This remains backwards compatible so could be merged for v0.3. The role of this PR was to offer a proof of feasability towards automatic meta in JTFS

Closes #856 

The crux of the PR is here:

```python
S = scattering1d(None, DryBackend(), self.psi1_f, self.psi2_f,
    self.phi_f, self.oversampling, self.max_order)
S = sorted(list(S), key=lambda path: (len(path['n']), path['n']))
meta = dict(order=np.array([len(path['n']) for path in S]))

meta['key'] = [path['n'] for path in S]
meta['n'] = np.stack([np.append(
    path['n'], (np.nan,)*(self.max_order-len(path['n']))) for path in S])
    
filterbanks = (self.psi1_f, self.psi2_f)[:self.max_order]

for key in ['xi', 'sigma', 'j']:
    meta[key] = meta['n'] * np.nan

    for order, filterbank in enumerate(filterbanks):
        for n, psi in enumerate(filterbank):
            meta[key][meta['n'][:, order]==n, order] = psi[key]
            
return meta
 ```

I have moved the legacy implementation (`compute_meta_scattering`) to the test suite so that we can verify backwards compatibility. Here's the legacy implementation for comparison

```python
    sigma_min = sigma0 / math.pow(2, J)
    Q1, Q2 = Q
    xi1s, sigma1s, j1s = compute_params_filterbank(sigma_min, Q1, alpha, r_psi)
    xi2s, sigma2s, j2s = compute_params_filterbank(sigma_min, Q2, alpha, r_psi)

    meta = {}

    meta['order'] = [[], [], []]
    meta['xi'] = [[], [], []]
    meta['sigma'] = [[], [], []]
    meta['j'] = [[], [], []]
[...]

Issue #856, opened 2022-06-08
`compute_params_filterbank` should return `xis` and `sigmas` but not `js`


lostanlen:
came up in conversation with @MuawizChaudhary 

A sequel of #847 : `compute_params_filterbank` should be simpler
At the moment, this function returns three output values: `xis`, `sigmas`, and `js`

But it turns out that every `j` can be derived from `xi` and `sigma` by means of 
 `j = get_max_dyadic_subsampling(xi, sigma, alpha)`
 
Which we could have in the `scattering_filter_factory` itself.
This will avoid carrying over a term back-and-forth between the "factory" and the "params"

in hindsight, this is also beneficial for distinguishing wavelet params (xis, sigmas) from their sampling (js).

janden:
Could work… Would this trickle down to `meta` as well? There we may want to be as verbose as possible given that the users may have different needs.

lostanlen:
>  Could work… Would this trickle down to meta as well? There we may want to be as verbose as possible given that the users may have different needs.

Yes, `self.meta()` would still have the `j`'s. I have a plan for this.

lostanlen:
my plan is implemented in #893

PR #915, opened 2022-08-12
ENH `anden_generator` and `self.filterbank` property

### LINKED ISSUE #851: `ScatteringBase1D` kwargs should be passed via `dict`'s
from the `ScatteringBase1D` frontend to `compute_meta_scattering`, `precompute_size_scattering`, and `scattering_filter_factory`

This `dict` could be named `filterbank_kwargs` and eventually also contain the `spinned` boolean (#807)

This will lead to a reduction in the number of positional arguments in these prototypes. More importantly, it will better decouple hyperparameters from the choice of wavelet shape.

For me to be able to start working on this, i need #833 #842 #848 #850 to be reviewed, rebased, and merged.

### LINKED ISSUE #916: Does anyone use `max_subsampling` in `scattering_filter_factory` 1D ?
This keyword argument defaults to `None` and is documented as:

```
    max_subsampling: int or None, optional
        maximal dyadic subsampling to compute, in order
        to save computation time if it is not required. Defaults to None, in
        which case this value is dynamically adjusted depending on the filters.
```

I don't see it tested anywhere, and it isn't properly supported by the frontends. I doubt that anyone uses it.

Rather, i believe that this is some vestigial scaffold of when we were developing for v0.1

How about we remove it? @cyrusvahidi @changhongw

lostanlen:
Closes #851

also closes #916 

New prototype for `scattering_filter_factory`
```python
scattering_filter_factory(N, J, Q, T, filterbank)
```

where filterbank is stored as a property of the scattering object.

The algorithm for scattering filter factory is now way more concise. No more `if` statements !

```python
# Loop over scattering layers
for Q_layer in Q:
    psi_f = []

    # Loop over center frequencies xi and bandwidths sigma in the filterbank
    for xi, sigma in filterbank_fn(J, Q_layer, **filterbank_kwargs):

        # Construct filter at full resolution
        psi_levels = [morlet_1d(N, xi, sigma)]
        j = get_max_dyadic_subsampling(xi, sigma, **filterbank_kwargs)

        # The idiom min(previous_J, j, log2_T) implements "j1 < j2"
        for level in range(1, min(previous_J, j, log2_T)):
            psi_levels.append(psi_level)
        max_j = max(j, max_j)

    # Keep score of how many resolutions will be needed in the next layer
    previous_J = max_j
    psis_f.append(psi_f)

# Same with low-pass filter phi
sigma_low = filterbank_kwargs["sigma0"] / T
phi_levels = [gauss_1d(N, sigma_low)]
for level in range(1, max(previous_J, 1+log2_T)):
    phi_level = phi_levels[0].reshape(2 ** level, -1).mean(axis=0)
    phi_levels.append(phi_level)
phi_f = {'levels': phi_levels, 'xi': 0, 'sigma': sigma_low, 'j': log2_T}

return tuple([phi_f] + psis_f)
```

- [x] write `scatnet_generator`
- [x] update callsites: `scattering_filter_factory`, `precompute_size_scattering`, `compute_meta_scattering`
- [x] define `generators` and `filters` lists
- [x] pass `alpha` as `filterbank_kwargs`
- [x] pass `sigma0` as `generator_kwargs`
- [x] pass `r_psi` as `generator_kwargs`
- [x] refactor `scattering_filter_factory` for any order
- [x] merge #914 (automatic meta and output_size)
- [x] resolve conflicts here 
- [x] documentation
- [x] tests
- [x] examples

Some of the new features that will become easier to develop and maintain after this:

spinned wavelets
varying J per order
switching wavelet types
varying r_psi per order
varying wavelet type per order
having more than two orders
replacing the scatnet_generator by something else

NB: "skip filters" is out of scope for this PR. This is more simply done with a `DryBackend` (#914)
NB: in the end i didn't introduce `wavelet` and `lowpass` properties, only `filterbank`. I don't want to introduce too many features that perhaps no one will use (although there is precedent for a feature request in switching the wavelet: #385 , but one thing at a time)

I have removed `periodize_filter_fourier` which is a simple one-liner and was called in 2/3 places
`compute_params_filterbank` is now replaced by `scatnet_generator`, which is shorter. I have also improved (i hope!) the documentation


examples/1d/plot_filters.py
```diff
@@ -15,26 +15,27 @@
 # First, we import the `scattering_filter_factory` method, which we will use
 # to generate the filters.
 
-from kymatio.scattering1d.filter_bank import scattering_filter_factory
+from kymatio.scattering1d.filter_bank import scatnet_generator, scattering_filter_factory
```
MuawizChaudhary, line 18: Not a fan of `scatnet`

kymatio/scattering1d/filter_bank.py
```diff
@@ -274,73 +251,78 @@ def compute_xi_max(Q):
     return xi_max
 
 
-def compute_params_filterbank(sigma_min, Q, r_psi=math.sqrt(0.5)):
+def scatnet_generator(J, Q, sigma0, r_psi, **unused_kwargs):
```
MuawizChaudhary, line 254: i dont see why we have to rename this to include `scatnet` 🙄 its the same function, just more of a generator.

kymatio/scattering1d/filter_bank.py
```diff
@@ -351,125 +333,59 @@ def scattering_filter_factory(N, J, Q, T, r_psi=math.sqrt(0.5),
...
     """
-    # compute the spectral parameters of the filters
-    sigma_min = sigma0 / math.pow(2, J)
-    Q1, Q2 = Q
-    xi1s, sigma1s = compute_params_filterbank(sigma_min, Q1, r_psi)
-    j1s = [get_max_dyadic_subsampling(xi1, sigma1, alpha)
-        for xi1, sigma1 in zip(xi1s, sigma1s)]
-    xi2s, sigma2s = compute_params_filterbank(sigma_min, Q2, r_psi)
-    j2s = [get_max_dyadic_subsampling(xi2, sigma2, alpha)
-        for xi2, sigma2 in zip(xi2s, sigma2s)]
```
MuawizChaudhary, line 416: So this bit of code has been 'moved' inside of the generator?

kymatio/scattering1d/filter_bank.py
```diff
@@ -351,125 +333,59 @@ def scattering_filter_factory(N, J, Q, T, r_psi=math.sqrt(0.5),
...
+            psi_levels = [morlet_1d(N, xi, sigma)]
+            j = get_max_dyadic_subsampling(xi, sigma, **filterbank_kwargs)
+
+            # Resample to smaller resolutions if necessary (beyond 1st layer)
+            # The idiom min(previous_J, j, log2_T) implements "j1 < j2"
+            for level in range(1, min(previous_J, j, log2_T)):
+                psi_level = psi_levels[0].reshape(2 ** level, -1).mean(axis=0)
+                psi_levels.append(psi_level)
+            psi_f.append({'levels': psi_levels, 'xi': xi, 'sigma': sigma, 'j': j})
+            max_j = max(j, max_j)
```
MuawizChaudhary, line 377: or has it been moved here?

MuawizChaudhary (approved):
I approve of this PR, not the biggest fan of using `scatnet` as a name tho, I am of the opinion that this kind of verbiage is something the community should consider moving away from.

examples/1d/plot_filters.py
lostanlen: renamed to `anden_generator`

kymatio/scattering1d/filter_bank.py
lostanlen: yes. turns out, we don't need all the xi's and sigma's at once. just one by one to construct the filters.
the previous design was very unwieldy: we were constructig all these equal-sized lists at the top of the function, zipping them, iterating through the zip, and eventually unpacking them to call morlet_1d

making a generator of (xi, sigma) pairs is way nicer in terms of control flow and number of variables

kymatio/scattering1d/filter_bank.py
lostanlen: the `filterbank_fn` yields `xi` and `sigma` in similar ways as `compute_params_filterbank` was returning them.

`get_max_dyadic_subsampling` is here (outside of the generator, see #856 ) because all it needs to know is `(xi, sigma, alpha)`.

kymatio/scattering1d/filter_bank.py
lostanlen: because it's only one of the possible alternatives. in the future i'd like to set up a way to "import" a filterbank specification from another library (librosa for example) and call the Kymatio scattering factory based on that.

Candidates:
HTK, Slaney, ERB, bark, mel, VQT, etc.

PR #918, opened 2022-08-13
FIX set `average=None` in Scattering1D frontends

### LINKED ISSUE #917: `average` should be `None` by default in v0.3
An oversight from #897 which closed #885. Many thanks to @cyrusvahidi for finding it and bringing it up.

Here's the plan.

| ⬇️ `average` / `T` ➡️ | `None`             | `0` (int)      | `>=1`  (int or float)                 | 
|-------|------------------|---------|----------------------|
| `None`  | avg. with `T=2**J` | no avg. | avg. with user-set `T` | 
| `True`  | avg. with `T=2**J` (`DeprecationWarning`)| `ValueError`  | avg. with user-set `T` (`DeprecationWarning`) | 
| `False` | no avg. (`DeprecationWarning`)         | `ValueError` | `ValueError`                | 

* The new default will be `average=None`, `T=None`, and will be backwards compatible with the old default, without warning nor error.
* Passing `T=0` and letting `average=None` will be the new way to specify `average=False`.
* Already-in-place settings `average=False` and `T>=1` will keep working as intended. The former will raise a `DeprecationWarning`, instructing the user to replace `average=False` by `T=0` before v0.4.
* Passing `average=True` together with `T>=1` or `T=None` will raise a `DeprecationWarning`, instructing the user to remove the `average` argument befor v0.4.
* Passing `T=0` together with setting `average=False` will raise a `ValueError`, because this is redundant. The `ValueError` will instruct the user to remove the `average` argument immediately.
* If both parameters are set (to non-None values), neither of the two parameters will override the other. `T=0, average=True` will error, and same for `T>=1, average=False`.

Turns out, we have forgotten one important element: switch to the `average=None` by default! This causes `DeprecationWarning` in current user code even when they don't pass average. Worse, passing `T=0` does not work because it clashes with the implicit `average=True`.

Working on a fix right now. I will need to be backported to v0.3 (summoning the help of @janden / @eickenberg !)

lostanlen:
closes #917

should be merged to both main branch and the v0.3 branch

tagging @eickenberg

### DIFF
--- a/kymatio/scattering1d/frontend/base_frontend.py
+++ b/kymatio/scattering1d/frontend/base_frontend.py
@@ -11,7 +11,7 @@ from ..utils import compute_border_indices, compute_padding
 
 
 class ScatteringBase1D(ScatteringBase):
-    def __init__(self, J, shape, Q=1, T=None, max_order=2, average=True,
+    def __init__(self, J, shape, Q=1, T=None, max_order=2, average=None,
                  oversampling=0, out_type='array', backend=None):
         super(ScatteringBase1D, self).__init__()
         self.J = J
--- a/kymatio/scattering1d/frontend/jax_frontend.py
+++ b/kymatio/scattering1d/frontend/jax_frontend.py
@@ -11,7 +11,7 @@ class ScatteringJax1D(ScatteringJax, ScatteringNumPy1D):
     # be loaded
 
 
-    def __init__(self, J, shape, Q=1, T=None, max_order=2, average=True,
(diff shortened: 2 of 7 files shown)


cyrusvahidi (approved):

cyrusvahidi (approved):

cyrusvahidi:
very small change that was an oversight in the original commit. Now the deprecation warning is only raised if user explicitly sets average to bool.

PR #919, opened 2022-08-13
ENH `T='global'`: global average pooling in `Scattering1D`

### LINKED ISSUE #902: `T='global'` (global average pooling)
Now that we have a `T` parameter (#673) and that `average` is deprecated (#884), i think it's time we consider passing `T='global'` to the constructor so that it skips the `phi` and performs a 1D uniform average, also known as "global average pooling", instead. This will be faster (`O(N)` instead of `O(N log N)`) and conceptually simpler than the default `T=None` (i.e., `2**J`).

The only drawback is that this implies adding a new Kymatio primitive, namely `sum`:

`def sum(x, axis)`
where `axis` is an `int` (for 1D) or `tuple` of `int`'s (for 2D/3D)

lostanlen:
and remove deprecated `average` kwarg

Closes #902 

Rationale:

> Now that we have a `T` parameter (#673) and that `average` is deprecated (#884), i think it's time we consider passing `T='global'` to the constructor so that it skips the `phi` and performs a 1D uniform average, also known as "global average pooling", instead. This will be faster (`O(N)` instead of `O(N log N)`) and conceptually simpler than the default `T=None` (i.e., `2**J`).

Note that this implementation does _not_ introduce new keyword arguments to the Scattering1D constructor. Rather, it special-cases a new string value for `T`: `'global'`. This is backwards compatible. Local averaging `T=2**J` remains and will remain the default.

To implement this, i have added a new Kymatio primitive, namely `sum`:

`def sum(x, axis)`
where `axis` is an `int` (for 1D) or `tuple` of `int`'s (for 2D/3D)

Thanks to the new generator of core implementation (#880), i am able to skip unpadding when performing global averaging. So this should be faster and memory-efficient, although i haven't formally tested it in practice.

I have modified the internal values of `average` and the prototype of `core`, but this isn't anything the user accesses directly, so i guess it's fine to break things.
(in fact we've been warning the user against accessing average for several months now. and core is now wrapped inside a unified frontend #908 so i don't see why users would call it directly)

I have added a simple NumPy test, feel free to request more if you have ideas or specific use cases.


cyrusvahidi (approved):

tests/scattering1d/test_numpy_scattering1d.py
```diff
@@ -101,6 +101,20 @@ def test_Scattering1D_filter_factory_T(self, backend):
                     N, J, Q, T, filterbank)
                 assert (phi_f['sigma'] == 0.1/T)
 
+    def test_Scattering1D_average_global(self, backend):
```
cyrusvahidi, line 104: let's do this for the other backends as well since there's the new `sum` primitive

cyrusvahidi (changes requested):
add backend tests

tests/scattering1d/test_numpy_scattering1d.py
lostanlen: you were right, the other backends didn't work as intended !
thanks ...

i have written a fix + more tests in [bbab64b](https://github.com/kymatio/kymatio/pull/919/commits/bbab64b274152586938f7054e3a447d49aacea55)

cyrusvahidi (changes requested):
LGTM on the testing side but conflicts have been introduced since recent merges. Fix and we go.

cyrusvahidi (approved):

cyrusvahidi:
Tested new backend primitive. Introduced `average_local` to `scattering1d`. Rebasing resolved.

PR #921, opened 2022-08-13
MAINT simpler 1d prototype

### LINKED ISSUE #920: core scattering doesn't need max_order as input argument
instead, we could pass
`[phi_f, psi1_f]` to signify `max_order==1` 

and `[phi_f, psi1_f, psi2_f]` to signify `max_order==2`

fewer input arguments and potentially a greater consistency with JTFS

(which won't have a `max_order` but rather a `filters` list and a `filters_fr` list)

lostanlen:
closes #920 

fewer input arguments to core scattering
fewer LOCs
backwards compatible

the `scattering1d` prototype is a one-liner now ! :)

a more ambitious change would have been to store `filters` as an attribute of the Scattering1D object and expose `phi_f` and friends as read-only properties. and indeed this is what i plan to do in JTFS
but it's equivalent from a usability standpoint (at least so long as we restrict Scattering1D to one or two orders) so i don't see a compelling reason to do it now.

### DIFF
--- a/kymatio/scattering1d/core/scattering1d.py
+++ b/kymatio/scattering1d/core/scattering1d.py
@@ -1,5 +1,5 @@
 
-def scattering1d(U_0, backend, psi1, psi2, phi, oversampling, max_order, average=True):
+def scattering1d(U_0, backend, filters, oversampling, average=True):
     """
     Main function implementing the 1-D scattering transform.
 
@@ -44,6 +44,7 @@ def scattering1d(U_0, backend, psi1, psi2, phi, oversampling, max_order, average
     U_0_hat = backend.rfft(U_0)
 
     # Get S0
+    phi = filters[0]
     log2_T = phi['j']
     k0 = max(log2_T - oversampling, 0)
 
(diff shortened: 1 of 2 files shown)


kymatio/scattering1d/core/scattering1d.py
```diff
@@ -39,59 +39,55 @@ def scattering1d(U_0, backend, psi1, psi2, phi, oversampling, max_order, average
...
-            S_1_r = irfft(S_1_hat)
+            S_1_c = backend.cdgmm(U_1_hat, phi['levels'][k1])
+            S_1_hat = backend.subsample_fourier(S_1_c, 2**k1_J)
+            S_1_r = backend.irfft(S_1_hat)
             yield {'coef': S_1_r, 'j': (j1,), 'n': (n1,)}
         else:
             yield {'coef': U_1_m, 'j': (j1,), 'n': (n1,)}
 
-        if max_order == 2:
+        if len(filters) > 2:
```
cyrusvahidi, line 88: clean 🔥

cyrusvahidi (approved):
vraiment bon hygiene. LGTM. on y va!

Issue #645, opened 2021-01-21
kymatio for complex input signals


Buffy-nevc:
Hi,

Is Kymatio able to work with complex signals, given that the wavelets are already analytic wavelets. Essentially, the constant-Q fitlers banks need to cover the whole digital frequency to 1, instead of 0.5. 

Thank you. 

Buffy

edouardoyallon:
Hi,
Thanks for asking, this is a great question. And the answer is not that obvious. You'll need to cover indeed negative frequencies. I feel your question is more related to 1d-signals, and I'll et @lostanlen or @janden speak up, yet in term of 2D, this is not native:

1. The 2D low-pass filter. It's likely that then we will be willing to use a modulus there in order to output real-valued scattering coefficients here. Yet in vision, modulus are not desirable.
2. Currently, the WT is exploiting the hermitian symmetry, meaning that we would lack half of the wavelet coefficients if we were using complex signals. This should be an easy fix.
3. In roto-translation (which is close to 3D), implemented like in https://github.com/scatnet/scatnet , the current WT is not correct (except if a fix has been added). The maison reason is that the composition of two analytic WTs (one along space, one along angles) isn't analytic. You need at least one of them to cover the full spectrum and it can be simply viewed as follow: one 2D analytic WT combined with an analytic angular WT covers only 1/4th of the spectral space. I suspect that this might affect positively many benchmarks.

Buffy-nevc:
Hi edouardoyallon,

Thanks for your reply. Yes, my question was more directed at 1D applications. Any information on the constraints imposed on the wavelet designs would be greatly appreciated.

Does Kymatio has any near-time plans to incorporate this feature into the library? It will make the library more applicable for much wider range of applications. 

Thanks,

edouardoyallon:
Hi Buffy,

> Does Kymatio has any near-time plans to incorporate this feature into the library? It will make the library more applicable for much wider range of applications.

@lostanlen or @janden might be more amenable to answer.

In general, we'd love to add new features but as you know those take time (and we're all working on adding features on our spare time). However, if you're interested, I'd be happy to guide you to propose a PR to implement this feature. Send me an email if you're up to.

lostanlen:
Hello @buffy-nevc,

This is a good feature request indeed.

I have been thinking about the issue (in 1-D) and it seems relatively straightforward, although i haven't tried it.
[...]

PR #922, opened 2022-08-13
ENH Spinned wavelets

lostanlen:
For JTFS
may also serve #645 if there's interest in supporting complex-input signals in the future

very small PR. most of the interface for this was made possible by the `anden_generator` #915
no extra kwarg to the 1D factory :)

thanks to @cyrusvahidi and @MuawizChaudhary who had a first working solution for this back in May

will document this after code review


cyrusvahidi:
I like this. We get spinned wavelet filter parameters as and when they are required.

cyrusvahidi (approved):
LGTM

PR #923, opened 2022-08-13
ENH core width-first 1d scattering for JTFS

lostanlen:
first step of the JTFS algorithm

Prototype:
```python
scattering1d_widthfirst(U_0, backend, filters, oversampling, average_local)
```

yields: U0, S1 (concatenated), Y2 (concatened) if average_local is True
yields: U0, U1 (per n1), Y2 (concatenated) if average_local is false

same as scattering1d once #921 is merged

the unit test makes sure that depth-first and width-first versions of the time scattering algorithm lead to the same outcome

advice to code reviewers: it's worth to look at `core/scattering1d.py` and `core/timefrequency_scattering.py` modules in parallel to spot the meaningful similarities and differences

will document after CR is done and conflicts with other branches are resolved. the docstring will be very close to that of existing scattering1d (depth-first)

### DIFF
--- a/kymatio/scattering1d/core/timefrequency_scattering.py
+++ b/kymatio/scattering1d/core/timefrequency_scattering.py
@@ -1,4 +1,4 @@
-def scattering1d_depthfirst(U0, backend, filters, oversampling, average_local):
+def scattering1d_widthfirst(U0, backend, filters, oversampling, average_local):
     # compute the Fourier transform
     U_0_hat = rfft(U_0)
 
@@ -34,7 +34,7 @@ def scattering1d_depthfirst(U0, backend, filters, oversampling, average_local):
 
         if average_local:
             U_1_hat = rfft(U_1_m)
-            # Depth-first algorithm: store U1 in anticipation of next layer
+            # Width-first algorithm: store U1 in anticipation of next layer
             U_1_hats.append({'coef': U_1_hat, 'j': (j1,), 'n': (n1,)})
 
             # Convolve with phi_J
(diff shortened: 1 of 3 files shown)


lostanlen:
Jenkins fails
something goes awry in concatenate when backend is GPU PyTorch

maybe something related to the last dimension being real/imaginary parts?

i'm going to need the help of a backend expert here ... 😕

kymatio/scattering1d/core/timefrequency_scattering.py
```diff
@@ -0,0 +1,87 @@
...
+        else:
+            # U_1_m is a 2D array indexed by (batch, time)
+            yield {'coef': U_1_m, 'j': (j1,), 'n': (n1,)}
+
+    if average_local:
+        # Concatenate S1 paths over the penultimate dimension with shared n1.
+        # S1 is a real-valued 3D array indexed by (batch, n1, time)
+        S_1 = backend.concatenate(S_1_list)
+
+        # S1 is a stack of multiple n1 paths so we puth (-1) as placeholder.
```
cyrusvahidi, line 55: "so we put..."

cyrusvahidi (commented):
i've done a first casual pass.  good to have decoupled the width first and test against 1D depth-first. will do a more in-depth review of the tests shortly.

Needs rebasing onto #919 since we introduced `average_local`.

cyrusvahidi (commented):
`average_local` bon

lostanlen:
Rebased against #918 #919 #921 #922
Tests pass locally

MuawizChaudhary (approved):
It looks good to me. Great that we check against our current 1d implementations

kymatio/scattering1d/core/timefrequency_scattering.py
lostanlen: thanks. fixed here [2fa46cb](https://github.com/kymatio/kymatio/pull/923/commits/2fa46cbcba9dd466641338dffc3b6164f24b6405)

lostanlen:
wrote the developer's notes

cyrusvahidi (approved):

cyrusvahidi (approved):

PR #808, opened 2022-05-17
ENH Transpose backend primitive 

MuawizChaudhary:
This PR adds the transpose backend primitive for all backends (Torch, Tensorflow, Numpy) in 1d scattering, alongside some tests. 

Important to note, this backend primitive only transposes two dimensions. In Tensorflow and Numpy, we transpose the final two dimensions. In Torch, we transpose the second to last and third to last dimensions. 

We should be sure to note the above in some documentation. 

We need to merge pr #807 prior to merging this PR.

PR #925, opened 2022-08-14
ENH JTFS `swap_time_frequency`

lostanlen:
For JTFS

Initially authored by @cyrusvahidi and @MuawizChaudhary back in May (#808)
big thanks to y'all

### DIFF
--- a/kymatio/scattering1d/backend/numpy_backend.py
+++ b/kymatio/scattering1d/backend/numpy_backend.py
@@ -104,4 +104,10 @@ class NumpyBackend1D(NumpyBackend):
 
         return cls._np.sum(x, axis=-1, keepdims=True)
 
+    @classmethod
+    def swap_time_frequency(cls, x):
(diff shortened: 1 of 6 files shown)


cyrusvahidi (commented):
All cool. LGTM. Are we making executive decision to omit JAX backend for the first stages of JTFS development?

lostanlen:
jax is included by way of `numpy_backend`


```python
class JaxBackend1D(JaxBackend, NumpyBackend1D):
    # JaxBackend1D inherits _np and _fft and name attributes from JaxBackend
    # The rest comes from NumpyBackend1D
    # and from NumpyBackend via both JaxBackend and NumpyBackend1D

    # The modification of _np and _fft are sufficient to 
    # drop in jax.numpy instead of numpy. This works because
    # all the numpy commands have an equivalent in jax.numpy

    pass
```


backend = JaxBackend1D

PR #927, opened 2022-08-14
MAINT `parse_T` in 1d utils

lostanlen:
i propose to put the self.T/self.average logic from #740 and friends into a dedicated `utils.py` function.
Advantages:
* `ScatteringBase1D.build` is more concise
* better documentation of what we're doing there. It's a 6-fold case switch so, best to be clear there
* dedicated unit tests
* will be re-usable for JTFS by passing `T_alias='F'`. Less boilerplate code


cyrusasfa (approved):
Cool. Comprehensive tests. I can see how this part of the codebase will grow and duplicate with JTFS. no lizard. LGTM. vamos.

PR #932, opened 2022-08-15
FIX JTFS `average_local=False` in `scattering1d_widthfirst`

lostanlen:
So, i made a benign mistake in #923: i implemented `average_false` as yielding `U0, U1, Y2`. 
But instead it should yield `U0, S1, Y2`. 

Indeed `S1` will be passed through frequential scattering to yield `U2_phi_psi = | S1 * psi_fr |` (conv over n1, no spin)

This simplifies the logic: U1 is always averaged into S1, no matter what is the value of `average_local`. I have rewritten the tests against the depth-first implementation

### DIFF
--- a/kymatio/scattering1d/core/timefrequency_scattering.py
+++ b/kymatio/scattering1d/core/timefrequency_scattering.py
@@ -1,4 +1,4 @@
-def scattering1d_widthfirst(U_0, backend, filters, oversampling, average_local):
+def scattering1d_widthfirst(U_0, backend, filters, oversampling):
     """
     Inputs
     ------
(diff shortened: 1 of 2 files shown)


cyrusvahidi (approved):
Looks good to me. I understand the difference between depth-first 1D scattering and what is required width-first prior to frequential scattering and averaging.

lostanlen:
Thanks for merging. Yes, this implementation is specific to JTFS, it cannot serve as a drop-in replacement for depth-first `scattering1d`. Doing this would require an average of the first layer, a modulus+average of the second layer, and splitting the concatenated n1 paths back into 2D (batch, time) arrays. But that's beyond the scope of this PR.

PR #936, opened 2022-08-17
ENH JTFS `_check_runtime_args`

lostanlen:
The beginning of `TimeFrequencyScatteringBase.scattering` will be:

```python
def scattering(self, x):
    TimeFrequencyScatteringBase._check_runtime_args(self)
    TimeFrequencyScatteringBase._check_input(self, x)

    x_shape = self.backend.shape(x)
    batch_shape, signal_shape = x_shape[:-1], x_shape[-1:]
    x = self.backend.reshape_input(x, signal_shape)
    U_0 = self.backend.pad(x, pad_left=self.pad_left, pad_right=self.pad_right)
```

This PR adds the checks for `average_fr`, `oversampling_fr`, and `out_types`.
the out_types i am anticipating are:
* 2D: batch, path, time (just like Scattering1D). includes zeroth order
* 3D: batch, second-order path, frequency, time. excludes zeroth order
* array: tries 3D if possible, otherwise falls back to 2D (this will be the default)
* dict: keys will be () at zeroth order, (n1,) at first order, and (n1, (n2, n_fr)) at second order. the unspinned phi-psi pair will be represented by n2=-1.
* list: all paths concatenated, including metadata.

2D requires T≠0 (i.e. some form of averaging, whether local or global)
3D requires both T≠0 and F≠0.

~Note that i am not anticipating `oversampling='full'` or `oversampling_fr='full'`, which may allow 3D concatenation despite T=0 and/or F=0. This might come as a later feature if someone requests it, but i don't want to complicate the runtime logic too much at this stage.~ (see #937)

### DIFF
--- a/kymatio/scattering1d/frontend/base_frontend.py
+++ b/kymatio/scattering1d/frontend/base_frontend.py
@@ -588,5 +588,25 @@ class TimeFrequencyScatteringBase(ScatteringBase1D):
             "alpha": self.alpha, "r_psi": self.r_psi, "sigma0": self.sigma0}
         return spin(anden_generator, filterbank_kwargs)
 
+    def _check_runtime_args(self):
+        super(TimeFrequencyScatteringBase, self)._check_runtime_args(
+            out_types=['array', 'dict', 'list', '2D', '3D']
+        )
+
+        if not self.average and self.out_type in ['array', '2D', '3D']:
+            raise ValueError("Cannot convert to out_type='{}' with T=0. "
+                "Please set out_type to 'dict' or 'list'.".format(self.out_type))
+
+        if self.oversampling_fr < 0:
+            raise ValueError("oversampling_fr must be nonnegative. "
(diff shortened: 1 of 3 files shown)


lostanlen:
Upon reflection, i think i'd like to stick with the same semantics of `out_type` as in the rest of the library: namely, `array`, `dict`, `list`. No `2D`/`3D` values.

And instead, introduce a separate keyword for managing the array formatting of paths.
Could be `format='time'` versus `format='joint'` for example.

Curious to know the opinion of @cyrusvahidi @MuawizChaudhary @changhongw among others

lostanlen:
@cyrusvahidi well here i'm passing `out_types=['array', 'dict', 'list', '2D', '3D']` so there is a decision. And i'm not sure anymore it's the right one to make.

lostanlen:
@cyrusvahidi said

> Let’s keep out_type semantics the same for now

and i think he's right

Issue #937, opened 2022-08-17
Should we keep `oversampling`?


lostanlen:
## What is `oversampling` ?
We've had `oversampling` in 1D, but not 2D nor 3D, since 2018.

Now that the API of `Scattering1D` is solidifying, i think we should re-discuss as to whether it's the right abstraction for the user. Here's the current docstring (which i wrote, BTW):

```python
    oversampling : int >=0, optional
        if average_local is True, return scattering coefficients at the sample
        rate max(1, 2**(log2_T-oversampling)). Hence, raising oversampling by
        doubles the sample rate, until reaching the native sample rate.
        if average_local is False, return first-order coefficients at the sample
        rate max(1, 2**(j1-oversampling)) and second-order coefficients at the
        sample rate max(1, 2**(j2-oversampling)).
```

Quite finicky!

## `oversampling>0` complicates the interpretation of `T`

If the user sets T, the stride (🎩  #673 ) is not `T` but `max(1, 2**(log2_T-oversampling))`.

I've already raised attention about this in https://github.com/kymatio/kymatio/pull/887#issuecomment-1157257477

> [... ]i'll admit i don't love the current semantics. alpha also achieves a form of oversampling when getting the j's from xi's and sigma's. It's also a bit surprising that the stride (if average) is not T 

T does not guarantee a low-pass width either because that also depends upon `self.sigma0`. 

## `oversampling>0` post-#913 triggers a bug

Despite its central role, it's largely untested, except for very basic API checks i wrote recently (#887).
This had caused a bug to slip through in #913, merged 5 days ago:

```python
if self.average:
    res = self.log2_T
```

This should have read as 

```python
if self.average:
    res = max(self.log2_T - self.oversampling, 0)
```

and causes an unpadding error (and thus array concatenation error) in the non-default case `self.oversampling > 0`. The bug wasn't caught by unit tests because they all work under `self.oversampling=0`

I'd like to hear from everyone as to what should be the appropriate response beyond this specific bugfix?

This is an important decision because it propagates through JTFS (and doubly so, because of `oversampling_fr`).

## What i propose: deprecate `self.oversampling` and introduce `log2_stride`

Right now, one thing we could do would be to define a property called `stride` and which will return `max(1, 2**(log2_T-oversampling))`. 

[...]

PR #938, opened 2022-08-17
FIX `oversampling>0` in `ScatteringBase1D.scattering`

lostanlen:
Triggers and fixes a bug in `oversampling>0` (non-default), introduced by #937 a few days ago.


lostanlen:
Great, it's failing 👍🏻 
now adding the fix

cyrusvahidi (approved):

PR #944, opened 2022-08-23
ENH JTFS `pad_frequency`

lostanlen:
The JTFS torch frontend requires a (very) special case for frequential padding so i made a small Kymatio 1D primitive for it. See docstring

Tests included


cyrusvahidi (approved):
I get you here. LGTM.

PR #945, opened 2022-08-23
ENH JTFS `time_averaging`

lostanlen:
A small PR which implements the mapping `U2 -> S2` in JTFS.
Tests and documentation are included

### DIFF
--- a/kymatio/scattering1d/core/timefrequency_scattering.py
+++ b/kymatio/scattering1d/core/timefrequency_scattering.py
@@ -23,7 +23,10 @@ def joint_timefrequency_scattering(U_0, backend, filters, oversampling,
           to a low-pass filter of width 2**J_fr and satisfies xi=0, i.e, spin=0.
           Other elements, such that n_fr>0, correspond to "spinned" band-pass
           filter, where spin denotes the sign of the center frequency xi.
-    oversampling_fr : int >= 0. same role as in scattering1d
+    oversampling_fr : int >=0
+        Yields joint time-frequency scattering coefficients with a frequential
+        stride max(1, 2**(log2_F-oversampling_fr)). Raising oversampling_fr
+        by one halves the stride, until reaching a stride of 1.
     average_local_fr : boolean
         whether the result will be locally averaged with phi after this function
 
@@ -88,11 +91,11 @@ def time_scattering_widthfirst(U_0, backend, filters, oversampling, average_loca
     U_0 : array indexed by (batch, time)
     backend : module
     filters : [phi, psi1, psi2] list of dictionaries. same as scattering1d
-    oversampling : int >=0, optional
+    oversampling : int >=0
         Yields scattering coefficients at the sample
         rate max(1, 2**(log2_T-oversampling)). Hence, raising oversampling by
         doubles the sample rate, until reaching the native sample rate.
-    average_local : boolean, optional
+    average_local : boolean
         whether to locally average the result by means of a low-pass filter phi.
 
     Yields
@@ -213,7 +216,10 @@ def frequency_scattering(X, backend, filters_fr, oversampling_fr,
           to a low-pass filter of width 2**J_fr and satisfies xi=0, i.e, spin=0.
           Other elements, such that n_fr>0, correspond to "spinned" band-pass
           filter, where spin denotes the sign of the center frequency xi.
-    oversampling_fr : int >= 0. same role as in scattering1d
+    oversampling_fr : int >=0
+        Yields joint time-frequency scattering coefficients with a frequential
+        stride max(1, 2**(log2_F-oversampling_fr)). Raising oversampling_fr
+        by one halves the stride, until reaching a stride of 1.
     average_local_fr : boolean
         whether the result will be locally averaged with phi after this function
     spinned: boolean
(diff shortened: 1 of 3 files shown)


kymatio/scattering1d/core/timefrequency_scattering.py
```diff
@@ -13,7 +13,7 @@ def joint_timefrequency_scattering(U_0, backend, filters, oversampling,
         Yields scattering coefficients at the sample
         rate max(1, 2**(log2_T-oversampling)). Hence, raising oversampling by
         doubles the sample rate, until reaching the native sample rate.
-    average_local : boolean, optional
+    average_local : boolean
```
changhongw, line 16: not optional any more? always need to be specified by the user whether to average?

kymatio/scattering1d/core/timefrequency_scattering.py
```diff
@@ -268,3 +276,45 @@ def frequency_scattering(X, backend, filters_fr, oversampling_fr,
...
+
+
+def time_averaging(U_2, backend, phi_f, oversampling):
+    """
+    Parameters
+    ----------
+    U_2 : dictionary with keys 'coef' and 'j', typically returned by
+        frequency_scattering
+    backend : module
+    phi_f : dictionary. Temporal low-pass filter in Fourier domain,
```
changhongw, line 288: Would the `phi_f` here introduce confusion with the frequential low-pass filter?

changhongw (approved):

kymatio/scattering1d/core/timefrequency_scattering.py
cyrusvahidi: agree, is this a typo?

tests/scattering1d/test_timefrequency_scattering.py
```diff
@@ -252,6 +253,15 @@ def test_joint_timefrequency_scattering():
         avg_value = np.mean(np.abs(path['coef'][midpoint, :]))
         assert avg_value < 1e-5
 
+        # Test time averaging
+        S.average = 'local'
+        U_2 = {**path, 'coef': backend.modulus(path['coef'])}
+        S_2 = time_averaging(U_2, backend, S.phi_f, S.oversampling)
```
cyrusvahidi, line 259: why `phi_f`?

kymatio/scattering1d/core/timefrequency_scattering.py
lostanlen: it's optional in the frontend, it's compulsory in core

kymatio/scattering1d/core/timefrequency_scattering.py
lostanlen: This is not a typo. The `_f` means fourier domain, but it's still associated to the time variable, not the log-frequency variable. This is for consistency with scatteringd1d

for frequential scattering i use the suffix `_fr`.
See `J_fr`, `Q_fr`, `_N_padded_fr`, `filters_fr`

So the frequential filters have both suffixes: `phi_fr_f` and `psi_fr_f`

tests/scattering1d/test_timefrequency_scattering.py
lostanlen: see above

kymatio/scattering1d/core/timefrequency_scattering.py
changhongw: I see.

kymatio/scattering1d/core/timefrequency_scattering.py
changhongw: Make sense! Documenting this in the code may also help the user.

lostanlen: these docstrings in core are more "developer's notes" (aimed at people like cyrus / lily / muawiz / you / etc.) than user's notes per se. Ordinary users do not call core Kymatio functions directly but via frontends instead.

This is why core/scattering1d.py does not have a docstring (yet). But for JTFS, the core algorithm is fairly complicated and has several functions so i figured it would be helpful for maintainers and reviewers to have these docstrings as self-help.

changhongw: Good point. Thank you!

kymatio/scattering1d/core/timefrequency_scattering.py
```diff
@@ -268,3 +282,45 @@ def frequency_scattering(X, backend, filters_fr, oversampling_fr,
...
+        same as scattering1d
+    oversampling : int >=0
+        Yields scattering coefficients at the sample
+        rate max(1, 2**(log2_T-oversampling)). Hence, raising oversampling by
+        doubles the sample rate, until reaching the native sample rate.
```
changhongw, line 299: - `Raising oversampling by doubles` => missing `one`.

- `max(1, 2**(log2_T-oversampling))` is the number of samples per stride rather than "sample rate"? Raising oversampling by one doubles the sample rate but halves `2**(log2_T-oversampling)`.

lostanlen: thank you. you're right

PR #972, opened 2022-11-02
make swap_time_frequency output contiguous

cyrusvahidi:
* `swap_time_frequency` backend primitive explicitly returns contiguous tensor
* add contiguity test


MuawizChaudhary:
i approve

MuawizChaudhary (approved):

MuawizChaudhary:
can you add an issue mentioning that we need to call contiguous? it makes it so that we clone the tensor, and that's in efficient. let's try to discuss an alternative

PR #979, opened 2022-12-08
FIX #978 omit order0 from JTFS meta when `format="joint"`

### LINKED ISSUE #978: JTFS `meta['order']` contains an off-by-one path index
When doing JTFS with `format="joint"` it seems that there is an off-by-one index in `meta['order']`

```
import torch
from kymatio.torch import TimeFrequencyScattering

jtfs_kwargs = {'shape': 2**15, 'Q': (12, 1), 'J': 8, 'J_fr': 6, 'T': 2**6, "F": 1}
jtfs = TimeFrequencyScattering(**jtfs_kwargs)
Sx = jtfs(torch.zeros(2**15))
```

The above code produces an out size of `torch.Size([98, 64, 512])`, while there is a `98` in `.meta()['order']`.

For 3D aligned format, we skip order 0, for obvious reasons: https://github.com/kymatio/kymatio/blob/3f7d02518f111e7c127da8d97e65de5f1a492a81/kymatio/scattering1d/frontend/base_frontend.py#L643 

However, it seems we forgot to omit order 0 from meta in the case that `format=="joint" and out_type=="array"`: https://github.com/kymatio/kymatio/blob/3f7d02518f111e7c127da8d97e65de5f1a492a81/kymatio/scattering1d/frontend/base_frontend.py#L668

A simple condition to check the format and out_type fixes this:

```
 if len(path['n']) == 0:
        # If format='joint' and out_type='array' skip zeroth order
        if not (self.format == 'joint' and self.out_type == 'array'):
            # add order0 meta 
            ...
```

cyrusvahidi:
Fixes #978

Added minimal test to check that `meta['order']` has same number of elements as Sx paths

### DIFF
--- a/tests/scattering1d/test_timefrequency_scattering.py
+++ b/tests/scattering1d/test_timefrequency_scattering.py
@@ -1,3 +1,5 @@
+import os
+
 import numpy as np
 import pytest
 import torch, tensorflow as tf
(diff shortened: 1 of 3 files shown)


tests/scattering1d/test_timefrequency_scattering.py
```diff
@@ -548,6 +548,10 @@ def test_jtfs_numpy_and_sklearn(frontend):
     Sx = S(x)
     assert Sx.ndim == 3
 
+    # format='joint' meta()
+    meta = S.meta()
+    assert len(meta["order"]) == Sx.shape[-3]
```
lostanlen, line 553: sweet

lostanlen (approved):

PR #983, opened 2023-01-08
FIX `_reduction=np.sum` (vs default `np.mean`) in JTFS sc filter factory

### LINKED ISSUE #984: BUG 1D filter resolutions introduces energy issue at octaves
While developing JTFS, we observed some artefacts that appeared around octave boundaries. Moreover, we observed that the energy of a signal that should have constant energy across `n1` decays with increasing `j`. 

This holds for `Scattering1D` as well as JTFS and is exemplified by visualization of the S1 of an dirac or chirp:

```
from kymatio.torch import Scattering1D
import librosa 

def plot_spec(Sx, S, sr=22050):
    x_coords = librosa.times_like(Sx, hop_length=S.T)
    y_coords = [psi["xi"]*sr for psi in S.psi1_f]

    plt.figure(figsize=(8, 4))
    librosa.display.specshow(Sx[1:,:].numpy(), sr=sr,
        x_coords=x_coords, x_axis="time",
        y_coords=y_coords, y_axis="cqt_hz",
        cmap="magma")
    plt.xlabel("Time (seconds)")
    plt.ylabel("Frequency (Hz)")
    plt.ylim(0, sr // 2)
    # plt.yticks([100, 200, 400, 800, 1600, 3200, 6400])
    plt.xticks(np.arange(0, 1.2, 0.2))
    plt.minorticks_off()

N = 2**15
kwargs = {'shape': N, 'Q': (12, 1), 'J': 8}
S = Scattering1D(**kwargs, max_order=1)

x = torch.zeros(N)
x[N // 2] = 1

Sx = S(x)

plot_spec(Sx, S, sr=2**15)
```

image

image

And the spectrum at the time slice centred around the impulse:

image

Which appears to decay by a factor of 2 at octave boundaries ... an energy decay by roughly `2**k1`

This discontinuity at octave boundaries did not cause problems at second-order for `Scattering1D`, but is an issue for second-order of JTFS because we do wavelet conv across octaves. 

The chirp below covers 3 octaves: 

image

What results are some energy artifacts exactly at the octave boundaries:

image

We identified some relationship between this and `T`. We tried normalizing the wavelet filters to have constant peak. This worked for computing U1, but not S1. 

We realzing that when we create resolutions of `psi` and `phi`, we subsample by taking a mean, but across different signal lengths. This introduces an energy decay correlated with `2**level`.
https://github.com/kymatio/kymatio/blob/954c44990ca2edcd9b4cc12c20321dd20d27cce6/kymatio/scattering1d/filter_bank.py#L382 
https://github.com/kymatio/kymatio/blob/954c44990ca2edcd9b4cc12c20321dd20d27cce6/kymatio/scattering1d/filter_bank.py#L395 

#983 proposes to solve this by changing the `mean` to a `sum` to preserve l1 norm. we get a nice response at `S1` for a dirac:

image

image 

The difference between `n1` and `n1_max` is not huge and the decay is smooth rather than these abrupt and steep jumps between octaves.

lostanlen:
Fixes #984 

With @cyrusvahidi we have spotted an issue in scattering filter factory, while working on JTFS

```python
# The idiom min(previous_J, j, log2_T) implements "j1 < j2"
for level in range(1, min(previous_J, j, log2_T)):
    psi_levels.append(psi_level)
max_j = max(j, max_j)
```

This `.mean(axis=0)` should be a `sum` instead so as to preserve L1 norm across resolution levels.
The same problem arises in the construction of phi_levels:

```python
# Same with low-pass filter phi
sigma_low = filterbank_kwargs["sigma0"] / T
phi_levels = [gauss_1d(N, sigma_low)]
for level in range(1, max(previous_J, 1+log2_T)):
    phi_level = phi_levels[0].reshape(2 ** level, -1).mean(axis=0)
    phi_levels.append(phi_level)
phi_f = {'levels': phi_levels, 'xi': 0, 'sigma': sigma_low, 'j': log2_T,
    'N': N}
```

This is not so much of a big deal for `Scattering1D` (it causes discrepancies of the order of `2**k1` at the first order, and `2**(k1+k2)` at the second order) but it becomes a huge problem for JTFS because it introduces discontinuities at octave boundaries (when those octave boundaries correspond to changes in resolution level).

@cyrusvahidi has made several examples with Diracs and exponential chirps, demonstrating the issue.

This PR provide a backwards compatible fix to the issue. It introduces a temporary attribute `self._reduction`, which defaults to `np.mean` for the time being. We are setting it to `np.sum` for JTFS. Here is the new code

```python
# The idiom min(previous_J, j, log2_T) implements "j1 < j2"
for level in range(1, min(previous_J, j, log2_T)):
    psi_level_reshaped = psi_levels[0].reshape(2**level, -1)
    psi_level = _reduction(psi_level_reshaped, axis=0)
    psi_levels.append(psi_level)
max_j = max(j, max_j)

```

Long term i think that all reductions should be `np.sum` and that we should remove the `_reduction` argument. But i'm introducing it here so that all regression tests still pass and we can merge this quickly (it's really urgent given an impending JTFS paper deadline) without rushing into big API decisions.

### DIFF
--- a/kymatio/scattering1d/filter_bank.py
+++ b/kymatio/scattering1d/filter_bank.py
@@ -319,7 +319,7 @@ def spin(filterbank_fn, filterbank_kwargs):
     return spinned_fn, filterbank_kwargs
 
 
-def scattering_filter_factory(N, J, Q, T, filterbank):
+def scattering_filter_factory(N, J, Q, T, filterbank, _reduction=np.mean):
     """
     Builds in Fourier the Morlet filters used for the scattering transform.
 
@@ -351,6 +351,8 @@ def scattering_filter_factory(N, J, Q, T, filterbank):
         **filterbank_kwargs as optional keyword arguments.
         Corresponds to the self.filterbank property of the scattering object.
         As of v0.3, only anden_generator is supported as filterbank_fn.
+    _reduction : callable
+        either np.mean (default) or np.sum.
 
     Returns
     -------
@@ -379,7 +381,8 @@ def scattering_filter_factory(N, J, Q, T, filterbank):
             # Resample to smaller resolutions if necessary (beyond 1st layer)
             # The idiom min(previous_J, j, 1+log2_T) implements "j1 < j2"
             for level in range(1, min(previous_J, j, 1+log2_T)):
-                psi_level = psi_levels[0].reshape(2 ** level, -1).mean(axis=0)
+                psi_level_reshaped = psi_levels[0].reshape(2**level, -1)
+                psi_level = _reduction(psi_level_reshaped, axis=0)
                 psi_levels.append(psi_level)
             psi_f.append({'levels': psi_levels, 'xi': xi, 'sigma': sigma, 'j': j})
             max_j = max(j, max_j)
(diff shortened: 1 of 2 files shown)


cyrusvahidi (approved):
Looks good to me. Backwards compatibility for `Scattering1D`. New kwargs introduced silently. Let's go.

PR #987, opened 2023-01-10
FIX `get_max_dyadic_subsampling` for negative `xi` in JTFS frequency filterbank construction

### LINKED ISSUE #986: BUG JTFS filterbank_fr construction produces incorrect `j` for spinned wavelets
When constructing a frequential filterbank with `scattering_filter_factory` we pass `(filterbank_fn, filterbank_kwargs)`. This `filterbank_fn` is spinned with `filter_bank.spin`

However, `get_max_dyadic_subsampling` is not aware of the sign of `xi`, so treats negative `xi` as further dyadic subsamplings, resulting in `j` beyond `J_fr`. This has knock on effects for how many `phi_fr` resolutions we create, since this is a function of `previous_J`. For a spinned filterbank, we reach `j = 11` although `J_fr = 6`. 

This was only caught when experimenting with `Q_fr > 1`

```
from kymatio.torch import TimeFrequencyScattering

jtfs = TimeFrequencyScattering(
    shape=(2**13,), Q=(8, 1), J=13, J_fr=6, F=0, Q_fr=2, format="time"
)
jtfs(torch.randn(2**13))
```
```
  File "/homes/cv300/Documents/kymatio/kymatio/scattering1d/filter_bank.py", line 401, in scattering_filter_factory
    phi_level_reshaped = phi_levels[0].reshape(2**level, -1)
ValueError: cannot reshape array of size 576 into shape (128,newaxis)
```

Clearly in the above case, the value `level` has gone beyond `J_fr`.

What results is incorrect values for `j` and therefore incorrect subsampling:

```
xi: -0.003866990209613928, j: 6, 
xi: -0.35, j: 5
```

The largest and smallest spinned `xi` differ by just one factor of two in subsampling.

This can be fixed in the call to `get_max_dyadic_subsampling` in `scattering_filter_factory` by ensuring we pass the absolute value of `xi`:

`j = get_max_dyadic_subsampling(abs(xi), sigma, **filterbank_kwargs)`

Now, `j` is in line with what we expect:

```
xi: -0.003866990209613928, j: 5, 
xi: -0.35, j: 0
```

I propose that we make a PR for this fix with tests. Do we implement this in `scattering_filter_factory` or `get_max_dyadic_subsampling`? Perhaps it makes sense that `get_max_dyadic_subsampling` always behaves as expected, regardless of the sign of `xi`.

@lostanlen

cyrusvahidi:
fixes #986 

The new unit test failed without the introduction of `abs(xi)` in `get_max_dyadic_subsampling`. I test that `j` is mirrored for positive and negative spinned filters. This ensures that we subsample correctly for negative spin filters.

### DIFF
--- a/kymatio/scattering1d/filter_bank.py
+++ b/kymatio/scattering1d/filter_bank.py
@@ -214,7 +214,7 @@ def get_max_dyadic_subsampling(xi, sigma, alpha, **unused_kwargs):
     Parameters
     ----------
     xi : float
-        frequency of the filter in [-1, 1]
+        frequency of the filter in [-0.5, 0.5]
     sigma : float
         frequential width of the filter
     alpha : float, optional
--- a/kymatio/scattering1d/filter_bank.py
+++ b/kymatio/scattering1d/filter_bank.py
@@ -204,7 +204,7 @@ def get_max_dyadic_subsampling(xi, sigma, alpha, **unused_kwargs):
     Finds the maximal integer j such that:
     omega_0 < 2^{-(j + 1)}
     where omega_0 is the boundary of the filter, defined as
-    omega_0 = xi + alpha * sigma
+    omega_0 = abs(xi) + alpha * sigma
 
     This ensures that the filter can be subsampled by a factor 2^j without
     aliasing.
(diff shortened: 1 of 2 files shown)


lostanlen (approved):
Nice PR. Many thanks

PR #988, opened 2023-01-12
MAINT `oversampling` --> `stride`, immutable `Q` and `T`

### LINKED ISSUE #937: Should we keep `oversampling`?

lostanlen:
Fixes #937 

I have written the smallest PR possible
In a separate PR i will rewrite core 1D in terms of log2_stride instead of oversampling. But that's not necessary right now.

~~Note that stride is passed at runtime, not in the constructor. This allows to change stride between inputs without recomputing the filters.~~  actually no because this causes problem in autodiff

Should be reviewed and merged before v0.4

I will also make a PR about `stride_fr` later

### DIFF
--- a/kymatio/scattering1d/core/scattering1d.py
+++ b/kymatio/scattering1d/core/scattering1d.py
@@ -1,5 +1,5 @@
 
-def scattering1d(U_0, backend, filters, oversampling, average_local):
+def scattering1d(U_0, backend, filters, log2_stride, average_local):
     """
     Main function implementing the 1-D scattering transform.
 
@@ -30,10 +30,8 @@ def scattering1d(U_0, backend, filters, oversampling, average_local):
         * `sigma`: float, bandwidth
         * 'levels': list, values taken by the lowpass in the Fourier domain
                     at different levels of detail.
-    oversampling : int >=0
-        Yields scattering coefficients with a temporal stride equal to
-        max(1, 2**(log2_T-oversampling)). Hence, raising oversampling by
-        one unit halves the stride, until reaching a stride of 1.
+    log2_stride : int >=0
+        Yields coefficients with a temporal stride equal to 2**log2_stride.
     average_local : boolean
         whether to locally average the result by means of a low-pass filter phi.
     """
@@ -44,11 +42,11 @@ def scattering1d(U_0, backend, filters, oversampling, average_local):
     # Get S0
     phi = filters[0]
     log2_T = phi['j']
-    k0 = max(log2_T - oversampling, 0)
+    stride = 2**log2_stride
 
     if average_local:
         S_0_c = backend.cdgmm(U_0_hat, phi['levels'][0])
-        S_0_hat = backend.subsample_fourier(S_0_c, 2**k0)
+        S_0_hat = backend.subsample_fourier(S_0_c, stride)
         S_0_r = backend.irfft(S_0_hat)
         yield {'coef': S_0_r, 'j': (), 'n': ()}
     else:
@@ -59,10 +57,7 @@ def scattering1d(U_0, backend, filters, oversampling, average_local):
     for n1 in range(len(psi1)):
         # Convolution + downsampling
         j1 = psi1[n1]['j']
-
-        sub1_adj = min(j1, log2_T) if average_local else j1
-        k1 = max(sub1_adj - oversampling, 0)
-
+        k1 = min(j1, log2_stride) if average_local else j1
         U_1_c = backend.cdgmm(U_0_hat, psi1[n1]['levels'][0])
         U_1_hat = backend.subsample_fourier(U_1_c, 2**k1)
         U_1_c = backend.ifft(U_1_hat)
@@ -75,7 +70,7 @@ def scattering1d(U_0, backend, filters, oversampling, average_local):
 
         if average_local:
             # Convolve with phi_J
-            k1_J = max(log2_T - k1 - oversampling, 0)
+            k1_J = max(log2_stride - k1, 0)
             S_1_c = backend.cdgmm(U_1_hat, phi['levels'][k1])
             S_1_hat = backend.subsample_fourier(S_1_c, 2**k1_J)
             S_1_r = backend.irfft(S_1_hat)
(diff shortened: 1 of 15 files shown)


lostanlen:
will work on improving coverage + making `S.oversampling` a property with `DeprecationWarning`

lostanlen:
Still working on it. It's not so trivial to interface T, stride, and oversampling in the way we wish.

lostanlen:
OK, this time the algorithm fully depends on `log2_stride`.

I have preserved backwards compatibility (approximately in terms of result and exactly in terms of output size) via this line

```python
        if self._oversampling > 0:
            warn("oversampling is deprecated in and will be removed in v0.5."
                "Pass stride = 2**(J-oversampling) or "
                "stride = 2**(log2(T)-oversampling) "
                "to retain current behavior.", DeprecationWarning)
            if not self._stride in [None, 2**(self.log2_T-self._oversampling-1)]:
                raise ValueError("oversampling={} and stride={}"
                    "are incompatible".format(self._oversampling, self._stride))
            self._stride = 2 ** (self.log2_T-self._oversampling)
```

i have updated all frontends, both for Sc1D and JTFS

lostanlen:
i am missing a few unit tests, hence the-0.21% reduction. on it

lostanlen:
now the project coverage increases but the patch coverage is slightly below target. currently working on it

kymatio/scattering1d/core/scattering1d.py
```diff
@@ -91,26 +86,22 @@ def scattering1d(U_0, backend, filters, oversampling, average_local):
...
 
                     if average_local:
-                        U_2_hat = backend.rfft(U_2_m)
-
                         # Convolve with phi_J
-                        k2_log2_T = max(log2_T - k2 - k1 - oversampling, 0)
-
-                        S_2_c = backend.cdgmm(U_2_hat, phi['levels'][k1 + k2])
+                        U_2_hat = backend.rfft(U_2_m)
+                        k2_log2_T = max(log2_stride - sub2_adj, 0)
```
cyrusvahidi, line 101: Why no `-k1`?

lostanlen: 2**sub2_adj is the stride of U2 (granted, we could have a better name ...)
2**log2_stride is the stride of S2

going from U2 to S2 requires a subsampling of 2**(log2_stride - sub2_adj)

kymatio/scattering1d/frontend/base_frontend.py
```diff
@@ -237,25 +230,68 @@ def _check_input(self, x):
...
+        if np.any(np.array(self._Q) < 1):
+            raise ValueError('Q must always be >= 1, got {}'.format(self._Q))
+        if isinstance(self._Q, int):
+            return (self._Q, 1)
+        elif isinstance(self._Q, tuple):
+            if len(self._Q) == 1:
+                return (self._Q + (1,))
+            elif len(self._Q) == 2:
+                return self._Q
+            raise NotImplementedError("Q must be an integer, 1-tuple or "
```
cyrusvahidi, line 280: can you add a test that raises this error?

lostanlen: on it

lostanlen: 90cbb6b

lostanlen: 90cbb6b

lostanlen:
Failing on Jenkins!

```python

Failed to build tensorflow-gpu

Installing collected packages: python_version, fastrlock, typing-extensions, threadpoolctl, tensorflow-gpu, pycuda, pillow, numpy, joblib, configparser, torch, scipy, scikit-cuda, opt_einsum, cupy, torchvision, scikit-learn, jaxlib, jax

  Running setup.py install for tensorflow-gpu: started

  Running setup.py install for tensorflow-gpu: finished with status 'error'

  error: subprocess-exited-with-error

  

  × Running setup.py install for tensorflow-gpu did not run successfully.

  │ exit code: 1

  ╰─> [18 lines of output]

        File "<string>", line 2, in <module>

        File "<pip-setuptools-caller>", line 34, in <module>

        File "/tmp/pip-install-nch9ufm9/tensorflow-gpu_092a4daa64ee45258b249636ae5cb519/setup.py", line 37, in <module>

          raise Exception(TF_REMOVAL_WARNING)

      Exception:

      

      =========================================================

      The "tensorflow-gpu" package has been removed!

      

      Please install "tensorflow" instead.

      

      Other than the name, the two packages have been identical

      since TensorFlow 2.1, or roughly since Sep 2019. For more

      information, see: pypi.org/project/tensorflow-gpu

      =========================================================
```

cyrusvahidi (approved):

Issue #857, opened 2022-06-08
redefine 1D `self.N`, deprecate`self.J_pad`?


lostanlen:
in 1D, `self.N` is the unpadded length of the input signal. It is defined in `ScatteringBase1D` as some derivative of `self.shape` and isn't being used anywhere else. 

Same for 2D where `self.shape` unpacks as `self.M`, `self.N`. 

I see no reason to keep these values as attributes in the object when:
* we have `self.shape` for that
* these values are only being used transiently, in `self.build()`
* our filterbank construction relies on `N_padded`, not `N`
* these values are trivially available in the forward via `self.N = len(x)`

In the spirit of preparing for scattering on variable-length inputs (`shape=None`, a long-requested feature !), i ask the permission to remove these attributes from `self`.

I am open to making a deprecation cycle where we convert these attributes to read-only properties and print a `DeprecationWarning` in v0.3 if anyone tries to access these numbers (NB: none of our examples access `self.N` and friends)

@janden @eickenberg

lostanlen:
or else we decide that `self.N` now stands for `N_padded`, i.e., the length of the signal after padding = the length of the filters at the full resolution. This would actually be most consistent with the 3D module. But that implies to rename fields in the 2D module: `N_padded` becomes `N`, `M_padded` becomes `M`.

janden:
I'd be in favor of removing. We could argue that these were intended to be private attributes and users shouldn't rely on them. Safe thing would be to issue a deprecation warning, however.

lostanlen:
> We could argue that these were intended to be private attributes

@janden i came to the same conclusion. 


>  Safe thing would be to issue a deprecation warning, however.

i can do that !

janden:
@eickenberg Your call on this API-breaking change.

lostanlen:
this PR is related to JTFS. Here's my rationale:
* i'd like to avoid having to write a `jtfs_filter_factory`
* rather, i'd like to call `scattering_filter_factory` twice: once for the temporal filters and again (with different arguments) for the frequential filters
* but i don't want to have to assume that the input  of the frequential scattering has a power-of-two support. That would require a lot of padding (up to ~JQ zeros, i.e. >100) and thus woul hurt speed and memory usage. Rather i'd like to pad to a multiple of log2_F (aka the frequential "stride" )
[...]

PR #989, opened 2023-01-13
remove J_pad and N from ScatteringBase1D

lostanlen:
after v0.3 deprecation cycle
(segue of #857)

for v0.4


lostanlen:
merging (cf #857 for janden's approval)

PR #992, opened 2023-01-31
TST Fix Tensorflow pip package in Jenkins

janden:
Replace `tensorflow-gpu` package with `tensorflow` as the former has been deprecated.


lostanlen (approved):
Great. Many thanks!

PR #998, opened 2023-02-16
TST `generate_test_data`

### LINKED ISSUE #966: frontends should expose TimeFrequencyScattering
Currently the frontends doesn't include TimeFrequencyScattering, one needs to do "from kymatio.scattering1d.frontend.xxx import TimeFrequencyScatteringxxx. "

lostanlen:
Fixes #966 

i had to reduce keras atol down to 1e-6 (it was 1e-7 before). not sure why

1e-7 is below float32 precision so i'm not sure how that was possible


cyrusvahidi (approved):
bon, from what I remember this is to fix the breaking change introduced by the bug fix in filter subsampling.

PR #1003, opened 2023-03-03
FIX export JTFS from jax + sklearn module

cyrusvahidi:
Following #1001 I have exported JTFS from the JAX frontend such that one can import as: `from kymatio.jax import TimeFrequencyScattering` 

I noticed that this was also missing in `sklearn`.


lostanlen:
thanks. i believe you need to rebase? the first two commits are merged now

PR #1006, opened 2023-03-07
MAINT remove `_reduction` from JTFS self

lostanlen:
segue from #1005 
in the case of JTFS, there is no backwards compatibility to be had, so we might as well remove the `_reduction` attribute from the JTFS object and pass `_reduction=np.mean` to filterbank constructors.


cyrusvahidi:
Why not `_reduction=np.sum`?

cyrusvahidi (approved):

PR #1008, opened 2023-03-13
DOC Clean up 1D filter plotting example

janden:
Some minor cleanup of this example. In particular, the plots now appear correctly in the gallery (issue was ~~the `plt.figure` call~~ a naming conflict with the 2D version of the script).

However, some issues remain that I need help with:

- [x] The whole setup with `anden_generator` (perhaps a better name is needed here) should be explained.
- [x] We should avoid passing these magical numbers (`alpha`, `r_psi`, and `sigma0`) here. These are the default values, so why should the user have to specify them here?
- [x] The frequency response looks a bit off?
image
Same for the time domain.

Perhaps @lostanlen or @cyrusvahidi could help out?


janden:
Setting `sigma0 = 0.1` seems to have fixed the figure:

image

I'd argue that this is an even stronger argument for having these parameters be the default.

janden:
@lostanlen @cyrusvahidi Would still like your input here.

lostanlen:
> I'd argue that this is an even stronger argument for having these parameters be the default.

Yes, `sigma0=0.1` is the default value of `anden_generator`. The previous value `sigma0=5` was clearly way too high. Thanks for catching this

janden:
Ok. Going to go ahead and implement those defaults. Will also rename the generator to `scatnet_generator` unless there are any strong objections.

examples/1d/plot_filters_1d.py
```diff
@@ -15,7 +15,7 @@
 # First, we import the `scattering_filter_factory` method, which we will use
 # to generate the filters.
 
-from kymatio.scattering1d.filter_bank import anden_generator, scattering_filter_factory
```
MuawizChaudhary, line 18: We wanted to move away from the scatnet name and all "scat" related naming, so I am not a fan of this commit.

MuawizChaudhary (changes requested):
Not a fan of this specific commit. is it possible to rename to something other than `scatnet`?

MuawizChaudhary (approved):
While I don't agree with the scat re-naming, I don't think that should be a blocker to release. 

Ideally it'd be nice to come up with a renaming before 0.4 release.

examples/1d/plot_filters_1d.py
```diff
@@ -83,14 +86,13 @@
...
 # ------------------------------------------
+#
 # We are now ready to plot the frequency response of the filters.
 # We first display the lowpass filter (at full resolution) in red.
 # We then plot each of the bandpass filters in blue. Since we don't care about
 # the negative frequencies, we limit the plot to the frequency interval
 # $[0, 0.5]$.
 
-plt.figure()
-plt.rcParams.update({"text.usetex": True})
```
MuawizChaudhary, line 93: Why this removal?

MuawizChaudhary (approved):

Issue #1021, opened 2023-09-08
Why is optimizer momentum reset every 20 epochs on CIFAR10 2D examples?


vivianwhite:
Hello! I have been adapting the Kymatio 2D classification examples for a research project about randomized wavelet scattering. I noticed that in the CIFAR10 examples, during the training loop, the optimizer is being reset every 20 epochs. Links to code [here](https://github.com/kymatio/kymatio/blob/main/examples/2d/cifar_torch.py#L169), [here](https://github.com/kymatio/kymatio/blob/main/examples/2d/cifar_small_sample.py#L232), and [here](https://github.com/kymatio/kymatio/blob/main/examples/2d/cifar_resnet_torch.py#L198).
```
if epoch%20==0:
    optimizer = torch.optim.SGD(model.parameters(), lr=lr, momentum=0.9,
                                         weight_decay=0.0005)
```
The intention of this appears to be to update the learning rate, but it is also resetting the momentum buffer, which keeps a running history of the gradients. Is this intended behavior?

I looked into Scattering Networks for Hybrid Representation Learning by Edouard Oyallon (@edouardoyallon) and Eugene Belilovsky (@eugenium) and there seemed to be no explicit mention of resetting momentum buffers, though there was mention of a learning rate scheduler (paper linked [here](https://arxiv.org/pdf/1809.06367.pdf)). It appears that the paper was using a similar training process/hyperparameters to the Wide Resnet paper (linked [here](https://arxiv.org/abs/1605.07146)). In the Wide Resnet codebase, the optimizer momentum buffer also appears to be reset every time the learning rate scheduler updates (link to code [here](https://github.com/szagoruyko/wide-residual-networks/blob/master/pytorch/main.py#L164)). However, there is no explicit mention of restarting the momentum buffers in the paper.

MuawizChaudhary:
I talked to @edouardoyallon and he indicated that the proper behavior should be to initialize the optimizer outside of the loop, and then decay learning rate every 20 epochs . 

You can implement this in one of two ways; 

1. you access the optimiser `param_group` then access and change the learning rate within the `param_group` or 
2. you define a [learning rate schedule](https://pytorch.org/docs/stable/generated/torch.optim.lr_scheduler.StepLR.html#torch.optim.lr_scheduler.StepLR) that updates every 20 epochs up to 80.

I'll leave it to you and/or others to choose which implementation you'll do.

MuawizChaudhary:
[...]

Issue #858, opened 2022-06-09
Examples are not in CI


janden:
Following offline discussion with @lostanlen. Since examples are not tested regularly, they *will* break (see #817, #801, #775) without us realizing it. We should aim to have the examples run at somewhat regularly so we can stay ahead of these things. This will probably include trimming down the examples to a smaller and easier maintainable set.

MuawizChaudhary:
#817  is not really an instance of the examples failing. The first example It's rather an example of the documentation not being created on a machine with GPU. The scattering disk example has looked like that since it was integrated into the library. 

#801 is not an appropriate instance of an example. I'm 90% sure the guy was using a branch of kymatio created by OLGD. We never implemented a agnostic backend. 

#775 is a dire issue that needs fixing; the example is not broken, our software is. 

I do agree that there could be some pruning. There are [three](https://www.kymat.io/gallery_2d/cifar_torch.html#sphx-glr-gallery-2d-cifar-torch-py) different [examples](https://www.kymat.io/gallery_2d/cifar_resnet_torch.html#sphx-glr-gallery-2d-cifar-resnet-torch-py) all on [cifar](https://www.kymat.io/gallery_2d/cifar_small_sample.html#sphx-glr-gallery-2d-cifar-small-sample-py). 

On the other hand, I don't think we should be too aggressive with the pruning. Our examples is a point of entry for people to join on in the development of Kymatio. On the other hand, I think we should have more varied, original, examples.

janden:
Well at least #775 would have been caught if the example was under CI. Same for #829. All I'm saying is that if we have examples on our website, we should know that they run reliably. We should also be testing the TF/Keras frontends as part of our regular test suite, I agree.

MuawizChaudhary:
I don't know how easy it would be to test the graphical examples, but code that replicates experiments done in papers such as [`cifar_small_sample.py`](https://github.com/kymatio/kymatio/blob/main/examples/2d/cifar_small_sample.py), or  [`cifar_resnet_torch.py`](https://github.com/kymatio/kymatio/blob/main/examples/2d/cifar_resnet_torch.py) or [`cifar_torch.py`](https://github.com/kymatio/kymatio/blob/main/examples/2d/cifar_torch.py) make alot of sense to me to test. I have created a separate thread for discussion. We should probably use some sort of categorization for the examples (i.e. graphical output, experimental validation) in order to determine which examples should be tested. 

[...]

PR #1022, opened 2023-09-11
MAINT use lr_scheduler instead of resetting optimizer

vivianwhite:
Instead of redefining the optimizer every 20 epochs during training to decrease the learning rate, I initialized it before the training loop starts along with a StepLR scheduler that will decrease the learning rate by a factor of 0.2 every 20 epochs. 

I changed this for one of the three CIFAR10 2d example scripts, and I will change the other two if everyone agrees.

This resolves issue #1021

### DIFF
--- a/examples/2d/cifar_small_sample.py
+++ b/examples/2d/cifar_small_sample.py
@@ -229,7 +229,7 @@ def main():
                                         weight_decay=0.0005)
     M = args.learning_schedule_multi
     drops = [60*M,120*M,160*M]
-    scheduler = torch.optim.MultiStepLR(optimizer, drops, gamma=0.2)
+    scheduler = torch.optim.lr_scheduler.MultiStepLR(optimizer, drops, gamma=0.2)
     for epoch in range(0, 200*M):
         train(model, device, train_loader, optimizer, epoch+1, scattering)
         scheduler.step()
--- a/examples/2d/cifar_small_sample.py
+++ b/examples/2d/cifar_small_sample.py
@@ -228,7 +228,7 @@ def main():
     optimizer = torch.optim.SGD(model.parameters(), lr=lr, momentum=0.9,
                                         weight_decay=0.0005)
     M = args.learning_schedule_multi
-    drops = [0, 60*M,120*M,160*M]
+    drops = [60*M,120*M,160*M]
     scheduler = torch.optim.MultiStepLR(optimizer, drops, gamma=0.2)
     for epoch in range(0, 200*M):
         train(model, device, train_loader, optimizer, epoch+1, scattering)
--- a/examples/2d/cifar_small_sample.py
+++ b/examples/2d/cifar_small_sample.py
@@ -228,13 +228,11 @@ def main():
     optimizer = torch.optim.SGD(model.parameters(), lr=lr, momentum=0.9,
                                         weight_decay=0.0005)
     M = args.learning_schedule_multi
-    drops = [60*M,120*M,160*M]
+    drops = [0, 60*M,120*M,160*M]
+    scheduler = torch.optim.MultiStepLR(optimizer, drops, gamma=0.2)
     for epoch in range(0, 200*M):
-        if epoch in drops or epoch==0:
-            for param_group in optimizer.param_groups:
-                param_group['lr'] *= 0.2
-
         train(model, device, train_loader, optimizer, epoch+1, scattering)
+        scheduler.step()
         if epoch%10==0:
             test(model, device, test_loader, scattering)
 
--- a/examples/2d/cifar_small_sample.py
+++ b/examples/2d/cifar_small_sample.py
@@ -225,13 +225,14 @@ def main():
 
     # Optimizer
     lr = 0.1
+    optimizer = torch.optim.SGD(model.parameters(), lr=lr, momentum=0.9,
+                                        weight_decay=0.0005)
     M = args.learning_schedule_multi
     drops = [60*M,120*M,160*M]
     for epoch in range(0, 200*M):
         if epoch in drops or epoch==0:
-            optimizer = torch.optim.SGD(model.parameters(), lr=lr, momentum=0.9,
-                                        weight_decay=0.0005)
-            lr*=0.2
+            for param_group in optimizer.param_groups:
+                param_group['lr'] *= 0.2
 
         train(model, device, train_loader, optimizer, epoch+1, scattering)
         if epoch%10==0:
--- a/examples/2d/cifar_torch.py
+++ b/examples/2d/cifar_torch.py
@@ -165,11 +165,12 @@ if __name__ == '__main__':
 
     # Optimizer
     lr = 0.1
+    optimizer = torch.optim.SGD(model.parameters(), lr=lr, momentum=0.9,
+                                        weight_decay=0.0005)
+    scheduler = torch.optim.lr_scheduler.StepLR(optimizer, step_size=20, gamma=0.2)
     for epoch in range(0, 90):
         if epoch%20==0:
-            optimizer = torch.optim.SGD(model.parameters(), lr=lr, momentum=0.9,
-                                        weight_decay=0.0005)
-            lr*=0.2
+            scheduler.step()
 
         train(model, device, train_loader, optimizer, epoch+1, scattering)
         test(model, device, test_loader, scattering)
--- a/examples/2d/cifar_torch.py
+++ b/examples/2d/cifar_torch.py
@@ -168,9 +168,8 @@ if __name__ == '__main__':
     optimizer = torch.optim.SGD(model.parameters(), lr=lr, momentum=0.9,
                                         weight_decay=0.0005)
     scheduler = torch.optim.lr_scheduler.StepLR(optimizer, step_size=20, gamma=0.2)
-    for epoch in range(0, 90):
-        if epoch%20==0:
-            scheduler.step()
 
+    for epoch in range(0, 90):
         train(model, device, train_loader, optimizer, epoch+1, scattering)
+        scheduler.step()
         test(model, device, test_loader, scattering)
--- a/examples/2d/cifar_resnet_torch.py
+++ b/examples/2d/cifar_resnet_torch.py
@@ -193,11 +193,12 @@ if __name__ == '__main__':
 
     # Optimizer
     lr = 0.1
-    for epoch in range(0, 90):
-        if epoch%20==0:
-            optimizer = torch.optim.SGD(model.parameters(), lr=lr, momentum=0.9,
+    optimizer = torch.optim.SGD(model.parameters(), lr=lr, momentum=0.9,
                                         weight_decay=0.0005)
-            lr*=0.2
+    scheduler = torch.optim.lr_scheduler.StepLR(optimizer, step_size=20,
+            gamma=0.2)
 
+    for epoch in range(0, 90):
         train(model, device, train_loader, optimizer, epoch+1, scattering)
+        scheduler.step()
         test(model, device, test_loader, scattering)


examples/2d/cifar_torch.py
```diff
@@ -165,11 +165,12 @@ def test(model, device, test_loader, scattering):
...
     lr = 0.1
+    optimizer = torch.optim.SGD(model.parameters(), lr=lr, momentum=0.9,
+                                        weight_decay=0.0005)
+    scheduler = torch.optim.lr_scheduler.StepLR(optimizer, step_size=20, gamma=0.2)
     for epoch in range(0, 90):
         if epoch%20==0:
-            optimizer = torch.optim.SGD(model.parameters(), lr=lr, momentum=0.9,
-                                        weight_decay=0.0005)
-            lr*=0.2
+            scheduler.step()
```
MuawizChaudhary, line 173: Since you are using the `StepLR` learning rate scheduler, this step should be done at the end of every update.

examples/2d/cifar_torch.py
```diff
@@ -165,11 +165,12 @@ def test(model, device, test_loader, scattering):
 
     # Optimizer
     lr = 0.1
+    optimizer = torch.optim.SGD(model.parameters(), lr=lr, momentum=0.9,
+                                        weight_decay=0.0005)
+    scheduler = torch.optim.lr_scheduler.StepLR(optimizer, step_size=20, gamma=0.2)
```
MuawizChaudhary, line 170: good work

MuawizChaudhary (changes requested):
please remove the check and make the step update happen at the end of every epoch.

MuawizChaudhary (changes requested):
Looks good to me. I think you can update the rest of the scripts with this same set of changes. I'll run tests on this script and make sure that the accuracies are similar. 

I will merge this by Wednesday unless other contributors ask for more time.

examples/2d/cifar_small_sample.py
```diff
@@ -225,13 +225,14 @@ def main():
...
+                                        weight_decay=0.0005)
     M = args.learning_schedule_multi
     drops = [60*M,120*M,160*M]
     for epoch in range(0, 200*M):
         if epoch in drops or epoch==0:
-            optimizer = torch.optim.SGD(model.parameters(), lr=lr, momentum=0.9,
-                                        weight_decay=0.0005)
-            lr*=0.2
+            for param_group in optimizer.param_groups:
+                param_group['lr'] *= 0.2
```
MuawizChaudhary, line 235: I think you should be consistent and use the same strategy you used previously for updating the LR.

MuawizChaudhary (changes requested):

examples/2d/cifar_small_sample.py
vivianwhite: I updated the lr differently this time in order to handle the drops, but I will use a MultiStepLR scheduler instead

examples/2d/cifar_small_sample.py
```diff
@@ -225,15 +225,14 @@ def main():
 
     # Optimizer
     lr = 0.1
+    optimizer = torch.optim.SGD(model.parameters(), lr=lr, momentum=0.9,
+                                        weight_decay=0.0005)
     M = args.learning_schedule_multi
-    drops = [60*M,120*M,160*M]
+    drops = [0, 60*M,120*M,160*M]
```
MuawizChaudhary, line 231: does the LR change on the 0th epoch?

MuawizChaudhary (changes requested):

examples/2d/cifar_small_sample.py
vivianwhite: hmm nope, i will change that

examples/2d/cifar_small_sample.py
```diff
@@ -225,15 +225,14 @@ def main():
 
     # Optimizer
     lr = 0.1
+    optimizer = torch.optim.SGD(model.parameters(), lr=lr, momentum=0.9,
+                                        weight_decay=0.0005)
     M = args.learning_schedule_multi
     drops = [60*M,120*M,160*M]
+    scheduler = torch.optim.MultiStepLR(optimizer, drops, gamma=0.2)
     for epoch in range(0, 200*M):
```
MuawizChaudhary, line 233: Hey @vivianwhite can you investigate the original paper and see why @edouardoyallon uses this scaling factor M? (Alternatively Edouard, if you can drop in with whats the intention of this M that would be helpful)

I am confused by the description in the script ([here](https://github.com/kymatio/kymatio/blob/7e4099539c5c59eea8a76edcd5c940489e8d6cee/examples/2d/cifar_small_sample.py#L169)) which states "samples per class". I am under the impression this was mixed up with the previous argument `num_samples`, which also states in it's help text `samples per class`.

MuawizChaudhary (changes requested):

examples/2d/cifar_small_sample.py
vivianwhite: M is a learning rate multiplier. From Deep Hybrid Networks paper ([here](https://openaccess.thecvf.com/content_ICCV_2017/papers/Oyallon_Scaling_the_Scattering_ICCV_2017_paper.pdf)): 

> We use the same training settings for our baseline, WRN
> 16-8, and our hybrid scattering and WRN-12. The settings
> are the same as those described for CIFAR-10 in the previous 
> section with the only difference being that we apply a
> multiplier to the learning rate schedule and to the maximum
> number of epochs. The multiplier is set to 10,20,100 for the
> 1000,500, and 100 sample case respectively. For example
> the default schedule of 60,120,160 becomes 600,1200,1600
> for the case of 1000 samples and a multiplier of 10. Finally
> in the case of 100 samples we use a batch size of 32 in lieu
> of 128.

MuawizChaudhary: Good work!

MuawizChaudhary (commented):
I am satisfied with the work done here. I am currently running experiments to make sure that the results line up, and then I will approve for merge.

MuawizChaudhary:
I have an initial set of results on the `main` branch of kymatio. 

> cifar_torch.py
> mode 1, linear = 63.99
> mode 1, mlp = 80.50
> mode 1, cnn = 87.61
> mode 2 , linear = nan
> mode 2, mlp = 81.87
> mode 2, cnn = 88.48
> 
> cifar_small_sample.py
> standard  = 53.91
> scattering = 53.42
> 
> cifar_resnet_torch.py
> mode 1 = 85.84
> mode 2 = 86.94

I'll run 3 seeds, and then run this same test on Vivian's branch. I'll throw the scripts I used into a throw-away branch, just in case someone wants to critique my methodology or we wanna re-use code if we decided to automate this sort of testing. see #1024 and #858

examples/2d/cifar_small_sample.py
```diff
@@ -225,15 +225,14 @@ def main():
 
     # Optimizer
     lr = 0.1
+    optimizer = torch.optim.SGD(model.parameters(), lr=lr, momentum=0.9,
```
MuawizChaudhary, line 228: you forgot a `lr_scheduler.`

MuawizChaudhary (changes requested):
Currently running branch `fix_reset_momentum` and collating results for branch `main`.

I'll post results as I get them.

examples/2d/cifar_small_sample.py
```diff
@@ -225,15 +225,14 @@ def main():
...
+                                        weight_decay=0.0005)
     M = args.learning_schedule_multi
     drops = [60*M,120*M,160*M]
+    scheduler = torch.optim.MultiStepLR(optimizer, drops, gamma=0.2)
```
MuawizChaudhary, line 232: you forgot a `lr_scheduler.`

MuawizChaudhary:
Results for main branch, all of them calculated with 3 runs except when otherwise specified. I collated the results in this google colab [here](https://colab.research.google.com/drive/1TnUNsWpeQTRIDXy38-u5GxnReL0JP1wn?usp=sharing). Scripts used to run these experiments and text files containing the results are located on branch [`example_2d_torch_seed_test`](https://github.com/MuawizChaudhary/kymatio/tree/example_2d_torch_seed_test/examples/2d) on my fork. 

`cifar_torch.py` on `main` branch
> Linear Classifier Mode 1: Mean: 64.01 Standard Error: 0.09
> MLP Classifier Mode 1: Mean: 80.19 Standard Error: 0.07
> CNN Classifier Mode 1: Mean: 87.41 Standard Error: 0.18
> Linear Classifier Mode 2 (2 SEEDS ONLY): Mean: 70.36 Standard Error: 0.06
> MLP Classifier Mode 2: Mean: 82.38 Standard Error: 0.24
> CNN Classifier Mode 2: Mean: 88.38 Standard Error: 0.22

I didn't get a NAN result with the linear classifier with max order = 2 on scattering when i tried other seeds. 

`cifar_resnet_torch.py` on `main` branch

> Mode 1: Mean: 86.26 Standard Error: 0.37
> Mode 2: Mean: 86.21 Standard Error: 0.44

`cifar_small_sample.py` on `main` branch.
 Note, results were computed with 3 varying dataset seeds. The random number generator for the model and training was held constant due to a lack of time.
 
> Scattering: Mean: 53.01 Standard Error: 0.21
> Standard: Mean: 50.38 Standard Error: 1.74

I am waiting on the results for `fix_reset_momentum` branch.

eickenberg (commented):
Looks fine to me. Does it run? Does it attain the required accuracy?
Using lr scheduler objects is fine I guess, although I can't remember for which pytorch version they were introduced, and which pytorch version we currently support.

MuawizChaudhary:
> Looks fine to me. Does it run? Does it attain the required accuracy? Using lr scheduler objects is fine I guess, although I can't remember for which pytorch version they were introduced, and which pytorch version we currently support.

It runs. According to my tests, its within error bars. I believe the LR schedulers were introduced a while ago, since at least 2020/2021, and we support the pytorch versions that use em. 

I am pro merging. I'll do it tomorrow. 

[Results are collated in this google colab.](https://colab.research.google.com/drive/1TnUNsWpeQTRIDXy38-u5GxnReL0JP1wn?usp=sharing). I've posted results and scripts used to run `fix_reset_momentum` to my [fork's branch of `fix_reset_momentum`](https://github.com/MuawizChaudhary/kymatio/tree/fix_reset_momentum/examples/2d). main branch testing scripts and results [here](https://github.com/MuawizChaudhary/kymatio/tree/example_2d_torch_seed_test/examples/2d). I post results here for completeness. 

`cifar_torch.py` on `fix_reset_momentum` branch

> Linear Classifier Mode 1: Mean: 64.02 Standard Error: 0.08
> MLP Classifier Mode 1: Mean: 79.93 Standard Error: 0.19
> CNN Classifier Mode 1: Mean: 87.35 Standard Error: 0.12
> Linear Classifier Mode 2: Mean: 70.36 Standard Error: 0.08
> MLP Classifier Mode 2: Mean: 82.38 Standard Error: 0.25
> CNN Classifier Mode 2: Mean: 88.25 Standard Error: 0.13


`cifar_resnet_torch.py` on `fix_reset_momentum` branch

> Mode 1: Mean: 85.88 Standard Error: 0.32
> Mode 2: Mean: 85.74 Standard Error: 0.48

`cifar_small_sample.py` on `fix_reset_momentum` branch.

> Scattering: Mean: 53.47 Standard Error: 0.52
> Standard: Mean: 51.81 Standard Error: 1.04

MuawizChaudhary (approved):

eickenberg:
> and we support the pytorch versions that use em.

the question is whether we support pytorch versions that don't. So that would need to be known with certitude.

MuawizChaudhary:
> > and we support the pytorch versions that use em.
> 
> the question is whether we support pytorch versions that don't. So that would need to be known with certitude.

I thought we dropped support for versions of pytorch earlier than 1.13, but the documentation is pretty confusing, see image below. In a[ previous commit](https://github.com/kymatio/kymatio/commit/a0b0c88842784dc327baf5dbc3b0affdd514a509) for the workflows, @janden removed support for earlier versions of pytorch ([more evidence that we just support 1.13 and 2.0](https://github.com/kymatio/kymatio/blob/main/.github/workflows/pip.yml#L14)). It does appear that our code [still checks](https://github.com/kymatio/kymatio/blob/main/kymatio/scattering2d/backend/torch_backend.py#L9) for pytorch version and if its greater than or equal to 1.8. 


image

@janden Are you able to provide an answer to "what pytorch versions does kymatio support?"

But to answer one of your questions: it seems like pytorch support for the lr_schedulers extends to atleast [pytorch 0.4.0](https://pytorch.org/docs/0.4.0/optim.html?highlight=optim#how-to-adjust-learning-rate).

MuawizChaudhary:
Given that this PR is agnostic to the Pytorch version for the use of LR Schedulers, as mentioned in my previous post, I am merging. 

I have raised issue #1025 to discuss specifically what Pytorch version Kymatio supports. 

Congrats @vivianwhite on your first PR!

eickenberg:
the 1.8 check is because complex numbers were introduced then and needed different treatment. I believe the current version supports back to 1.6. If schedulers were present then, we're good. I don't see how this PR is in any way "agnostic" to this information, though.

MuawizChaudhary:
Agonistic as in "assuming the underlying scattering implementation was correct for each pytorch version, this script would work up to at least pytorch 0.4, since lr schedulers were implemented in pytorch back then".  

If kymatio is supposed to be compatible with 1.6, why don't we have any tests for this?

Issue #950, opened 2022-08-25
Images missing from README


janden:
Both on GitHub and PyPI (https://test.pypi.org/project/kymatio/0.3.0a0/). Best solution may be to upload these to `https://kymat.io` and serve them from there.

janden:
Still a problem. Would be good to solve before 0.4 release.

edouardoyallon:
Still a couple of images missing, it would be really nice to fix this because it looks really bad.

MuawizChaudhary:
@vivianwhite If you are interested, I think this is an issue you could solve. There are two images in the readme, and we want to serve them through kymat.io. we can store images in a folder called `_static`. you just need to figure out which images we want to server and how to serve them through `_static`.

MuawizChaudhary:
Closed by #1030

PR #1030, opened 2023-10-31
DOC Move ENS logo to kymat.io website

janden:
So this should solve #950. We now self-host both the FI and ENS logos. This should enable them to be seen both on the GH starting page and on kymat.io.

Note, however, that these logos won't show up on the README in PyPI since we explicitly filter out links in https://github.com/kymatio/kymatio/blob/8f2bbdfdbe974de82caf52b07826859c02d3bd24/setup.py#L17.

If we want to change that, we can discuss, but that's the current behavior.


lostanlen:
@janden i see nothing at https://www.kymat.io/_static/ens_logo.jpg at the moment?

janden:
> @janden i see nothing at https://www.kymat.io/_static/ens_logo.jpg at the moment?

Right. This won't show up on kymat.io until we merge in the PR and run the push docs pipeline.

edouardoyallon (approved):

MuawizChaudhary (approved):
I approve these changes.

PR #1036, opened 2023-11-03
DOC Fix docstrings

janden:
Fixes some issues with docstrings that have cropped up due to API updates and Sphinx version drift.

### DIFF
--- a/doc/source/index.rst
+++ b/doc/source/index.rst
@@ -4,6 +4,7 @@ Content of this website:
 ------------------------
 
 .. toctree::
+   :maxdepth: 2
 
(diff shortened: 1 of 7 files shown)


janden:
@lostanlen Would be good if you could take a look at this. Need to merge this before getting the JTFS docstrings in.

lostanlen (approved):
```
        # Sphinx will not show docstrings for inherited methods, so we add a
        # dummy method here that will just call the super.
```

good to know!

janden:
> good to know!

Yes was quite a pain, but now we know…

Issue #1045, opened 2024-04-24
BUG - my bad


edouardoyallon:
Woops - I screw up by merging some approved PRs whose tests were passing 6 months ago but not now .. maybe we should fix that ..

MuawizChaudhary:
@janden I propose we merge the latest reversion #1047 and then try to fix the tests. Or should we fix in #1047?

MuawizChaudhary:
> FAILED tests/scattering3d/test_utils_scattering3d.py::test_utils - TypeError: exceptions must be derived from Warning, not <class 'NoneType'>

This error occurs in python 3.8. I'll focus on solving this issue first.

MuawizChaudhary:
Seems like this issue results from deprecation of `pytest.warns(None)`. the solution according to this reply [here](https://github.com/scikit-learn/scikit-learn/issues/22572#issuecomment-1047316960) is to replace with `warnings.catch_warnings(record=True)`.

MuawizChaudhary:
>ValueError: Cannot convert '512' to a shape.
This error occurs in the scattering1d test, for keras, in versions >= python3.9.

MuawizChaudhary:
I have no idea why this error occurs in the change in versions. All I know is that this error occurs for versions >= python 3.9, and it occurs on the first test of the keras 1d suite of tests `test_Scattering1D`.

MuawizChaudhary:
#1048 solves this issue.

MuawizChaudhary:
#1048 has resolved this issue.

PR #1048, opened 2024-05-04
Fixes TypeError, Value Error in tests.

MuawizChaudhary:
This PR updates one of our tests to not use deprecated pytest code for checking if no warnings were thrown on complex input. 

Another test is also updated to convert shape inputs of type `int` to type `tuple`.

Should Solve #1045


janden (approved):
Looks good to me. Thanks @MuawizChaudhary for putting this together!

MuawizChaudhary:
I'm squash+merging in light of @janden's review.

Issue #1029, opened 2023-10-31
Bug in JTFS S1 shape and `subsample_fourier` during frequency scattering (numpy and tensorflow)


cyrusvahidi:
In time-frequency scattering we stack the S1 paths along the 2nd axis frequency:

https://github.com/kymatio/kymatio/blob/8f2bbdfdbe974de82caf52b07826859c02d3bd24/kymatio/scattering1d/core/timefrequency_scattering.py#L157 

This is correct in torch, since the default value for `axis=2`, however, `axis=1` in `numpy_backend` and `tensorflow_backend`, resulting in an array of shape `[batch, \lambda_1, 1, t]`. This has knock on effects for frequency scattering. This is a simple fix:

`S_1 = backend.stack(S_1_list, dim=2)`.

Another issue arises in `subsample_fourier` in tensorflow and numpy during frequency scattering:

https://github.com/kymatio/kymatio/blob/8f2bbdfdbe974de82caf52b07826859c02d3bd24/kymatio/scattering1d/core/timefrequency_scattering.py#L265 

Numpy and TF backends follow the same logic for reshaping:

https://github.com/kymatio/kymatio/blob/8f2bbdfdbe974de82caf52b07826859c02d3bd24/kymatio/scattering1d/backend/numpy_backend.py#L28 

https://github.com/kymatio/kymatio/blob/8f2bbdfdbe974de82caf52b07826859c02d3bd24/kymatio/scattering1d/backend/tensorflow_backend.py#L30

which ignores the extra first dimension. This has serious consequences during subsampling, as the time dimension is swallowed into the batch dimnension. 

The torch backend makes sure to handle all dimensions upto the last dimension that we're subsampling over:

https://github.com/kymatio/kymatio/blob/8f2bbdfdbe974de82caf52b07826859c02d3bd24/kymatio/scattering1d/backend/torch_backend.py#L46 

The numpy and tensorflow implementations can be easily fixed by fixing the reshaping in `subsample_fourier`.

This arose when implementing Keras frontend. I will make a PR that fixes the stacking of S1, the reshaping logic in np and tf `subsample_fourier` and implements the Keras frontend.

eickenberg:
these should all be aligned and the backend should regularize everything to
numpy convention (which for stack means axis=0 by default)

cyrusvahidi:
> these should all be aligned and the backend should regularize everything to numpy convention (which for stack means axis=0 by default)
> […](#)

[...]

PR #1052, opened 2024-05-25
Fix outstanding Tensorflow and Numpy errors as mentioned in #1029

MuawizChaudhary:
As noted in #1029, the stack operation stacks frequency(?) coefficients on the wrong axis in numpy and tensorflow implementation of frequency scattering, and the Fourier sub-sampling absorbs specific dims that it shouldn't in Numpy and Tensorflow. 

This PR modifies the core time frequency algorithm to stack on the 2nd axis, modifies the Fourier sub-sampling to be agnostic in dim up to sub-sampling dim in numpy and tensorflow, and slightly modifies tests to account for the correct stacking behavior and to include tensorflow in more tests. 

@lostanlen @cyrusvahidi would appreciate your thoughts and review.


kymatio/scattering1d/backend/tensorflow_backend.py
```diff
@@ -27,9 +27,14 @@ def subsample_fourier(cls, x, k):
         """
         cls.complex_check(x)
 
-        y = tf.reshape(x, (-1, k, x.shape[-1] // k))
+        shape = list(x.shape[:-1]) + [k, x.shape[-1]//k]
+        if shape[0] is None:
+            shape[0] = -1
```
MuawizChaudhary, line 32: @janden @eickenberg do y'all think this additional code is fine? 

The Keras backend was erroring out when I tried to reshape. x.shape produces a datatype with None as the first element. The None specifies a general "batch" shape in keras. I need to convert this None to a -1 to play nice with Tensorflow.

edouardoyallon: This sounds good to me!

edouardoyallon (approved):

edouardoyallon:
looks good to me (on the other hand my last PR was a failure ..)

PR #1060, opened 2025-01-26
Fix Unit Tests

MuawizChaudhary:
This PR fixes a jax unit test to import config directly from jax. The previous method (importing from jax.config module) has been deprecated. 

We also set the version of Numpy to 1.24.0 to get the Pytorch suite of unit tests running again. Should be discussed if this change is fine, or how we want to work around it. 

We should discuss how we want to handle the harmonic scattering test failing due to a OOM error on certain versions of Python + PyTorch.

### DIFF
--- a/.github/workflows/conda.yml
+++ b/.github/workflows/conda.yml
@@ -35,7 +35,7 @@ jobs:
                   conda create -q -n ${CONDA_ENV} python=${{ matrix.python-version }}
             - name: Install dependencies
               run: |
-                  conda install -n ${CONDA_ENV} numpy scipy pytest pytest-cov
+                  conda install -n ${CONDA_ENV} numpy=1.26.4 scipy pytest pytest-cov
                   conda install -n ${CONDA_ENV} pytorch=${{ matrix.pytorch-version }} torchvision cpuonly -c pytorch
 
                   conda run -n ${CONDA_ENV} python3 -m pip install --upgrade pip
--- a/.github/workflows/conda.yml
+++ b/.github/workflows/conda.yml
@@ -35,7 +35,7 @@ jobs:
                   conda create -q -n ${CONDA_ENV} python=${{ matrix.python-version }}
             - name: Install dependencies
               run: |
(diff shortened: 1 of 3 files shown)


lostanlen:
what happened to Conda / Pip CI's ?

MuawizChaudhary:
I'm unsure. I am aware there is one issue where when pytorch calls `from_numpy` it fails with "numpy is not available" or some error of that sort. 

Currently investigating, but I may have to hold off given I have a job interview and some meetings to attend to tomorrow morning.

MuawizChaudhary:
The Conda and Pip tests are mostly running after using an earlier numpy version and updating typing extensions for the conda set of tests. Most tests run to completion. 

Locally `test_solid_harmonic_scattering` has been causing OOM errors. I believe that its specific to torch=2.0.0 the error, tho it likely happens at versions greater than 2.0.0. Ideally it'd be nice to fix this. However, in the short term, I believe we should disable this test until we figure out whats going on, or see if we can page more memory for the test. Paging @eickenberg. 

That being said, I think this PR can be merged and we can continue development.

lostanlen:
Thanks @MuawizChaudhary 
seems like you are pinning a specific version of numpy with this PR? `1.24`
isnt't this a problem for future maintenance?
i also fail to see the connection with JAX

MuawizChaudhary:
I should generalize this PR to be a "Fix All Unit Tests" PR.

MuawizChaudhary:
On the Numpy front, yes this is essentially mandating that the numpy version is kept at whatever supports python 3.9 to 3.11 inclusive for the sake of testing torch.  Numpy 1.24.0 is one such version. 


Ideally, we would have framework specific (and even version specific) workflows that are ran by GitHub Actions. we could have tests for the numpy specific scattering on the current version this way, and not interfere with the pytorch evaluation. Remember, we are using an old version of pytorch released a few years ago. So the accompanying libraries should be of a version specific to the pytorch version. 

Or we could update the pytorch version and possibly navigate around this issue of Numpy not being built for the 2.0 version of Pytorch.

PR #1065, opened 2026-02-10
FIX scipy deprecated function

hugo-paugesteros:
`scipy.special.sph_harm` has been deprecated since since scipy 1.15.0 (jan 2025) and removed in scipy 1.17.0 (jan 2026). 
Replaced by `scipy.special.sph_harm_y`

### DIFF
--- a/kymatio/scattering3d/filter_bank.py
+++ b/kymatio/scattering3d/filter_bank.py
@@ -1,7 +1,7 @@
 __all__ = ['solid_harmonic_filter_bank']
 
 import numpy as np
-from scipy.special import sph_harm, factorial
+from scipy.special import sph_harm_y, factorial
 from .utils import get_3d_angles, double_factorial, sqrt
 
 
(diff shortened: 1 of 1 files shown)


MuawizChaudhary:
Hi Hugo, thanks for the PR.

I last tried this in 2024, but I believe that scipy.special.sph_harm_y did not give same output as the original scipy.special.sph_harm.

Do the tests run on your end? Do they pass?

Unfortunately the cicd appears to be broken, the tests don't appear to be running. I won't have time for a while to fix the issue but I'll try to diagnose why there is an issue.

RichieHakim:
This is of interest to me as well. Kymatio is an important package for something I maintain and it is holding back scipy version updates. (thanks for maintaining this great library!)

NathanRWoodward:
Also would be helpful for me, currently having to monkey patch before import:

```
import scipy.special
scipy.special.sph_harm = scipy.special.sph_harm_y
from kymatio.torch import Scattering1D 
```

PR #1059, opened 2025-01-23
API adding barebones keras timefreqscatter. 

MuawizChaudhary:
Resolves #1053. We need to figure out how to `compute_output_shape`, given that #839 was not incorporated. 

@lostanlen your advices here would be appreciated. See #1053 for prior discussion.

PR #1067, opened 2026-06-08
ci: ubuntu-22.04 runners + fix jax/scipy dependency rot

cyrusasfa:
## Why
CI had gone dark: every workflow pinned `runs-on: ubuntu-20.04`, whose GitHub-hosted image was retired, so the conda/pip jobs sat `queued` forever (no matching runner). Once runnable again, the matrix surfaced pre-existing dependency rot.

## Changes
- **ubuntu-22.04 runners** (`conda.yml`, `pip.yml`, `docs.yml`) — replaces the retired ubuntu-20.04 image so jobs get a runner. Chose 22.04 (lowest-risk hop from 20.04; reliably serves the py3.8 matrix) over `ubuntu-latest`.
- **jax/jaxlib find-links** — `pip install jax jaxlib` failed on py3.8 (`No matching distribution found for jaxlib`); PyPI ships no cp38 jaxlib wheels. Added `-f https://storage.googleapis.com/jax-releases/jax_releases.html`, which serves `jaxlib-0.4.13-cp38`, so py3.8 resolves without dropping the version.
- **scipy<1.15** (`requirements.txt` + the conda install) — `scipy.special.sph_harm` (used in `scattering3d/filter_bank.py`) is removed in scipy>=1.15, which breaks `import kymatio` on newer-Python jobs. Pinned as a stopgap.
- **`fail-fast: false`** on both matrices — so a single run surfaces every failure instead of cancelling siblings on the first error.

## Notes
- Unblocks CI for the JTFS Keras frontend work in #1059 (that PR will rebase onto this once merged).
- Follow-ups: proper `scipy.special.sph_harm_y` port (arg-order + polar/azimuthal convention change) to drop the scipy pin; revisit `deeplake[audio]` / old-package installs if they surface with fail-fast off.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
 numpy
-scipy<1.15  # scipy.special.sph_harm removed in scipy>=1.15; sph_harm_y port pending (gh-1053 follow-up)
+scipy<1.15
 appdirs
 configparser
 packaging
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
 numpy
-scipy
+scipy<1.15  # scipy.special.sph_harm removed in scipy>=1.15; sph_harm_y port pending (gh-1053 follow-up)
 appdirs
 configparser
 packaging
--- a/.github/workflows/conda.yml
+++ b/.github/workflows/conda.yml
@@ -9,6 +9,7 @@ jobs:
         runs-on: ubuntu-22.04
 
         strategy:
+            fail-fast: false
             matrix:
                 python-version: ['3.8', '3.9', '3.10']
                 pytorch-version: ['1.10', '1.13', '2.0']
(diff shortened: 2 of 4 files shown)


requirements.txt
```diff
@@ -1,5 +1,5 @@
 numpy
-scipy
+scipy<1.15  # scipy.special.sph_harm removed in scipy>=1.15; sph_harm_y port pending (gh-1053 follow-up)
```
MuawizChaudhary: Remove this comment.

MuawizChaudhary (commented):
Looks good to me. just remove the comment and I'm pro merging.

MuawizChaudhary (approved):

MuawizChaudhary:
Why is jenkins not running here?

cyrusasfa:
Thanks @MuawizChaudhary!

- **Comment removed** in `68962e4` (the rationale lives in the commit message + PR description). Full matrix is green — 36/36 on the pin commit; the comment-only push is just re-running and will match.
- **Jenkins:** that's the separate self-hosted **GPU CI at Flatiron** (`Jenkinsfile`: `--gpus 1`, CUDA / `tf.test.is_gpu_available()` checks), not the GitHub-hosted conda/pip Actions. It's currently not reporting to GitHub at all — `main`'s HEAD also shows no Jenkins status/check-runs — so its absence isn't specific to this PR; the Jenkins↔GitHub integration just isn't posting right now (server/webhook side, nothing in this PR touches it). These Actions checks aren't gated on it.

Issue #1068, opened 2026-06-30
Scipy 1.17 deprecated methods


SBouchard01:
Scipy 1.17 has deprecated `scipy.special.sph_harm` in favor of `scipy.special.sph_harm_y`. This crashes all current calls to kymatio.

Is there a possibility to fix this soon? 
> [!WARNING]
> Also note that the order of the n, m arguments and the definition of the theta, phi angles has changed from `sph_harm` to `sph_harm_y`! `theta` used to be the azimuthal angle, now it is the polar angle. So even though the order of `theta`, `phi` in the call signature remains the same they are actually swapped.
> 

PR #1065 seems to fix part of it but does not inverts the order of parameters.

Thank you very much !

gvalogia:
Yes, I second this as I have also come across this issue!

PR #1069, opened 2026-07-03
[Fix] Update to scipy 1.17 (#1068)

SBouchard01:
Because of the update to `scipy` 1.17, `sph_harm` [has been changed](https://docs.scipy.org/doc/scipy-1.16.1/reference/generated/scipy.special.sph_harm.html) to `sph_harm_y` (and inverted polar and azimutal angles).

This PR updates scipy to 1.17 in the requirements, workflows and readme, and updates the `sph_harm` implementation in `kymatio/scattering2d/filter_bank.py`.

> I hope this mitigate the workflow issues from #1065

### DIFF
--- a/.github/workflows/conda.yml
+++ b/.github/workflows/conda.yml
@@ -36,7 +36,7 @@ jobs:
                   conda create -q -n ${CONDA_ENV} python=${{ matrix.python-version }}
             - name: Install dependencies
               run: |
-                  conda install -n ${CONDA_ENV} numpy "scipy<1.15" pytest pytest-cov
+                  conda install -n ${CONDA_ENV} numpy "scipy>=1.17" pytest pytest-cov
                   conda install -n ${CONDA_ENV} pytorch=${{ matrix.pytorch-version }} torchvision cpuonly -c pytorch
 
                   conda run -n ${CONDA_ENV} python3 -m pip install --upgrade pip
--- a/README.md
+++ b/README.md
@@ -59,7 +59,7 @@ Andreux M., Angles T., Exarchakis G., Leonarduzzi R., Rochette G., Thiry L., Zar
 Kymatio requires:
 
 * Python (>= 3.7)
-* SciPy (>= 0.13)
+* SciPy (>= 1.17)
 
 
 ### Standard installation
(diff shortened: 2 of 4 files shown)


scaramouche-00:
Independent verification of this PR, in case it helps it move — we hit #1068 from the other side (kymatio 0.3.0 is simply not importable on current scipy, and `tests/scattering3d/` cannot even be collected), and have been carrying a local `sph_harm` shim to work around it.

**Environment:** Python 3.12.3, scipy 1.18.0, numpy 2.5.1, PyTorch 2.10.0+rocm7.14.0, CPU only (no GPU used, so the skcuda backend is untested here).

**1. The tests pass on this branch with no shim of any kind.**

```
pytest tests/scattering3d/test_torch_scattering3d.py \
       tests/scattering3d/test_numpy_scattering3d.py \
       tests/scattering3d/test_utils_scattering3d.py
11 passed
```

**2. The argument reordering is numerically correct.**

This is the part worth checking, because scipy 1.17 swapped *both* the (degree, order) pair and the (theta, phi) convention, so the edit

```python
-  solid_harm[i_m] = sph_harm(m, l, azimuthal, polar) * polynomial_gaussian
+  solid_harm[i_m] = sph_harm_y(l, m, polar, azimuthal) * polynomial_gaussian
```

could look right and be silently wrong — and a wrong spherical harmonic would not necessarily fail the existing tests, which are largely self-consistency checks.

I compared `sph_harm_y(l, m, polar, azimuthal)` against an independent reimplementation of the old `sph_harm(m, l, azimuthal, polar)` from the definition (associated Legendre plus explicit normalisation, *not* scipy's removed function, so this isn't circular), over 4096 random angle pairs for every `(l, m)` with `l <= 6`:

```
worst absolute difference over all (l,m) with l<=6: 2.407e-13
```

i.e. identical to machine precision. The reordering is right.

**One question rather than an objection:** this pins `scipy>=1.17` in `requirements.txt` and the README, which drops support for older scipy entirely. A `try: from scipy.special import sph_harm_y / except ImportError:` fallback to the old call would keep both working, if you would rather not raise the floor. Your call — the hard pin is simpler and I have no stake in it.

Unrelated to this PR, but found while testing on the same branch: `HarmonicScattering3D` returns a finite forward and an all-NaN gradient for sparse float32 inputs, because `modulus_rotation` ends in a bare `torch.sqrt` on a quantity that reaches exactly zero. Filed separately as #1070.

*Verification carried out by Claude (Anthropic), an AI assistant working on a cosmological emulator that uses the 3D scattering transform. I have reviewed it.*

